PR #1, opened 2020-06-25
Add operator skeleton from operator-sdk

vatsalparekh:
Added the operator skeleton generated by `operator-sdk`
Also added the make file to pin the version of sdk we use


Makefile
```diff
@@ -0,0 +1,11 @@
...
+IMAGE_REGISTRY ?= quay.io
+OPERATOR_IMAGE ?= kubevirt-ssp-operator
+IMAGE_TAG ?= latest
+
+operator-sdk:
+	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
+	chmod 0755 operator-sdk
+
+operator-build:
+	operator-sdk build $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
```
omeryahud, line 11: We should run `go mod tidy` and `go mod vendor` before building so we don't miss dependencies and add the `--go-build-args` to `operator-sdk build` to make sure we are building with vendored dependencies

omeryahud: Let's also add an `operator-push` target to make image updates quicker

omeryahud: Calls to operator-sdk should be made with `./operator-sdk [args]`, as `operator-sdk [args]` would use the `$PATH` binary and not the downloaded one in the repo directory

.gitignore
```diff
@@ -0,0 +1,79 @@
+# Temporary Build Files
```
akrejcir, line 1: I know this `.gitignore` file is generated, but can you remove all entries not generated by the build process of the operator? 

All entries related to IDEs can be in a single global gitignore file on the developer's machine, not related to this repo.

go.mod
```diff
@@ -0,0 +1,17 @@
...
+require (
+	github.com/operator-framework/operator-sdk v0.18.2
+	github.com/spf13/pflag v1.0.5
+	k8s.io/api v0.18.2
+	k8s.io/apimachinery v0.18.2
+	k8s.io/client-go v12.0.0+incompatible
+	sigs.k8s.io/controller-runtime v0.6.0
+)
+
+replace (
```
akrejcir, line 14: Why do these packages need to be in the `replace` block and not only in `require`?

Makefile
```diff
@@ -0,0 +1,11 @@
+OPERATOR_SDK_VERSION ?= v0.18.2
+IMAGE_REGISTRY ?= quay.io
```
ksimon1, line 2: Shouldn't be here something like `quay.io/oyahud` ?

Makefile
vatsalparekh: Added

.gitignore
vatsalparekh: Moved to manually add only what we need

go.mod
vatsalparekh: I'm not sure here, this was generated by `operator-sdk` binary

Makefile
vatsalparekh: Added

go.mod
akrejcir: Ok, I was just curious. If it was generated, we can assume there was a reason for it.

vatsalparekh: I verified it on a example operator also, this is default generated for me

Makefile
```diff
@@ -1,11 +1,17 @@
...
 OPERATOR_IMAGE ?= kubevirt-ssp-operator
 IMAGE_TAG ?= latest
+OPERATOR_BUILD_ARGS ?= "-v -mod=vendor"
 
 operator-sdk:
 	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
-	chmod 0755 operator-sdk
+	chmod 0755 ./operator-sdk
 
 operator-build:
```
akrejcir, line 11: This make target uses operator-sdk, so it should depend on the `operator-sdk` target:
``` make
operator-build: operator-sdk
    go mod tidy
    ...
```

Makefile
```diff
@@ -1,11 +1,17 @@
...
-	operator-sdk build $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+	go mod tidy
+	go mod vendor
+	./operator-sdk build $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) --go-build-args $(OPERATOR_BUILD_ARGS)
+
+operator-push:
+	docker push $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
```
akrejcir, line 17: The targets `operator-build` and `operator-push` are not names of files, so it is a good idea to declare them as phony:
``` make
.PHONY: operator-build operator-push
```

cmd/manager/main.go
```diff
@@ -0,0 +1,211 @@
...
+	"os"
+	"runtime"
+	"strings"
+
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
+
+	"github.com/omeryahud/ssp-operator/pkg/apis"
+	"github.com/omeryahud/ssp-operator/pkg/controller"
```
ksimon1, line 17: What about already starting using kubevirt package path? I know we should use `github.com/omeryahud/ssp-operator`, but in few weeks we will migrate it under kubevirt org and then we will have to change all paths.

akrejcir: I think it depends on users of this package. Are we going to migrate to the 'kubevirt.io' path before the HCO will use this package? Wouldn't it cause problems with go vendor if the module package path is different from the repository?

.gitignore
```diff
@@ -1,79 +1,6 @@
...
-# temporary
-.netrwhist
-# auto-generated tag files
-tags
-### VisualStudioCode ###
 .vscode/*
 .history
 .idea
-# End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
+operator-sdk
```
akrejcir, line 4: When adding Travis, I have noticed that this line also excludes all files in directory `vendor/github.com/operator-framework/operator-sdk/`. This is probably not what we want. Can you change the line to `/operator-sdk` so only the file at the top level directory is excluded?

cmd/manager/main.go
omeryahud: I agree, we should generate the operator to declare the `kubevirt.io/ssp-operator` module.
This would require developers to clone `github.com/omeryahud/ssp-operator` to `$GOPATH/src/kubevirt.io/ssp-operator` manually for a while, but it would save us a huge headache when migrating the operator to the kubevirt organization.
Vatsal, could you re-generate the operator? If I'm not mistaken, the only manual changes made here were to the makefile/gitignore that can be easily migrated over.

Makefile
vatsalparekh: Added

cmd/manager/main.go
vatsalparekh: `kubevirt.io/ssp-operator` seems correct way to do it, generated a new skeleton now

omeryahud:
/lgtm
@ksimon1 @akrejcir Please take a final look and lets merge this

PR #4, opened 2020-07-21
Add metrics package

akrejcir:
The metrics component creates a `PrometheusRule` or updates it during reconciliation.


akrejcir:
@ksimon1 @omeryahud @vatsalparekh

pkg/controller/ssp/metrics/reconcile.go
```diff
@@ -0,0 +1,66 @@
...
+		request.Logger.Info(fmt.Sprintf("Creating PrometheusRule: %s", rule.Name))
+		return request.Client.Create(request.Context, rule)
+	}
+	if err != nil {
+		return err
+	}
+
+	found.Spec = rule.Spec
+	request.Logger.Info(fmt.Sprintf("Updating PrometheusRule: %s", found.Name))
+	return request.Client.Update(request.Context, &found)
```
omeryahud, line 42: From my experience, `Update()`ing resources this way might cause an infinite reconciliation loop (each update triggers another reconciliation loop which triggers another update and so on, even if the updated resource was not actually changed), have you encountered something to that effect?
If so, you can use `Patch()`, which does not trigger a reconcile loop if the patched resource was not actually changed.

akrejcir: In my test it did not cause an infinite loop, but the update was executed twice. I will check it again.

akrejcir: I have added a check for equality.

pkg/controller/ssp/metrics/reconcile.go
```diff
@@ -0,0 +1,71 @@
+package metrics
```
omeryahud, line 1: I have a suggestion for a folder structure:
```
internal/pkg/operands/$operand/
                     | - reconcile.go
                     | - resources/
                          | - prometheusrules.go
                          | - deployments.go
                          | - services.go
                          | - $resourcename.go
```

and all resource generating functions like `newPrometheusRule` go in the respective resource file.
And we get the protection of not exporting those packages as a bonus.

What do you think?

akrejcir: I would simplify it. If we have only one function for each resource type, we don't need to put them to separate files. They can all be in a single `resources.go` file.

And I'm not sure about using `internal` directory. Probably no one is going to import modules from `pkg/controller`.
I couldn't find any recommendations when to make packages internal. It would make sense if we were developing a library, but for an application, I don't know.

omeryahud: Sure simplifying it to a single `resources.go` file is also a great alternative.
with regards to the `internal` folder, my approach is "unexport everything exept for...", I just think it makes more sense.

akrejcir: Ok, I changed the directory structure.

omeryahud:
Can you see if you can add some unit tests, maybe by mocking `req.Client`?
If you'd like, you can get inspiration/ideas from this code: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-controller/watch/vm_test.go

akrejcir:
Ok, I have added simple unit tests for the reconcile of metrics operand.

PR #5, opened 2020-08-20
Add validator operand

akrejcir:
Creates and watches template-validator resources.


internal/common/resource_test.go
```diff
@@ -0,0 +1,184 @@
...
+			},
+			Logger: log,
+		}
+	})
+
+	It("should create resource", func() {
+		Expect(CreateOrUpdateResource(&request,
+			newTestResource(namespace),
+			newEmptyResource(),
+			func(_ Resource, _ Resource) bool {
```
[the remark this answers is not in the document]
akrejcir: Changed

internal/common/resource.go
```diff
@@ -0,0 +1,115 @@
...
+func CreateOrUpdateClusterResource(request *Request, resource Resource, found Resource, updateResource ResourceUpdateFunc) error {
+	addOwnerAnnotations(resource, request.Instance)
+	return createOrUpdate(request, resource, found, updateResource)
+}
+
+func createOrUpdate(request *Request, resource Resource, found Resource, updateResource ResourceUpdateFunc) error {
+	err := request.Client.Get(request.Context,
+		types.NamespacedName{Name: resource.GetName(), Namespace: resource.GetNamespace()},
+		found)
+	if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
akrejcir: No risk, the `errors.IsNotFound` returns `false` if `err` is `nil`.

internal/common/environment.go
```diff
@@ -0,0 +1,21 @@
+package common
+
+import "os"
+
+const (
+	templateValidatorImage = "IMAGE_REFERENCE"
```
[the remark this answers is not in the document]
akrejcir: Renamed

pkg/apis/ssp/v1/ssp_types.go
```diff
@@ -4,14 +4,14 @@ import (
...
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
-
 // SSPSpec defines the desired state of SSP
 type SSPSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// ValidatorReplicas is the number of replicas of the template validator pod
+	ValidatorReplicas int32 `json:"template_validator_replicas,omitempty"`
```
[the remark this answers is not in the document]
akrejcir: Changed

internal/operands/template-validator/resources.go
```diff
@@ -0,0 +1,203 @@
...
+	virtTemplateValidator = "virt-template-validator"
+)
+
+var commonLabels = map[string]string{
+	kubevirtIo: virtTemplateValidator,
+}
+
+func clusterRoleName(namespace string) string {
+	const clusterRoleName = "template:view"
+	return clusterRoleName + "-" + namespace
```
omeryahud, line 32: Why do we integrate the namespace name into the clusterRole name?

internal/operands/template-validator/resources.go
```diff
@@ -0,0 +1,203 @@
...
+			Namespace: namespace,
+			Labels:    commonLabels,
+		},
+	}
+}
+
+func newClusterRoleBinding(namespace string) *rbac.ClusterRoleBinding {
+	return &rbac.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   serviceAccountName + "-" + namespace,
```
omeryahud, line 62: Same question here

pkg/apis/ssp/v1/ssp_types.go
```diff
@@ -4,14 +4,18 @@ import (
...
+}
 
 // SSPSpec defines the desired state of SSP
 type SSPSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// TemplateValidator is configuration of the template validator operand
+	TemplateValidator TemplateValidator `json:"template_validator"`
```
omeryahud, line 18: I would use `json:"templateValidator"` instead, I think it is easier on the eye and easier to type :)

pkg/controller/ssp/ssp_controller.go
```diff
@@ -50,19 +57,46 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	return addSecondaryWatches(c,
+	err = watchClusterResources(c,
+		template_validator.WatchClusterTypes,
+	)
```
omeryahud, line 62: Maybe add `// TODO - add other watch types here` here too?

pkg/controller/ssp/ssp_controller.go
```diff
@@ -126,3 +173,47 @@ func (r *ReconcileSSP) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+}
+
+func hasFinalizer(object metav1.Object, finalizer string) bool {
+	for _, f := range object.GetFinalizers() {
+		if finalizer == f {
+			return true
+		}
+	}
+	return false
+}
```
omeryahud, line 219: Missing empty line here

omeryahud (changes requested):
Looks really good!
Some small nits

pkg/controller/ssp/ssp_controller.go
akrejcir: Fixed

pkg/controller/ssp/ssp_controller.go
akrejcir: Added

pkg/apis/ssp/v1/ssp_types.go
akrejcir: Changed. I see that other fields use camelCase too. The underscore version was added automatically by IDE.

internal/operands/template-validator/resources.go
akrejcir: Multiple `SSP` resources can be created in different namespaces. In that case the operator creates multiple cluster roles, and they need different names. If we only used a single cluster role for all `SSPs`, then it would be tricky to implement deletion.

internal/operands/template-validator/resources.go
akrejcir: The same reson.

omeryahud:
Now that the placement API is merged, please add it to this PR as well (https://github.com/kubevirt/kubevirt-ssp-operator/pull/226)
I was thinking about something like this:
```
SSP.spec.templateValidator.nodeSelector
SSP.spec.templateValidator.affinity
SSP.spec.templateValidator.tolerations
```

akrejcir:
Added placement API.

internal/common/environment.go
```diff
@@ -0,0 +1,21 @@
...
+func envOrDefault(envName string, defVal string) string {
+	val := os.Getenv(envName)
+	if val == "" {
+		return defVal
+	}
+	return val
+}
+
+func GetTemplateValidatorImage() string {
+	const defaultVersion = "v0.6.6"
```
ksimon1, line 18: there is a new version v0.7.0 https://github.com/kubevirt/kubevirt-template-validator/releases/tag/v0.7.0

internal/operands/metrics/reconcile_test.go
```diff
@@ -57,32 +55,16 @@ var _ = Describe("Metrics operand", func() {
...
 			Logger: log,
 		}
-	})
-
-	It("should create the prometheus rule", func() {
-		Expect(Reconcile(&request)).ToNot(HaveOccurred())
-		expectEqualRuleExists(newPrometheusRule(namespace), request)
-	})
-
-	It("should update the prometheus rule", func() {
```
ksimon1, line 67: this test is no longer needed?

akrejcir: No. The tests for the reconciliation logic were moved to `internal/common/resource_test.go`.
The tests here only check if specific resources are reconcilied, in this case the prometheus rule.

internal/common/environment.go
akrejcir: Updated

ksimon1:
Looks good to me, just one thing. Would be possible to add tests which tests placement api? e.g. create CR with affinity, then remove affinity. Because currently in old ssp we encountered bug, that it can't remove affinity from deploy manifest.

akrejcir:
Would the placement values depend on the cluster where the tests are run? I will try to add the tests, maybe the deployment will be created even with incorrect placement values.

ksimon1:
I would do it like we do it in old ssp https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/functests/10-test-template-validator-placement.sh (with fake values)

akrejcir:
Added the test. Removing placement works fine.

ksimon1:
/lgtm 
@vatsalparekh can you please review?

PR #9, opened 2020-09-16
Update operator-sdk version to v1.0.0

akrejcir:
This PR recreates the operator using the operator-sdk v1.0.0.
All the old generated files were removed, but our code is the same.


Makefile
```diff
@@ -1,34 +1,123 @@
...
+endif
+BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
-operator-sdk:
-	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
-	chmod 0755 ./operator-sdk
+# Image URL to use all building/pushing image targets
+IMG ?= quay.io/oyahud/kubevirt-ssp-operator:latest
+# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
+CRD_OPTIONS ?= "crd:trivialVersions=true"
```
ksimon1, line 17: will this work with openshift?

akrejcir: Probably yes. I run the functional tests on my cluster that is ocp 4.5.7.

But I'm not sure if we need this parameter, maybe it can be removed.

akrejcir: I removed the parameter and generated manifests are the same.

.gitignore
```diff
@@ -4,3 +4,4 @@
 /operator-sdk
 _out/*
 build/_output/*
+bin/*
```
vatsalparekh, line 7: Can you add a new line at the end

ksimon1 (commented):
/lgtm

Makefile
```diff
@@ -1,34 +1,121 @@
-OPERATOR_SDK_VERSION ?= v0.18.2
-IMAGE_REGISTRY ?= quay.io/oyahud
-OPERATOR_IMAGE ?= kubevirt-ssp-operator
-IMAGE_TAG ?= latest
-OPERATOR_BUILD_ARGS ?= "-v -mod=vendor"
+# Current Operator version
+VERSION ?= 0.0.1
```
vatsalparekh, line 2: Shouldn't this be `1.0.0`?

Makefile
```diff
@@ -1,34 +1,121 @@
...
+ifneq ($(origin CHANNELS), undefined)
+BUNDLE_CHANNELS := --channels=$(CHANNELS)
+endif
+ifneq ($(origin DEFAULT_CHANNEL), undefined)
+BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
+endif
+BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
-operator-sdk:
-	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
```
vatsalparekh, line 8: I don't see any replacement of this, how do we ensure that operator-sdk version is pinned to the exact

vatsalparekh (changes requested):

Makefile
akrejcir: `1.0.0` is usually the first complete released version. This `0.0.1` was auto-generated. I think for now it is good, and later we can update it when the operator is released.

Makefile
akrejcir: `operator-sdk` binary is not needed to build, it was only used to generate the repo, add API and controller code. The make targets use `kustomize` and `controller-gen`. Both of them have fixed versions in the makefile.

We could mention the version in a readme or documentation.

.gitignore
akrejcir: Done

omeryahud (approved):

PR #15, opened 2020-09-30
Fill default fields for resources

akrejcir:
Without the default fields filled, the resource would always be updated to remove them.

It would not cause an infinite loop, but the operator would always be in 'updating' state and never finish.
This would cause problems when status and conditions.

**Release note**:
```release-note
NONE
```


internal/operands/template-validator/resources.go
```diff
@@ -113,6 +117,7 @@ func newDeployment(namespace string, replicas int32, image string) *apps.Deploym
 	const volumeName = "tls"
 	const certMountPath = "/etc/webhook/certs"
 	trueVal := true
+	percentVal := intstr.FromString("25%")
```
ksimon1, line 120: Is this value taken from somewhere? Or how do you know this is correct value?

akrejcir: All defaults are from the documentation that is on the struct fields.
I have ale checked it in OCP 4.5.7 cluster.

akrejcir: Unfortunately, the code that sets these defaults is part of the `k8s.io/kubernetes` repo, which cannot be easily imported as module.

omeryahud:
I think moving all default values assignments to dedicated functions would be more readable, something like `ApplyDefaultValues()` function for each resource in an operand, this makes it clearer what we set explicitly, and what is defaulted.
It can look something like this:
```
validatorDeployment := &apps.Deployment{}
ApplyDefaultValues(validatorDeployment)

// Explicitly set non-default fields
...

return validatorDeployment
```
This makes it that when we decide to eventually update a field which was previously defaulted, we can do so by just modifying it after `ApplyDefaultValues()`

akrejcir:
Ok, I can do that. There are already functions that fill in the defaults in kubernetes repo:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/v1/defaults.go#L35

But because importing kubernetes module does not work (their module dependencies are somehow broken), I will copy the code.

akrejcir:
The new code uses the same code as kubernetes to set the defaults.

akrejcir:
It may be possible to import kubernetes as a module. Moving to draft for now.

akrejcir:
To import kubernetes, `go.mod` file would need to `replace` all the wrong module versions used in kubernetes. It is probably not worth doing it.

omeryahud:
I just had another idea, maybe instead of chasing defaults for resources we use (or might use in the future), maybe we can think about logic like this:
```
existingDeploymentStateJSON := json.Marshal(existingDeployment)
mergedDeploymentStateJSON := json.Marshal(existingDeployment)

wantedDeploymentStateJSON := json.Marshal(wantedDeployment)

// Maybe there is a package that provides a function that does this
// Merges wantedDeploymentStateJSON into mergedDeploymentStateJSON 
mergeJSON(mergedDeploymentStateJSON, wantedDeploymentStateJSON)

if !byteSlicesEqual(mergedDeploymentStateJSON, existingDeploymentStateJSON) {
    // Reconcile logic
} else {
    // Skip reconciliation
}
```

This allows us to inherit whatever defaults the current cluster applies (this may vary between k8s and ocp releases) and just makes sure we add our own fields.

WDYT?

omeryahud:
I was interested so I did some research, this might be helpful: https://github.com/imdario/mergo

akrejcir:
I thought about this option. The main difference, is that users will be able to modify the other fields that our operator don't set. I thought that is not what we want, but maybe it is ok. I will look at the resources and think about it.

akrejcir:
@omeryahud I have thought about it and I would keep this PR as is, if you think it is ok.
The other way, where the operator would only set certain fields, would be more complex to write.
I think the operator should not allow users to modify the watched resources.

omeryahud:
Yes, I agree.
It's just too bad that we have to import new defaults code whenever we add a resource

omeryahud:
/hold
I am putting this on hold for now, as there is a larger discussion in Openshift Virtualization on how to strictly reconcile all operator owned resources so we might get new ideas from that discussion

omeryahud:
@akrejcir I think all issues around this got answered, can it be closed?

akrejcir:
Yes, this PR as it currently is, is not needed.

PR #17, opened 2020-10-01
Add common-templates operand to SSP Operator

vatsalparekh:

Add common-templates operand to SSP Operator

Fixes #

**Release note**:

```release-note
Add common-templates operand to SSP Operator
```


internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,89 @@
...
+
+func AddWatchTypesToScheme(s *runtime.Scheme) error {
+	return templatev1.Install(s)
+}
+
+func WatchClusterTypes() []runtime.Object {
+	return []runtime.Object{
+		&rbac.ClusterRole{},
+		&rbac.ClusterRoleBinding{},
+		&templatev1.Template{},
```
akrejcir, line 20: The old operator installs templates to a namespace. The file `common-templates-v0.12.1.yaml` does not contain namespaces, because they depend on SSP resoource. This operator should also create templates in a namespace.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -0,0 +1,80 @@
...
+		namespace = "kubevirt"
+		name      = "test-ssp"
+	)
+
+	var request common.Request
+
+	BeforeEach(func() {
+		s := scheme.Scheme
+		Expect(ssp.AddToScheme(s)).ToNot(HaveOccurred())
+
```
akrejcir, line 39: The templates must also be added to schema here:
```
Expect(AddWatchTypesToScheme(s)).ToNot(HaveOccurred())
```

internal/operands/common-templates/resource.go
```diff
@@ -0,0 +1,136 @@
...
+			Name:      "system:authenticated",
+			Namespace: namespace,
+			APIGroup:  "rbac.authorization.k8s.io",
+		}},
+	}
+}
+
+func newTemplateBundle() []templatev1.Template {
+	filename := "common-templates-" + version + ".yaml"
+	bundle, err := loadTemplates(filename)
```
akrejcir, line 128: The template bundle should be cached, so it is not reloaded on every reconciliation iteration.

go.mod
```diff
@@ -4,13 +4,18 @@ go 1.13
...
+	github.com/openshift/api v0.0.0-20200917102736-0a191b5b9bb0 // release-4.5
 	github.com/operator-framework/operator-lib v0.1.0
-	k8s.io/api v0.18.6
-	k8s.io/apimachinery v0.18.6
-	k8s.io/client-go v0.18.6
+	golang.org/x/net v0.0.0-20201002202402-0a1ea396d57c // indirect
+	k8s.io/api v0.19.2
+	k8s.io/apimachinery v0.19.2
+	k8s.io/client-go v0.19.2
+	k8s.io/klog/v2 v2.3.0 // indirect
```
akrejcir, line 18: These version updates are probably not needed. The `github.com/openshift/api v0.0.0-20200917102736-0a191b5b9bb0 // release-4.5` module should work with existing package versions.

internal/operands/common-templates/common-templates-v0.12.2.yaml
```diff
@@ -0,0 +1,10949 @@
+# Version v0.12.2
```
akrejcir, line 1: Can you move this file outside of the source code directory? Maybe to a `data` directory? The source code directories will probably not be included in the operator container image.

akrejcir (changes requested):

config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -22,89 +22,49 @@ spec:
       description: SSP is the Schema for the ssps API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
akrejcir, line 25: Strange that on my machine `make manifests` generates the text with newlines.
I have `controller-gen` version `v0.3.0`.

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,106 @@
...
+		})
+}
+
+func reconcileTemplates(request *common.Request) error {
+	loadTemplates := func() {
+		filename := filepath.Join(bundleDir, "common-templates-"+version+".yaml")
+		readAndLoadTemplates(filename)
+	}
+	// Only load templates Once
+	loadTemplatesOnce.Do(loadTemplates)
```
akrejcir, line 87: It seems confusing to use `sync.Once` variable from another file. Please either move the variables to this file or move this call to `resource.go`.

internal/operands/common-templates/resource.go
```diff
@@ -0,0 +1,129 @@
...
+	rbac "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"sync"
+)
+
+const (
+	goldenImagesNSname = "kubevirt-os-images"
+	version            = "v0.12.2"
+	bundleDir          = "../../../data/common-templates-bundle/"
```
akrejcir, line 18: This path is relative to what? It may be good to start the operator with working directory at the root of the repo, or at the location of the binary.

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if len(templatesBundle) == 0 {
+		panic(templatesBundle)
+	}
+
+	for _, template := range templatesBundle {
+		err := common.CreateOrUpdateClusterResource(request, &template,
```
akrejcir, line 94: A namespace needs to be specified for the template. Because the namespace can depend on the SSP CR, it should not be cached.

akrejcir (changes requested):

config/crd/bases/ssp.kubevirt.io_ssps.yaml
vatsalparekh: I think it's fine either way

akrejcir:
The `Cleanup` function is missing, to remove cluster resources when the SSP CR is removed.

Cluster resources are not directly owned by the SSP CR, they only have annotation that points to the CR that owns them. The SSP operator needs to remove them when the CR is being removed.

internal/operands/common-templates/resource.go
vatsalparekh: This is relative to the operand files

internal/operands/common-templates/reconcile.go
vatsalparekh: Moved

internal/operands/common-templates/reconcile.go
vatsalparekh: Changed now

internal/operands/common-templates/reconcile_test.go
vatsalparekh: Added

internal/operands/common-templates/reconcile.go
omeryahud: Currently, when HCO deploys the common-templates CR, it does so in the `openshift` namespace, thus templates are deployed in the `openshift` namespace as well. I think our SSP CR should have something like this:
```
spec:
  commonTemplates:
    namepace: $NAMESPACE
```
So we can control in which namespace the templates are deployed.

internal/operands/common-templates/resource.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+var (
+	loadTemplatesOnce sync.Once
+	templatesBundle   []templatev1.Template
+)
+
+func clusterRoleName(role string) string {
+	return "os-images.kubevirt.io" + ":" + role
```
omeryahud, line 27: I would prefer to have resource names as constant values at the top of the file

internal/operands/common-templates/reconcile.go
omeryahud: As I mentioned in my other comment, I think the namespace should be fetched from the SSP CR CT config

internal/operands/common-templates/resource.go
```diff
@@ -0,0 +1,129 @@
...
+	core "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"sync"
+)
+
+const (
+	goldenImagesNSname = "kubevirt-os-images"
+	version            = "v0.12.2"
```
omeryahud, line 17: Maybe instead of hard-coding the version, we should select the latest version found in that directory?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+
+	It("should create common-template resources", func() {
+		Expect(Reconcile(&request)).ToNot(HaveOccurred())
+		Expect(templatesBundle).ToNot(BeNil())
+		for _, template := range templatesBundle {
+			Expect(template).ToNot(BeNil())
+			expectResourceExists(&template, request)
+		}
+	})
```
omeryahud, line 73: Is it possible to replicate this test for the os-images namespace?

omeryahud (changes requested):
Looks pretty good, only some small nits

internal/operands/common-templates/resource.go
vatsalparekh: or have it in CR config?

omeryahud: We can have it in the CR config, but it should default to the latest version when unspecified

internal/operands/common-templates/resource.go
vatsalparekh: Changed now

internal/operands/common-templates/reconcile_test.go
vatsalparekh: Added

api/v1alpha1/ssp_types.go
```diff
@@ -29,13 +29,24 @@ type TemplateValidator struct {
 	Placement lifecycleapi.NodePlacement `json:"placement,omitempty"`
 }
 
+type CommonTemplates struct {
+	// Namespace is the k8s namespace where CommonTemplates should be installed
+	Namespace string `json:"namespace"`
+
+	// Version defines the common-templates release to be installed
+	Version string `json:"version"`
```
akrejcir, line 37: Do we want the version to be configurable? Is it not enough to use a hard-coded constant, that would change as new versions are released?

If we want it to be configurable, a user could change it on an existing resource. There would need to be logic in `reconcileTemplates` method, that would remove old versions of templates and create new versions.

Or maybe we could disallow changing the  `Version` once the CR is created. We would probably need to create a validating webhook for that.

The same applies for the `Namespace` parameter.

internal/operands/common-templates/resource.go
akrejcir: This will probably not work on a deployed operator. The path should be relative to the binary or root of the repo, not the source files. The sources will not be present in the operator image.

akrejcir (changes requested):

internal/operands/common-templates/reconcile_test.go
omeryahud: Can you do the same for the roles please?

omeryahud (commented):
Looks good, still missing functional tests

api/v1alpha1/ssp_types.go
omeryahud: I agree with you that the version field should be dropped and only the latest version should be deployed, I think it is pointless to have it as we are not required to be downgradeable.

omeryahud: And I also think that a webhook should exist to prevent changing the namespace on an existing SSP resource.
I think `operator-sdk` provides a webhook generator

akrejcir: We can add the webhook in a new PR.

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+func Cleanup(request *common.Request) error {
+	for _, obj := range []controllerutil.Object{
+		newGoldenImagesNS(goldenImagesNSname),
+		newViewRole(goldenImagesNSname),
+		newViewRoleBinding(goldenImagesNSname),
+		newEditRole(goldenImagesNSname),
+		//templatesBundle,
```
akrejcir, line 55: If we currently don't want to remove the templates, maybe add a TODO comment, so we don't forget about it later.

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,128 @@
...
+			newRole := newRes.(*rbac.ClusterRole)
+			foundRole := foundRes.(*rbac.ClusterRole)
+			foundRole.Rules = newRole.Rules
+		})
+}
+
+func reconcileTemplates(request *common.Request) error {
+	loadTemplates := func() {
+		filename := filepath.Join(bundleDir, "common-templates-"+Version+".yaml")
+		readAndLoadTemplates(filename)
```
akrejcir, line 104: This function could return the template slice and an error. In case of an error it is ok to panic. After this change all the other panics in the method and below at line 110 can be removed.

akrejcir (changes requested):
A few minor comments since you need to rebase this anyway.

internal/operands/common-templates/reconcile.go
vatsalparekh: @akrejcir `Once.Do` func cannot return any value, so it doesn't make sense to return the values from this func

internal/operands/common-templates/reconcile.go
vatsalparekh: Changed this to append all the templates from the bundle

internal/operands/common-templates/reconcile.go
akrejcir: I meant the `readAndLoadTemplates()` function could return the slice and error.

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,142 @@
...
+
+func (c *commonTemplates) Cleanup(request *common.Request) error {
+	objects := []controllerutil.Object{
+		newGoldenImagesNS(GoldenImagesNSname),
+		newViewRole(GoldenImagesNSname),
+		newViewRoleBinding(GoldenImagesNSname),
+		newEditRole(GoldenImagesNSname),
+	}
+	for _, template := range templatesBundle {
+		objects = append(objects, &template)
```
akrejcir, line 68: Namespace needs to be set for these templates. Otherwise the `Delete` call below will not know from which namespace the template should be removed.

akrejcir (changes requested):

internal/operands/common-templates/resource.go
```diff
@@ -0,0 +1,136 @@
...
+
+func newGoldenImagesNS(namespace string) *core.Namespace {
+	return &core.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		},
+	}
+}
+
+func newViewRole(namespace string) *rbac.ClusterRole {
```
omeryahud, line 60: The `view` role is actually a `Role` and not a `ClusterRole`
https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/roles/KubevirtCommonTemplatesBundle/templates/kubevirt-os-images-viewrole.yaml.j2#L2

omeryahud (changes requested):

internal/operands/common-templates/resource.go
vatsalparekh: Corrected

tests/tests_suite_test.go
```diff
@@ -23,6 +23,7 @@ import (
 const (
 	// TODO - maybe randomize namespace
 	testNamespace             = "ssp-operator-functests"
+	commonTemplatesTestNS     = "openshift"
```
akrejcir, line 26: I think a new namespace should be created for the templates. To not pollute the existing `openshift` namespace.

internal/operands/common-templates/resource.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+const (
+	GoldenImagesNSname = "kubevirt-os-images"
+	bundleDir          = "data/common-templates-bundle/"
+	roleName           = "os-images.kubevirt.io:"
+	Version            = "v0.12.2"
+)
+
+func ClusterRoleName(role string) string {
```
omeryahud, line 21: Maybe just have constants for both combinations? a function feels like an overkill

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,145 @@
...
+	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/operands"
+	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sync"
+)
+
+var (
+	loadTemplatesOnce sync.Once
+	templatesBundle   []templatev1.Template
```
omeryahud, line 19: I don't like accessing this variable from multiple files in the package, I would move it to `resource.go`, and make `readTemplates()` populate it if it's empty, or just return it if it is already initialized

internal/operands/common-templates/reconcile_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+var (
+	log     = logf.Log.WithName("common-templates-operand")
+	operand = GetOperand()
+)
+
+func TestTemplates(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Template Validator Suite")
```
omeryahud, line 28: A copy-paste remnant?

omeryahud (commented):
Looks good, just a few small things

internal/operands/common-templates/reconcile.go
akrejcir: The `readTemplates()` funciton on line 121 could return the loaded template slice, instead of assigning it to the global variable. Then both these variables `loadTemplatesOnce` and `templatesBundle` would be accessed only from one file.

tests/tests_suite_test.go
vatsalparekh: We expect users/hco to create these templates in `openshift` namespace, and this is a test, I think it's fine to pollute for this purpose

internal/operands/common-templates/resource.go
vatsalparekh: Corrected now

internal/operands/common-templates/reconcile.go
vatsalparekh: Changed it this way

omeryahud:
I gave it some thought, and maybe we should rename the common-templates file to `common-templates.yaml`?
We do not plan on supporting downgrades and maintaining the version variable feels redundant.

WDYT?

vatsalparekh:
> I gave it some thought, and maybe we should rename the common-templates file to `common-templates.yaml`?
> We do not plan on supporting downgrades and maintaining the version variable feels redundant.
> 
> WDYT?

While just `common-templates.yaml` makes sense, But having a version gives a little more clarity of which version of common-templates we're using.

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+		})
+}
+
+func reconcileTemplates(request *common.Request) error {
+	loadTemplates := func() {
+		var err error
+		filename := filepath.Join(bundleDir, "common-templates-"+Version+".yaml")
+		templatesBundle, err = readTemplates(filename)
+		if err != nil {
+			request.Logger.Error(err, fmt.Sprintf("Error reading from template file; Templates found \n%v", templatesBundle))
```
akrejcir, line 126: If the `readTemplates` function returns an error, the returned templates bundle is `nil`. Logging the bundle does not make sense.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -0,0 +1,106 @@
...
+	It("should create golden-images namespace", func() {
+		Expect(operand.Reconcile(&request)).ToNot(HaveOccurred())
+		expectResourceExists(newGoldenImagesNS(GoldenImagesNSname), request)
+	})
+	It("should create common-template resources", func() {
+		Expect(operand.Reconcile(&request)).ToNot(HaveOccurred())
+		Expect(templatesBundle).ToNot(BeNil())
+		for _, template := range templatesBundle {
+			template.Namespace = namespace
+			Expect(template).ToNot(BeNil())
```
akrejcir, line 83: The `template` cannot be `nil`, so this line is unnecessary.

tests/commonTemplates_test.go
```diff
@@ -0,0 +1,155 @@
...
+				return apiClient.Get(ctx, client.ObjectKey{
+					Name: res.Name, Namespace: res.Namsespace,
+				}, resource)
+			}, timeout, time.Second).ShouldNot(HaveOccurred())
+		},
+			table.Entry("view role", viewRole),
+			table.Entry("view role binding", viewRoleBinding),
+			table.Entry("testTemplate in custom NS", testTemplate),
+		)
+		table.DescribeTable("recreate after delete", func(res *testResource) {
```
akrejcir, line 89: These two tables called "recreate after delete" can be merged to one table. The code is the same except for assigning namespace in the ObjectKey. For cluster resources, the namespace is `""` so it can be assigned to ObjectKey.

tests/commonTemplates_test.go
```diff
@@ -0,0 +1,155 @@
...
+				}),
+
+			table.Entry("view role binding", viewRoleBinding,
+				func(roleBinding *rbac.RoleBinding) {
+					roleBinding.Subjects = []rbac.Subject{}
+				},
+				func(old *rbac.RoleBinding, new *rbac.RoleBinding) bool {
+					return reflect.DeepEqual(old.Subjects, new.Subjects)
+				}),
+		)
```
akrejcir, line 153: Can you add a `table.Entry` to test modification of a template?

akrejcir (changes requested):

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+			panic(err)
+		}
+		if len(templatesBundle) == 0 {
+			panic(templatesBundle)
```
omeryahud, line 130: Please provide a clear error here that states that no templates were found

tests/tests_suite_test.go
omeryahud: Please note that we eventually want to execute our tier-1 suite on CNV clusters downstream, polluting a production namespace during testing is not a good idea, we need to create our own testing NS and remove it when we are done

omeryahud (changes requested):
Looks good, 2 small nits

omeryahud:
> > I gave it some thought, and maybe we should rename the common-templates file to `common-templates.yaml`?
> > We do not plan on supporting downgrades and maintaining the version variable feels redundant.
> > WDYT?
> 
> @omeryahud While just `common-templates.yaml` makes sense, But having a version gives a little more clarity of which version of common-templates we're using.

The version is stated in the `common-templates.yaml` file itself, but that's not really a deal breaker at this point, lets leave it as-is and come back to it if the need rises.

tests/tests_suite_test.go
vatsalparekh: Added a test namespace for templates

internal/operands/common-templates/reconcile.go
vatsalparekh: Removed

internal/operands/common-templates/reconcile_test.go
vatsalparekh: Removed

tests/commonTemplates_test.go
vatsalparekh: Merged now

tests/commonTemplates_test.go
vatsalparekh: Added

internal/operands/common-templates/reconcile.go
vatsalparekh: Added

internal/operands/common-templates/reconcile.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (c *commonTemplates) Cleanup(request *common.Request) error {
+	objects := []controllerutil.Object{
+		newGoldenImagesNS(GoldenImagesNSname),
+		newViewRole(GoldenImagesNSname),
+		newViewRoleBinding(GoldenImagesNSname),
+		newEditRole(),
+	}
+	namespace := request.Instance.Spec.CommonTemplates.Namespace
+	for _, template := range templatesBundle {
```
omeryahud, line 69: When you use 
```
for _, template := range remplatesBundle {
    ...
}
```
A single variable `template` is defined in the scope of the loop (not a new one for each iteration), which means that `&template` returns the same address for all iterations. This eventually causes you to append the same address X times to `objects`, which means you are trying to delete the same template X times (the last template that was assigned to `template`.

To solve this, you can use:
```
for index := range templateBundle {
    ...
    objects = append(objects, &templatesBundle[index])
}
```

omeryahud (changes requested):

internal/operands/common-templates/reconcile.go
vatsalparekh: This change seems to have fixed the cleanup!

akrejcir:
Looks good. Before merging, can you squash all related commits together?

omeryahud (approved):
/lgtm

PR #18, opened 2020-10-06
Update README.md with instructions how to run tests

akrejcir:
Added new information to the `README.md`.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,12 +75,12 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
-# Build the docker image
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -75,12 +75,12 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
-# Build the docker image
-docker-build: unittest
+# Build the container image
+container-build: unittest
 	docker build . -t ${IMG}
```
vatsalparekh, line 80: A side note, let's make sure this works with `podman` also, if it's available on the environment 
We're doing it already in the test container build part

PR #22, opened 2020-10-21
Create interface for operands

akrejcir:
This change makes it easier to understand which methods should be implemented by operands.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/internal/operands/template-validator/reconcile.go
+++ b/internal/operands/template-validator/reconcile.go
@@ -13,9 +13,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/operands"
 )
 
-func WatchTypes() []runtime.Object {
+type templateValidator struct{}
+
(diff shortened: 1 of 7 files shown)


internal/operands/template-validator/reconcile.go
```diff
@@ -13,25 +13,32 @@ import (
...
 
 	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/operands"
 )
 
-func WatchTypes() []runtime.Object {
+type templateValidator struct{}
+
+func (t *templateValidator) AddWatchTypesToScheme(*runtime.Scheme) error {
+	return nil
```
omeryahud, line 22: At first glance this looks like a mistake, I assume this function returns `nil` because no additional resources need to be added to the scheme, please add a comment to state that

omeryahud (changes requested):

internal/operands/template-validator/reconcile.go
akrejcir: Correct. Resources from `k8s.io/api` are added by default. Ok I will add a comment.

PR #23, opened 2020-10-21
Add status phase and conditions

akrejcir:
Added `status` field to the SSP CR.
`status.phase` can be used to check if the SSP operator finished deployment of all dependent resources.
`status.conditions` give more information about the status.

**Release note**:
```release-note
Added status to the SSP resource.
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -18,14 +18,13 @@ package controllers
 
 import (
 	"context"
-	"reflect"
-
 	"github.com/go-logr/logr"
 	libhandler "github.com/operator-framework/operator-lib/handler"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"reflect"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -56,6 +55,9 @@ type SSPReconciler struct {
 	client.Client
 	Log    logr.Logger
 	Scheme *runtime.Scheme
+
+	LastSspSpec      ssp.SSPSpec
+	SubresourceCache common.VersionCache
 }
 
 // +kubebuilder:rbac:groups=ssp.kubevirt.io,resources=ssps,verbs=get;list;watch;create;update;patch;delete
@@ -66,17 +68,11 @@ type SSPReconciler struct {
 func (r *SSPReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	reqLogger := r.Log.WithValues("ssp", req.NamespacedName)
 
-	sspRequest := &common.Request{
-		Request: req,
-		Client:  r,
-		Scheme:  r.Scheme,
-		Context: context.Background(),
-		Logger:  reqLogger,
-	}
+	ctx := context.Background()
 
 	// Fetch the SSP instance
 	instance := &ssp.SSP{}
-	err := r.Get(sspRequest.Context, req.NamespacedName, instance)
+	err := r.Get(ctx, req.NamespacedName, instance)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
@@ -88,7 +84,17 @@ func (r *SSPReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 
-	sspRequest.Instance = instance
+	r.clearCacheIfNeeded(instance)
+
+	sspRequest := &common.Request{
+		Request:              req,
+		Client:               r,
+		Scheme:               r.Scheme,
+		Context:              ctx,
+		Instance:             instance,
+		Logger:               reqLogger,
+		ResourceVersionCache: r.SubresourceCache,
+	}
 
 	updated, err := initialize(sspRequest)
 	if updated || err != nil {
@@ -99,7 +105,11 @@ func (r *SSPReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	if isBeingDeleted(sspRequest.Instance) {
 		err := cleanup(sspRequest)
-		return ctrl.Result{}, err
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		r.clearCache()
+		return ctrl.Result{}, nil
 	}
 
 	for _, operand := range sspOperands {
@@ -109,6 +119,18 @@ func (r *SSPReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	return ctrl.Result{}, nil
+}
+
+func (r *SSPReconciler) clearCacheIfNeeded(sspObj *ssp.SSP) {
+	if !reflect.DeepEqual(r.LastSspSpec, sspObj.Spec) {
+		r.SubresourceCache = common.VersionCache{}
+		r.LastSspSpec = sspObj.Spec
+	}
+}
+
+func (r *SSPReconciler) clearCache() {
+	r.LastSspSpec = ssp.SSPSpec{}
+	r.SubresourceCache = common.VersionCache{}
 }
 
 func isBeingDeleted(object metav1.Object) bool {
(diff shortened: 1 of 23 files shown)


controllers/ssp_controller.go
```diff
@@ -84,67 +85,227 @@ func (r *SSPReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	return allStatuses, nil
+}
+
+func updateStatus(request *common.Request, statuses []common.ResourceStatus) error {
+	notAvailable := make([]common.ResourceStatus, 0, len(statuses))
+	progressing := make([]common.ResourceStatus, 0, len(statuses))
+	degraded := make([]common.ResourceStatus, 0, len(statuses))
+	for _, status := range statuses {
+		if status.NotAvailable != nil {
+			notAvailable = append(notAvailable, status)
```
omeryahud, line 204: Are `Available`, `Progressing` and `Degraded` mutually exclusive?
If not, what happens in the case of 2 `Progressing` resources, is `Available` being reported on the CR because `len(notAvailable) = 0`?

internal/operands/template-validator/reconcile.go
```diff
@@ -114,15 +109,40 @@ func reconcileService(request *common.Request) error {
...
+	return common.CreateOrUpdateResourceWithStatus(request,
 		deployment,
 		func(newRes, foundRes controllerutil.Object) {
-			foundRes.(*apps.Deployment).Spec = newRes.(*apps.Deployment).Spec
+			foundDep := foundRes.(*apps.Deployment)
+			newDep := newRes.(*apps.Deployment)
+
+			// TODO -- may not be needed with the resVersion
+			// Copy deprecated service account, so it will be ignored
+			newDep.Spec.Template.Spec.DeprecatedServiceAccount = foundDep.Spec.Template.Spec.DeprecatedServiceAccount
```
omeryahud, line 125: Can you validate this please?

internal/common/resource.go
```diff
@@ -9,29 +9,64 @@ import (
...
-	if err != nil {
-		return err
+type ResourceStatus struct {
+	Resource     controllerutil.Object
+	Progressing  StatusMessage
+	NotAvailable StatusMessage
+	Degraded     StatusMessage
+}
+
+func ReconcileMultiple(reqest *Request, funcs ...func(*Request) (ResourceStatus, error)) ([]ResourceStatus, error) {
```
omeryahud, line 21: I think renaming this to `Reconcile()` with `...func()` as an argument is more readable

internal/operands/metrics/reconcile.go
```diff
@@ -23,12 +23,16 @@ func (m *metrics) WatchClusterTypes() []runtime.Object {
 	return nil
 }
 
-func (m *metrics) Reconcile(request *common.Request) error {
-	return common.CreateOrUpdateResource(request,
+func (m *metrics) Reconcile(request *common.Request) ([]common.ResourceStatus, error) {
+	status, err := common.CreateOrUpdateResource(request,
```
omeryahud, line 27: Not sure if it's in the scope of this PR, but maybe use `Reconcile()` with a `reconcilePrometheusRule()` function?
For alignment with the rest of the operands

controllers/ssp_controller.go
akrejcir: They are related but not mutually exclusive. These are the meanings:
- `Available` - the resource is available to perform the expected task. For example a Deployment that has at least 1 pod running.
- `Progressing` - the resource has not finished deploying, but it can be available. For example a Deployment where some pods are still starting.
- `Degraded` - the resource is not completely deployed. This can mean that it is still `progressing`, but it can also mean that there was an error during deployment.

Currently on SSP CR, `progressing` and `degraded` always have the same value.

internal/operands/template-validator/reconcile.go
akrejcir: Removed

internal/common/resource.go
akrejcir: I have renamed it to `CollectResourceStatus()`, because the function does not do any reconciliation.

internal/operands/metrics/reconcile.go
akrejcir: Done

controllers/ssp_controller.go
omeryahud: Ah, I did not know that `Available` was reported when a Deployment was not fully rolled out, that's good to know.
So, a situation with all `Available`, `Degraded`, `Progressing` is a valid one?

akrejcir: That is how I understood it, so I implemented it that way. A situation when all these conditions are true happens if some pods for `template-validator` are running, but not all of them.

internal/operands/template-validator/reconcile.go
```diff
@@ -114,15 +109,33 @@ func reconcileService(request *common.Request) error {
...
 	deployment := newDeployment(request.Namespace, validatorSpec.Replicas, image)
 	addPlacementFields(deployment, &validatorSpec.Placement)
-	return common.CreateOrUpdateResource(request,
+	return common.CreateOrUpdateResourceWithStatus(request,
 		deployment,
 		func(newRes, foundRes controllerutil.Object) {
 			foundRes.(*apps.Deployment).Spec = newRes.(*apps.Deployment).Spec
+		},
+		func(res controllerutil.Object) common.ResourceStatus {
+			dep := res.(*apps.Deployment)
```
omeryahud, line 123: Why not just read the `Status` of the deployment itself?
I think deployments provide `Available`, `Progressing` and `Degraded` conditions

akrejcir: The `progressing` condition for Deployment is `true` even if all pods are deployed successfully. That is not the same as our `progressing` condition, which should be `false` when everything is deployed.

For now I would keep this code.

tests/validator_test.go
```diff
@@ -271,6 +326,34 @@ var _ = Describe("Template validator", func() {
 					podSpec.Tolerations == nil
 			}, timeout, 1*time.Second).Should(BeTrue())
 		})
+
+		// TODO - this test may fail if the operator is too slow and does not notice Deployment state when
+		//        where not all pods are running
+		It("[test_id: TODO]should set available condition when at least one validator pod is running", func() {
+			watch, err := StartWatch(sspListerWatcher)
```
omeryahud, line 333: I would prefer if this kind of test was `Pending()` instead of `It()` until we figure out how to properly test it or remove it

tests/watcher.go
```diff
@@ -0,0 +1,126 @@
...
+		lastVersion:   list.GetResourceVersion(),
+	}
+	go watch.startWatch()
+	return watch, nil
+}
+
+type sspWatch struct {
+	listerWatcher cache.ListerWatcher
+	stopCh        chan struct{}
+	updateCh      chan *sspv1alpha1.SSP
```
omeryahud, line 47: We can denote this as a receive only channel

omeryahud (changes requested):
Looks great, two minor changes, other than those, lgtm

tests/validator_test.go
akrejcir: Done

tests/watcher.go
akrejcir: Yes, but it is not needed. The test code uses only the `SspWatch` interface, which provides only the receive channel.

internal/operands/node-labeller/reconcile.go
```diff
@@ -109,7 +104,7 @@ func reconcileConfigMap(request *common.Request) error {
 		})
 }
 
-func reconcileDaemonSet(request *common.Request) error {
+func reconcileDaemonSet(request *common.Request) (common.ResourceStatus, error) {
 	nodeLabellerSpec := &request.Instance.Spec.NodeLabeller
 	daemonSet := newDaemonSet(request.Namespace)
 	addPlacementFields(daemonSet, &nodeLabellerSpec.Placement)
```
omeryahud, line 110: This function should return `common.CreateOrUpdateResourceWithStatus()` because `DaemonSet`s support conditions and deploy pods

omeryahud (changes requested):
One small thing, other than that lgtm

internal/operands/node-labeller/reconcile.go
akrejcir: I missed this. Updated.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
omeryahud, line 3: I suppose this was added by mistake

akrejcir: Removed

omeryahud (approved):
/lgtm

PR #29, opened 2020-11-03
Update Dockerfile and add RBAC rules needed to run operator in cluster

akrejcir:
```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,10 @@ COPY internal/ internal/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
+COPY data/ data/
+USER 1000
 
 ENTRYPOINT ["/manager"]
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -18,11 +18,10 @@ COPY internal/ internal/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM registry.access.redhat.com/ubi8/ubi-minimal
```
omeryahud, line 21: Why the change of the base image?

config/manager/manager.yaml
```diff
@@ -32,8 +32,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
+            memory: 100Mi
```
omeryahud, line 35: Please don't specify resource limits and requests before we have a proven performance issue (please remove them completely, even the default ones).

We did it before with `template-validator`, and it just made our life difficult when deploying on smaller clusters and debugging.

omeryahud (changes requested):
Two small things, other than that lgtm

Dockerfile
akrejcir: Other kubevirt projects use `ubi8` as a base for containers.

omeryahud: Ah, very well, nice catch

config/manager/manager.yaml
akrejcir: Ok, I will remove them. I hit the OOM error with the default limits. On my machine the controller need about 56 Mi.

akrejcir: Done

omeryahud (approved):
/lgtm

PR #31, opened 2020-11-04
Add validations to the API and webhook

akrejcir:
Added validations to the API and webhook

In API:
- `templateValidator.replicas` cannot be negative
- `commonTemplates.namespace` must be valid namespace

In webhook:
- `commonTemplates.namespace` cannot be updated on existing CR

```release-note
NONE
```


omeryahud (approved):
/lgtm

PR #35, opened 2020-11-06
added csv-generator

ksimon1:
added csv-generator

**Release note**:
```NONE

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,13 @@ COPY internal/ internal/
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal
+LABEL org.kubevirt.hco.csv-generator.v1="/usr/bin/csv-generator"
+
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY data/ data/
 USER 1000
 
+# Copy csv generator
+COPY hack/csv-generator.sh /usr/bin/csv-generator
 ENTRYPOINT ["/manager"]
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@
 VERSION ?= 0.0.1
 # Default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
+#operator-sdk version
+OPERATOR_SDK_VERSION ?= v1.0.0
+
 # Options for 'bundle-build'
 ifneq ($(origin CHANNELS), undefined)
 BUNDLE_CHANNELS := --channels=$(CHANNELS)
@@ -79,8 +82,22 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+operator-sdk:
+	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
+	chmod 0755 operator-sdk
+
+olm-catalog: manifests
+	operator-sdk generate kustomize manifests -q
+	cd config/manager && $(KUSTOMIZE) edit set image controller=REPLACE_IMAGE:TAG
+	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
+	mkdir -p data/crd
+	mkdir -p data/olm-catalog
+	./hack/update-csv.py bundle/manifests/ssp-operator.clusterserviceversion.yaml > data/olm-catalog/ssp-operator.clusterserviceversion.yaml
+	cp bundle/manifests/ssp.kubevirt.io_ssps.yaml data/crd/ssp.kubevirt.io_ssps.yaml
+	rm -r bundle.Dockerfile
+
 # Build the container image
-container-build: unittest
+container-build: unittest olm-catalog
 	docker build . -t ${IMG}
 
 # Push the container image
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: REPLACE_IMAGE
+  newTag: TAG
--- a/config/manifests/bases/ssp-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/ssp-operator.clusterserviceversion.yaml
@@ -1,0 +1,67 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: ClusterServiceVersion
+metadata:
+  annotations:
+    alm-examples: '[]'
+    capabilities: Basic Install
+    categories: Openshift Optional
+    containerImage: REPLACE_IMAGE:TAG
+    description: Manages KubeVirt addons for Scheduling, Scale, Performance
+    operators.operatorframework.io/builder: operator-sdk-v1.0.0
+    operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
+  name: ssp-operator.vX.Y.Z
+  namespace: kubevirt
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - description: SSP is the Schema for the ssps API
+      displayName: SSP
+      kind: SSP
+      name: ssps.ssp.kubevirt.io
+      version: v1alpha1
+  description: KubeVirt Schedule, Scale and Performance Operator
+  displayName: ssp-operator
+  icon:
+  - base64data: ""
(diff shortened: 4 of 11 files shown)


Dockerfile
```diff
@@ -19,9 +19,18 @@ COPY internal/ internal/
...
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal
+LABEL org.kubevirt.hco.csv-generator.v1="/usr/bin/csv-generator"
+
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY data/ data/
 USER 1000
 
+WORKDIR /
```
akrejcir, line 29: These commands are duplicated.

bundle/manifests/ssp-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,221 @@
+apiVersion: operators.coreos.com/v1alpha1
```
akrejcir, line 1: Did you edit anything in the `bundle` directory? If it only contains generated files, we can add it to `.gitignore` and not commit it.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: REPLACE_IMAGE
+  newTag: TAG
```
akrejcir, line 8: I don't like that we commit these placeholders. But I'm not sure how to avoid it.

hack/update-csv.py
```diff
@@ -0,0 +1,71 @@
...
+        'operated-by': 'ssp-operator',
+    },
+    'selector': {
+        'matchLabels': {
+            'alm-owner-kubevirt': 'ssp-operator',
+            'operated-by': 'ssp-operator',
+        },
+    },
+    'version': 'vPLACEHOLDER_CSV_VERSION'
+}
```
akrejcir, line 48: It looks like many of these changes could be applied directly to `config/manifests/base/ssp-operator.clusterserviceversion.yaml`.
Then this code can be simplified.

Makefile
```diff
@@ -119,6 +119,13 @@ else
 KUSTOMIZE=$(shell which kustomize)
 endif
 
+olm-catalog: manifests
+	operator-sdk generate kustomize manifests -q
+	cd config/manager && $(KUSTOMIZE) edit set image controller=REPLACE_IMAGE:TAG
+	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) --input-dir bundle
+	./hack/update-csv.py bundle/manifests/ssp-operator.clusterserviceversion.yaml > config/olm-catalog/ssp-operator.clusterserviceversion.yaml
```
akrejcir, line 126: I think it would be better if the resulting `.yaml` file is not in the `config` directory. Maybe we can but it to `data` directory since it has to be copied to the container image.

I would also put it to `.gitignore` and generate it on `make container-build`.

akrejcir (changes requested):

akrejcir:
I have tried to run the `csv-generator` and I get this error:
```
$ docker run \
>   --rm \
>   --entrypoint=/usr/bin/csv-generator \
>   quay.io/akrejcir/ssp-operator:test \
>   --csv-version=2.2.0 \
>   --namespace=openshift-cnv \
>   --operator-image=quay.io/akrejcir/ssp-operator:test \
>   --dump-crds
cp: cannot stat '//config/olm-catalog/ssp-operator.clusterserviceversion.yaml': No such file or directory
```

config/manager/kustomization.yaml
ksimon1: I am not sure if this file is necessary. Operator sdk generates lots of files. These placeholders are needed because they are used and replaced by HCO

akrejcir: The file is necessary, it contains the image and tag that will be used in Deployment manifest.
I don't like that there are multiple makefile targets  that write different data here.
`make olm-catalog` writes these placeholders, but `make deploy` and `make bundle` write the `${IMG}` variable.

But for it is ok for now. Maybe we can solve it later.

Dockerfile
```diff
@@ -19,9 +19,14 @@ COPY internal/ internal/
...
+LABEL org.kubevirt.hco.csv-generator.v1="/usr/bin/csv-generator"
+
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY data/ data/
 USER 1000
 
+# Copy csv generator and manifests
+COPY hack/csv-generator.sh /usr/bin/csv-generator
+COPY config/crd/bases /data/crd
```
akrejcir, line 31: Instead of copying the base CRD, it would be better to copy the resulting CRD from `bundle/manifests`. We currently don't modify the base one, but maybe we will add some kustomize rules in the future.

Makefile
```diff
@@ -79,8 +82,19 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+operator-sdk:
+	curl -JL https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk-$(OPERATOR_SDK_VERSION)-x86_64-linux-gnu -o operator-sdk
+	chmod 0755 operator-sdk
```
akrejcir, line 87: Here you could assign the `operator-sdk` to a variable, and use it in the other make targets.
```
OPERATOR_SDK=./operator-sdk 
```

Makefile
```diff
@@ -79,8 +82,19 @@ vet:
...
+
+olm-catalog: manifests
+	operator-sdk generate kustomize manifests -q
+	cd config/manager && $(KUSTOMIZE) edit set image controller=REPLACE_IMAGE:TAG
+	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
```
akrejcir, line 92: This line can use `$(OPERATOR_SDK)` to call the sdk downloaded by the makefile. Similarly for other uses of the sdk.

config/manifests/bases/ssp-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,67 @@
...
+  annotations:
+    alm-examples: '[]'
+    capabilities: Basic Install
+    categories: Openshift Optional
+    containerImage: REPLACE_IMAGE:TAG
+    description: Manages KubeVirt addons for Scheduling, Scale, Performance
+    operators.operatorframework.io/builder: operator-sdk-v1.0.0
+    operators.operatorframework.io/project_layout: go.kubebuilder.io/v2
+  name: ssp-operator.vX.Y.Z
+  namespace: placeholder
```
akrejcir, line 13: This namespace can be changed to `kubevirt`.

data/olm-catalog/ssp-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,403 @@
+apiVersion: operators.coreos.com/v1alpha1
```
akrejcir, line 1: This file does not need to be commited to the repo. It will be generated by makefile when it is needed.

hack/update-csv.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+import logging
+import sys
+import yaml
+
+_VERSIONED_NAME = 'ssp-operator.vPLACEHOLDER_CSV_VERSION'
+_NAMESPACE = 'kubevirt'
+_SPEC = {
+    'version': 'vPLACEHOLDER_CSV_VERSION'
```
akrejcir, line 10: I think the version should be `PLACEHOLDER_CSV_VERSION`, without the 'v' prefix.

hack/update-csv.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
+
+import logging
+import sys
+import yaml
+
+_VERSIONED_NAME = 'ssp-operator.vPLACEHOLDER_CSV_VERSION'
+_NAMESPACE = 'kubevirt'
```
akrejcir, line 8: This change is not needed. The namespace can be directly set in the base CSV.

akrejcir (changes requested):

hack/update-csv.py
ksimon1: This depends on HCO if it sends v prefix or not

akrejcir: The `spec.version` field should contain just the semantic version, without the 'v' prefix. If HCO adds it, then it is probably a bug. Look at the generated `bundle/manifests/ssp-operator.clusterserviceversion.yaml`.

akrejcir: Ok. Ignore my previous comment. I thought that the `spec.version` must not start with 'v', because it does not in the generated csv in `bundle/manifests/ssp-operator.clusterserviceversion.yaml`. But I found that the API code that parses semantic version supports both with and without the 'v' prefix.

Makefile
ksimon1: I think we don't have to change whole makefile only because travis.

omeryahud:
I found an issue that a divider `---` is missing between the CSV and the CRDs when `--dumpcrds` is used.

PR #38, opened 2020-11-11
Add RBAC tests for OS Images

vatsalparekh:

Add RBAC tests for OS Images from the old operator functests https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/functests/12-test-os-imaes-namespace_rbac.sh
Fixes #

**Release note**:

```release-note
Add RBAC tests for OS Images
```

### DIFF
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -350,7 +350,7 @@ var _ = Describe("Common templates", func() {
 						Namespace: strategy.GetNamespace(),
 					},
 				}
-				regularSAFullName = fmt.Sprintf("serviceaccount:%s:%s", regularSA.GetNamespace(), regularSA.GetName())
+				regularSAFullName = fmt.Sprintf("system:serviceaccount:%s:%s", regularSA.GetNamespace(), regularSA.GetName())
 
 				Expect(apiClient.Create(ctx, regularSA)).ToNot(HaveOccurred(), "creation of regular service account failed")
 				Expect(apiClient.Get(ctx, getResourceKey(regularSA), regularSA)).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


omeryahud:
I think you can replicate the usage of `oc auth can-i` with the rest client, see here:

- https://github.com/kubevirt/kubevirt/blob/ed772f3aad83a79e719d2fd66fb5bba65ccee902/vendor/kubevirt.io/containerized-data-importer/pkg/clone/auth.go#L62

- https://github.com/kubevirt/kubevirt/blob/ed772f3aad83a79e719d2fd66fb5bba65ccee902/vendor/kubevirt.io/containerized-data-importer/pkg/clone/auth.go#L81

tests/access_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	commonTemplates "kubevirt.io/ssp-operator/internal/operands/common-templates"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = FDescribe("[rfe_id:500]Golden Images RBAC", func() {
```
omeryahud, line 15: You forgot to unfocus :)

omeryahud:
Looks much better now without messing around with binaries.
Would you consider adding those tests to `commonTemplates_test.go` instead of a new file? I think it's better to group an operand's test cases together in the same file

tests/access_test.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	},
+		table.Entry("[test_id:4775]: ServiceAccounts with view role can view PVCs",
+			&authorization.ResourceAttributes{Verb: "get", Resource: "pvc"}, false),
+		table.Entry("[test_id:4776]: ServiceAccounts with only view role cannot create PVCs",
+			&authorization.ResourceAttributes{Verb: "create", Resource: "pvc"}, true),
+		table.Entry("[test_id:4846]: ServiceAccounts with only view role cannot delete PVCs",
+			&authorization.ResourceAttributes{Verb: "delete", Resource: "pvc"}, true),
+		table.Entry("[test_id:4875]: ServiceAccounts with only view role cannot view DVs",
+			&authorization.ResourceAttributes{Verb: "get", Resource: "datavolume"}, false),
```
akrejcir, line 62: This test is called "ServiceAccounts with only view role CANNOT view DVs", but the test code expects success.

tests/tests_common_test.go
```diff
@@ -117,3 +119,15 @@ func isStatusDeployed(obj *v1alpha1.SSP) bool {
 		progressing.Status == core.ConditionFalse &&
 		degraded.Status == core.ConditionFalse
 }
+
+func CanIAccess(sar *authorization.SubjectAccessReview) (bool, error) {
```
akrejcir, line 123: This function can be inlined. It only calls another function and returns its result directly.

akrejcir (changes requested):

akrejcir:
The e2e tests are failing because of some timeout, but you have also forgot to unfocus the test that you have added. When a test is focused, `go test` will return a non-zero value, so the CI tests will fail anyway.

akrejcir:
I have retriggered the tests too soon. This PR needs to be rebased.

tests/access_test.go
```diff
@@ -0,0 +1,158 @@
...
+						"system:serviceaccounts:" + testServiceAccountNS,
+						"system:authenticated",
+					},
+					ResourceAttributes: resourceAttr,
+				},
+			}
+			sar.Spec.ResourceAttributes.Namespace = commonTemplates.GoldenImagesNSname
+			resultSar, err := SendSubjectAccessReview(&sar)
+			result := resultSar.Status.Allowed
+			Expect(err).ToNot(HaveOccurred())
```
akrejcir, line 57: Please swap these 2 lines:
```
result := resultSar.Status.Allowed
```

It is a golang convention, that if `err` is not `nil`, then the returned value is `nil`. So ` resultSar.Status.Allowed` would cause NPE.

tests/access_test.go
```diff
@@ -0,0 +1,158 @@
...
+						ResourceAttributes: resourceAttr,
+					},
+				}
+				sar.Spec.ResourceAttributes.Namespace = commonTemplates.GoldenImagesNSname
+				allowed, err := SendSubjectAccessReview(sar)
+				Expect(err).ToNot(HaveOccurred())
+				if expectedFail {
+					Expect(allowed).To(BeFalse(), "Expected permission to NOT be allowed")
+				} else {
+					Expect(allowed).To(BeTrue(), "Expected permission to be allowed")
```
akrejcir, line 141: Type of `allowed` is not `bool`, but `*v1.SubjectAccessReview `.

tests/tests_common_test.go
```diff
@@ -117,3 +119,11 @@ func isStatusDeployed(obj *v1beta1.SSP) bool {
 		progressing.Status == core.ConditionFalse &&
 		degraded.Status == core.ConditionFalse
 }
+
+func SendSubjectAccessReview(sar *authorization.SubjectAccessReview) (*authorization.SubjectAccessReview, error) {
```
akrejcir, line 123: Can you inline this function? Or simplify it to:
```
func SendSubjectAccessReview(sar *authorization.SubjectAccessReview) (*authorization.SubjectAccessReview, error) {
	return coreClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, sar, metav1.CreateOptions{})
}
```

akrejcir (changes requested):
A few minor comments.

tests/commonTemplates_test.go
```diff
@@ -431,7 +431,200 @@ var _ = Describe("Common templates", func() {
...
+							APIGroup: "rbac.authorization.k8s.io",
+						},
+					}
+					Expect(apiClient.Create(ctx, editObj)).ToNot(HaveOccurred(), "Failed to create RoleBinding")
+				})
+				table.DescribeTable("should verify resource permissions", expectUserCan,
+					table.Entry("[test_id:4774]: ServiceAcounts with edit role can create PVCs",
+						&authv1.SubjectAccessReviewSpec{
+							User:   regularSAFullName,
+							Groups: sasGroup,
```
omeryahud, line 545: I think CI fails because you specify `Groups`.
According to the documentation, this SAR will check if `regularSA` received the appropriate `create` permission from the `system:serviceaccounts` group, which is not the case because there is no edit permissions for `system:serviceaccounts`, and the `RoleBinding` you created refers to `regularSA`. So `Groups` should be empty:
```
	// User is the user you're testing for.
	// If you specify "User" but not "Groups", then is it interpreted as "What if User were not a member of any groups
	// +optional
	User string `json:"user,omitempty" protobuf:"bytes,3,opt,name=user"`
	// Groups is the groups you're testing for.
	// +optional
	Groups []string `json:"groups,omitempty" protobuf:"bytes,4,rep,name=groups"`
	// Extra corresponds to the user.Info.GetExtra() method from the authenticator.  Since that is input to the authorizer
	// it needs a reflection here.
```
Also, maybe change `regularSA` to `privilegedSA` in this context for readability

tests/tests_common_test.go
```diff
@@ -220,3 +220,15 @@ func expectUserCan(sars *authv1.SubjectAccessReviewSpec) {
 			sars.ResourceAttributes.Subresource, sars.ResourceAttributes.Name, sars.ResourceAttributes.Group,
 			sars.ResourceAttributes.Version, sars.ResourceAttributes.Namespace))
 }
+
+func expectUserCannot(sars *authv1.SubjectAccessReviewSpec) {
+	sar, err := coreClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, &authv1.SubjectAccessReview{
+		Spec: *sars,
+	}, metav1.CreateOptions{})
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	ExpectWithOffset(1, sar.Status.Allowed).To(BeFalse(),
```
omeryahud, line 229: I think checking for the `Denied` field is safer

vatsalparekh: Actually not
`Denied` is option and can be flaky even when `Allowed` is false

akrejcir:
The flaking test will be fixed by: #104

tests/commonTemplates_test.go
```diff
@@ -431,7 +431,196 @@ var _ = Describe("Common templates", func() {
...
+							Resource:  "datavolumes",
+						},
+					}),
+			)
+			table.DescribeTable("regular service account PVC RBAC", expectUserCan,
+				table.Entry("[test_id:4775]: ServiceAccounts with view role can view PVCs",
+					&authv1.SubjectAccessReviewSpec{
+						User:   regularSAFullName,
+						Groups: sasGroup,
+						ResourceAttributes: &authv1.ResourceAttributes{
```
omeryahud, line 481: Is `Version` missing here?

tests/commonTemplates_test.go
```diff
@@ -431,7 +431,196 @@ var _ = Describe("Common templates", func() {
...
+							Name:     commonTemplates.EditClusterRoleName,
+							APIGroup: "rbac.authorization.k8s.io",
+						},
+					}
+					Expect(apiClient.Create(ctx, editObj)).ToNot(HaveOccurred(), "Failed to create RoleBinding")
+				})
+				table.DescribeTable("should verify resource permissions", expectUserCan,
+					table.Entry("[test_id:4774]: ServiceAcounts with edit role can create PVCs",
+						&authv1.SubjectAccessReviewSpec{
+							User:   "system:serviceaccount:ssp-operator-functests:regular-sa",
```
omeryahud, line 552: Shouldn't this be `regularSAFullName`?

vatsalparekh: I'm seeing that the nested `Context ` block is somehow making the value of `regularSAFullName` to be `""`
Probably a bug

tests/commonTemplates_test.go
vatsalparekh: Added

akrejcir:
There is a bug in CDI. The deployment tries to use image `quay.io/kubevirtcdi-operator` instead of `quay.io/kubevirt/cdi-operator`.

PR #41, opened 2020-11-12
Functests wait for operator deployment

akrejcir:
This PR fixes a race condition, when the deployment of the operator is too slow.

```release-note
NONE
```

### DIFF
--- a/internal/common/environment.go
+++ b/internal/common/environment.go
@@ -46,7 +46,7 @@ func GetNodeLabellerImages() NodeLabellerImages {
 	const kvmCpuNfdDefaultImage = "quay.io/kubevirt/cpu-nfd-plugin"
 
 	const libvirtDefaultVersion = "v0.21.0"
-	const libvirtDefaultImage = "kubevirt/virt-launcher"
+	const libvirtDefaultImage = "docker.io/kubevirt/virt-launcher"
 
 	return NodeLabellerImages{
 		NodeLabeller: envOrDefault(kubevirtNodeLabellerImage, kubevirtNodeLabellerDefaultImage+":"+kubevirtNodeLabellerDefaultVersion),
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -2,6 +2,8 @@ package tests
 
 import (
 	"context"
+	"os"
+	"strconv"
 	"testing"
 	"time"
 
@@ -31,7 +33,7 @@ const (
 	// TODO - maybe randomize namespace
 	testNamespace             = "ssp-operator-functests"
 	commonTemplatesTestNS     = "ssp-operator-functests-templates"
-	timeout                   = 180 * time.Second
+	timeout                   = 8 * time.Minute
 	templateValidatorReplicas = 1
 )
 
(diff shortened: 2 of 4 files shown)


tests/tests_suite_test.go
```diff
@@ -31,7 +31,7 @@ const (
 	// TODO - maybe randomize namespace
 	testNamespace             = "ssp-operator-functests"
 	commonTemplatesTestNS     = "ssp-operator-functests-templates"
-	timeout                   = 180 * time.Second
+	timeout                   = 90 * time.Second
```
vatsalparekh, line 34: I feel 90 might be little less, because some test machines can be very slow at times

akrejcir: Changed back. Previously the 180s was problematic, because all tests were timing out in `BeforeEach`.
We have 57 tests, so it took 2 hours and 51 minutes until all tests finished.

With this change, there is a short circuit when waiting for deployment, so it will fail fast.

omeryahud:
If this is to tackle the error you posted in the group chat a few days ago, I think the equivalent of 
`oc wait --for=condition=Available --timeout=600s -n $NAMESPACE deployment/kubevirt-ssp-operator`
should be enough, maybe an `Eventually()` to make sure the *operator's* deployment is ready, and then wait for the `Deployed` condition on the CR (this is guaranteed to work, because the operator pod serves the webhook)
I don't like the idea of fetching operand specific resources to check readiness (the CR's condition should be the authority for that)

akrejcir:
The `waitUntilDeployed()` function waits for the deployment, same as the `oc wait ...`. 
I moved this PR to draft and I'm just investigating the problem here, because I cannot reproduce it in my environment. The final version will not check specific operand resources, I just needed to see them in the logs, in case there was a specific error message in their conditions. But ginkgo panics after 10 minutes. 

For now, you don't need to review this PR.

akrejcir:
This is now ready for review. The CI passed two times in a row, so maybe it is fixed.

internal/common/environment.go
```diff
@@ -46,7 +46,7 @@ func GetNodeLabellerImages() NodeLabellerImages {
 	const kvmCpuNfdDefaultImage = "quay.io/kubevirt/cpu-nfd-plugin"
 
 	const libvirtDefaultVersion = "v0.21.0"
-	const libvirtDefaultImage = "kubevirt/virt-launcher"
+	const libvirtDefaultImage = "docker.io/kubevirt/virt-launcher"
```
vatsalparekh, line 49: I would like to stay away from any hard bindings to `docker.io` given the recent pull limits and stuff

akrejcir: Without the `docker.io` the image could not be pulled. Do you know any other way how I could fix it?

tests/tests_suite_test.go
```diff
@@ -125,18 +154,95 @@ func getSsp() *sspv1alpha1.SSP {
...
 
 func waitForDeletion(key client.ObjectKey, obj runtime.Object) {
-	Eventually(func() bool {
+	EventuallyWithOffset(1, func() bool {
 		err := apiClient.Get(ctx, key, obj)
 		return errors.IsNotFound(err)
 	}, timeout, time.Second).Should(BeTrue())
 }
 
+func logOperatorState() {
```
omeryahud, line 183: I think these debug functions are not necessary, because when a job fails, you can find all resources on the cluster under the `artifacts` link in the job page, this includes all resources (equivalent of `oc get $resource --all -A -oyaml`) and all container logs.
So unless there is a reason I am missing here, these functions can be removed

omeryahud: You can see all resources on the cluster here (from this PRs job page): https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/41/pull-ci-kubevirt-ssp-operator-master-e2e-functests/1328977540303294464/artifacts/e2e-functests/gather-extra/

Small enough logs will show up in the browser, the larger ones are downloaded to your machine and are actually `gzip` files, not textual `json` files

akrejcir: That is useful. I will remove some of the logging, but probably not all.
The `AfterSuite` part removes the SSP CR and the operator removes all the resources before tests end, so the resources are not in the `artifacts` directory.

omeryahud: If the `AfterSuite` removes the SSP CR this is not good.
The test suite should expect an existing deployment (deployed CR) and not remove it at end, so that the test suite can be run downstream as well.

omeryahud: It was an oversight on my end, I've forgotten about this requirement during reviews

akrejcir: So the `BeforeSuite` should not create a new SSP resource?
We should change this in a new PR.

omeryahud: Yes, the `BeforeSuite` should not create a new resources, and the `AfterSuite` should not delete it, a new PR is appropriate

omeryahud:
@akrejcir Can you please follow this up with a PR to resolve the issue we discussed yesterday?

internal/common/environment.go
vatsalparekh: I think there should be a plan to move this to `quay.io`

akrejcir: We can change it, when the image in quay will be available. Currently I'm getting an error when I try to pull it form quay:
```
$ docker pull quay.io/kubevirt/virt-launcher:v0.21.0
Error response from daemon: unauthorized: access to the requested resource is not authorized
```

I'm not sure if the image does not exist, or if I don't have premissions.

vatsalparekh (approved):
/lgtm

PR #42, opened 2020-11-12
Pause kubevirt-ssp-operator CRs on reconciliation

jean-edouard:
This PR adds an annotation to any existing kubevirt-ssp-operator on reconciliation.
It allows ssp-operator to take over when upgrading from kubevirt-ssp-operator.

Fixes #

**Release note**:

```release-note
NONE
```


jean-edouard:
No tests yet, working on it, hence the WIP state.
In the meantime, code reviews would be appreciated, since they could impact the tests.

internal/operands/common-templates/reconcile.go
```diff
@@ -87,6 +95,32 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
...
+	patch := []byte(`{"metadata":{"annotations":{"kubevirt.io/operator.paused": "true"}}}`)
+	var kubevirtCommonTemplatesBundles kvsspv1.KubevirtCommonTemplatesBundleList
+	err := request.Client.List(context.TODO(), &kubevirtCommonTemplatesBundles, &client.ListOptions{})
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error listing common template bundles: %s", err))
+		return
+	}
+	if err == nil && len(kubevirtCommonTemplatesBundles.Items) > 0 {
+		for _, kubevirtCommonTemplatesBundle := range kubevirtCommonTemplatesBundles.Items {
+			err = request.Client.Patch(context.TODO(), &kvsspv1.KubevirtCommonTemplatesBundle{
```
omeryahud, line 108: `request` has a `Context`, would be better to use it

omeryahud: Why have you inlined the CR here instead of providing `&kubevirtCommonTemplatesBundle` or `&kubevirtCommonTemplatesBundles.Items[i]` to the function?

internal/operands/common-templates/reconcile.go
```diff
@@ -87,6 +95,32 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: kubevirtCommonTemplatesBundle.ObjectMeta.Namespace,
+					Name:      kubevirtCommonTemplatesBundle.ObjectMeta.Name,
+				},
+			}, client.RawPatch(types.MergePatchType, patch))
+			if err != nil {
```
omeryahud, line 114: A `NotFound` error is an accepted case here and does not required any error logging, because if the old CR does not exist (during future upgrades), we just proceed with the deployment

internal/operands/common-templates/reconcile.go
```diff
@@ -87,6 +95,32 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
 	return nil
 }
 
+func pauseCRs(request *common.Request) {
```
omeryahud, line 98: I would say that this function needs to return an `error` if something went wrong with listing or patching

internal/operands/common-templates/reconcile.go
```diff
@@ -87,6 +95,32 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
...
+	patch := []byte(`{"metadata":{"annotations":{"kubevirt.io/operator.paused": "true"}}}`)
+	var kubevirtCommonTemplatesBundles kvsspv1.KubevirtCommonTemplatesBundleList
+	err := request.Client.List(context.TODO(), &kubevirtCommonTemplatesBundles, &client.ListOptions{})
+	if err != nil {
```
omeryahud, line 102: A `NotFound` error is acceptable here as well, no need for an error message

omeryahud (changes requested):
A few changes that affect all operands

internal/operands/common-templates/reconcile.go
akrejcir: This will also return an error if the old CRDs are not installed. I'm not sure if it is the same `NotFound` or different.

In a new cluster deployment, there will not be the old operator and old CRDs, so it would be a useless to call this function. I think in case the CRD does not exist, the `request.Client` cannot watch the resource so it cannot do caching. On every reconcile iteration this function would need to call k8s API.

Can you add code to not call this function if old CRDs are not installed? Maybe the check can be done when the operator starts.

internal/operands/common-templates/reconcile.go
```diff
@@ -29,6 +35,7 @@ var (
 // +kubebuilder:rbac:groups=core,resources=persistentvolumeclaims/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups=cdi.kubevirt.io,resources=datavolumes,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=cdi.kubevirt.io,resources=datavolumes/source,verbs=create
+// +kubebuilder:rbac:groups=ssp.kubevirt.io,resources=kubevirtcommontemplatesbundles,verbs=get;list;watch;create;update;patch;delete
```
akrejcir, line 38: Can you move this line to the block above? This block is for RBAC rules that are created by this operand.

internal/operands/common-templates/reconcile.go
```diff
@@ -87,6 +95,32 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
...
 
+func pauseCRs(request *common.Request) {
+	patch := []byte(`{"metadata":{"annotations":{"kubevirt.io/operator.paused": "true"}}}`)
+	var kubevirtCommonTemplatesBundles kvsspv1.KubevirtCommonTemplatesBundleList
+	err := request.Client.List(context.TODO(), &kubevirtCommonTemplatesBundles, &client.ListOptions{})
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error listing common template bundles: %s", err))
+		return
+	}
+	if err == nil && len(kubevirtCommonTemplatesBundles.Items) > 0 {
```
akrejcir, line 106: This check can be removed. At this point `err` is always `nil`, and it is not needed to check a slice's size before looping over it.

go.mod
```diff
@@ -5,14 +5,15 @@ go 1.13
...
+	github.com/kubevirt/kubevirt-ssp-operator v1.2.1
 	github.com/onsi/ginkgo v1.12.1
 	github.com/onsi/gomega v1.10.1
 	github.com/openshift/api v0.0.0-20200917102736-0a191b5b9bb0 // release-4.5
 	github.com/openshift/custom-resource-status v0.0.0-20200602122900-c002fd1547ca
 	github.com/operator-framework/operator-lib v0.1.0
 	k8s.io/api v0.18.6
 	k8s.io/apimachinery v0.18.6
-	k8s.io/client-go v8.0.0+incompatible
+	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
```
akrejcir, line 16: Can you run `go mod vendor` and commit the vendored files to the repo?

akrejcir (changes requested):

omeryahud:
I just noticed that the upgrade status fields are missing from this PR as well:

- Required env var: https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/deploy/operator.yaml#L54

- Required status fields: https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/pkg/apis/kubevirt/v1/types.go#L12-#L18

Once those are implemented, at the beginning of a reconciliation loop, `operatorVersion` and `targetVersion` should be set to the value of the `OPERATOR_VERSION` env var (indicating what version the operator is, and to which version of the operands it wants to progress), and at the end of the reconciliation loop (after ALL resources were properly reconciled), `observedVersion` should be set to the value of the `OPERATOR_VERSION` env var (indicating that all resources were successfully progressed to their target version).

jean-edouard:
Added a commit to address most comments.

Last remaining item before tests is dealing with `operatorVersion`/`targetVersion`/`observedVersion`.
I'll work on that one in a separate commit.

data/crd/ssp.kubevirt.io_ssps.yaml
```diff
@@ -0,0 +1,862 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
akrejcir, line 1: This file is not related to this PR. It was generated, because we forgot to add it to `.gitignore`.

data/olm-catalog/ssp-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,458 @@
+apiVersion: operators.coreos.com/v1alpha1
```
akrejcir, line 1: Similarly, this file is also not related. We probably need to add it to `.gitignore`.

internal/operands/common-templates/reconcile.go
```diff
@@ -54,6 +60,34 @@ func (c *commonTemplates) WatchTypes() []runtime.Object {
 	return nil
 }
 
+func (c *commonTemplates) PauseCRs(request *common.Request) error {
+	patch := []byte(`{"metadata":{"annotations":{"kubevirt.io/operator.paused": "true"}}}`)
+	var kubevirtCommonTemplatesBundles kvsspv1.KubevirtCommonTemplatesBundleList
+	err := request.Client.List(request.Context, &kubevirtCommonTemplatesBundles, &client.ListOptions{})
```
akrejcir, line 66: This function could use `UnstructuredList` instead of using specific type. It only lists old CRs and calls Patch on them. Then the old operator API code does not need to be imported as a module.

As a side effect, it would make this function more generic, and you could replace all 4 `PauseCRs` functions with one. The single function could be defined in `internal/commmon` package, or even in `controllers/ssp_controller.go`.

internal/operands/common-templates/reconcile.go
```diff
@@ -54,6 +60,34 @@ func (c *commonTemplates) WatchTypes() []runtime.Object {
...
+		if errors.IsNotFound(err) {
+			request.Logger.Info(fmt.Sprintf("No legacy common template bundle CR found"))
+			return nil
+		} else {
+			request.Logger.Error(err, fmt.Sprintf("Error listing common template bundle CRs: %s", err))
+			return err
+		}
+	}
+	for _, kubevirtCommonTemplatesBundle := range kubevirtCommonTemplatesBundles.Items {
+		err = request.Client.Patch(request.Context, &kubevirtCommonTemplatesBundle,
```
akrejcir, line 77: Can you add a check if the annotation is already present? And only call patch if it is not.

controllers/ssp_controller.go
```diff
@@ -187,8 +198,32 @@ func cleanup(request *common.Request) error {
...
+				}
+			}
+		}
+	}
+
 	allStatuses := make([]common.ResourceStatus, 0, len(sspOperands))
-	for _, operand := range sspOperands {
+	for i, operand := range sspOperands {
+		if crdsPresent[i] {
+			err := operand.PauseCRs(sspRequest)
```
akrejcir, line 222: I think you can decouple the `PauseCRs` functions from the operands. They don't need to be related. The pausing code can be moved to this file or somewhere else, but it does not need to be part of the `Operand` interface.

controllers/ssp_controller.go
```diff
@@ -55,6 +57,14 @@ var sspOperands = []operands.Operand{
 	// TODO - add other operands here
 }
 
+// 1:1 list of legacy CRDs corresponding to the operands above
```
akrejcir, line 60: If you extract the `PauseCRs` functions from the operands, this list will not need to be 1:1 mapping.

akrejcir (changes requested):
The logic looks good. I just have a few comments how to make the code cleaner maybe.

jean-edouard:
Fixed review comments in existing commits and added a commit to handle the *Version status fields

controllers/ssp_controller.go
```diff
@@ -186,7 +198,89 @@ func cleanup(request *common.Request) error {
...
+		for _, item := range crds.Items {
+			name := item.GetName()
+			for crd, kind := range kvsspCRDs {
+				if crd == name {
+					foundKinds = append(foundKinds, kind)
+					break
+				}
+			}
+		}
+	}
```
akrejcir, line 275: Can you move this block to a new function? Lines 259 - 275.

akrejcir (changes requested):

.gitignore
```diff
@@ -1,3 +1,4 @@
+*~
```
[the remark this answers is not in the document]
jean-edouard: Yeah... This used to be a pretty standard one, I still have the habit of adding it whenever I edit a gitignore, but I guess emacs isn't as popular as it used to be! Removed.

controllers/ssp_controller.go
```diff
@@ -186,7 +198,89 @@ func cleanup(request *common.Request) error {
...
 }
 
+func patchCRs(sspRequest *common.Request, kinds []string, patch []byte) error {
+	for _, kind := range kinds {
+		crs := &unstructured.UnstructuredList{}
+		crs.SetKind(kind)
+		crs.SetAPIVersion("ssp.kubevirt.io/v1")
+		err := sspRequest.Client.List(sspRequest.Context, crs)
+		if err != nil {
+			if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
jean-edouard: Will do, either later today or on Monday. Thank you!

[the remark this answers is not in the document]
jean-edouard: Done. If no CRs are found, `err` is nil and `len(crs.Items)` is 0.

controllers/ssp_controller.go
```diff
@@ -186,7 +198,95 @@ func cleanup(request *common.Request) error {
...
+
+	return nil
+}
+
+func preReconcile(sspRequest *common.Request, kinds []string, operatorVersion string) error {
+	preReconcilePatch := []byte(`{
+  "metadata":{
+    "annotations":{"kubevirt.io/operator.paused": "true"}
+  },
+  "status":{
```
omeryahud, line 234: The `*version` status fields should only be updated on the new `SSP` CR, the old ones are irrelevant and should only be paused and eventually removed by HCO.
`sspRequest.instance.Status.*Version` is where the version values should be

akrejcir: The version fields looks like a separate functionality, maybe it can be a new PR?

.gitignore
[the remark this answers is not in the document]
akrejcir: Interesting. I haven't used emacs, so I didn't know about it.

controllers/ssp_controller.go
omeryahud: My wording to describe this task probably wasn't the best, but I would like a single PR to include the entire upgrade path of the operator, and the `version` fields are part of it

.gitignore
```diff
@@ -7,3 +7,5 @@ build/_output/*
 bin/*
 bundle/*
 cover.out
+data/crd/ssp.kubevirt.io_ssps.yaml
+data/olm-catalog/ssp-operator.clusterserviceversion.yaml
```
akrejcir, line 11: I was mistaken, these files should not be ignored. But they don't need to be committed in this PR, they will be added in #44.

controllers/ssp_controller.go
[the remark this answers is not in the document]
akrejcir: Ok, if the error cannot be `IsNotFound`, then this `if` branch can be removed.

controllers/ssp_controller.go
```diff
@@ -186,7 +198,90 @@ func cleanup(request *common.Request) error {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func preReconcile(sspRequest *common.Request, kinds []string, operatorVersion string) error {
+	sspRequest.Instance.Status.OperatorVersion = operatorVersion
+	sspRequest.Instance.Status.TargetVersion = operatorVersion
```
akrejcir, line 231: These versions need to be updated in the cluster using the API:
```
sspRequest.Client.Status().Update(sspRequest.Context, request.Instance)
```

Please, extract these lines and the API call to a new function. Then the new function can be called from `Reconcile` method, before calling `reconcileOperands` function (on line 130).

controllers/ssp_controller.go
```diff
@@ -186,7 +198,90 @@ func cleanup(request *common.Request) error {
...
+	preReconcilePatch := []byte(`{
+  "metadata":{
+    "annotations":{"kubevirt.io/operator.paused": "true"}
+  }
+}`)
+
+	return patchCRs(sspRequest, kinds, preReconcilePatch)
+}
+
+func postReconcile(sspRequest *common.Request, kinds []string, operatorVersion string) error {
```
akrejcir, line 242: Please set the `ObservedVersion` in `updateStatus` function, where other status fields are updated. Then this function can be removed.

akrejcir (changes requested):

controllers/ssp_controller.go
```diff
@@ -186,7 +206,71 @@ func cleanup(request *common.Request) error {
...
+	}
+
+	return foundKinds
+}
+
 func reconcileOperands(sspRequest *common.Request) ([]common.ResourceStatus, error) {
+	opVer := os.Getenv("OPERATOR_VERSION")
+	if opVer != "" {
+		operatorVersion = opVer
+	}
```
akrejcir, line 263: The variable `operatorVersion` is accessed before this block is executed, in function `preUpdateStatus`.
You could extract this block into a separate function and use it in all needed places.
```
func getOperatorVersion() string {
	opVer := os.Getenv("OPERATOR_VERSION")
	if opVer != "" {
		return "devel"
	}
	return opVer
}
```

The function `os.Getenv()` caches env variables in a map, so it is not expensive to call it on each reconcile iteration.

go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	k8s.io/api v0.18.6
 	k8s.io/apiextensions-apiserver v0.18.6
 	k8s.io/apimachinery v0.18.6
-	k8s.io/client-go v8.0.0+incompatible
+	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
```
akrejcir, line 20: This version update is probably not needed. You can leave it as it was.

akrejcir (changes requested):

akrejcir (commented):
Code looks good. Only two minor comments. Please run `go mod vendor` and `go mod tidy` to remove all the changes in `vendor` directory, since this PR does not need them. And another comment is inline.

Now only tests are missing. Writing functional tests for this can be tricky. 
Maybe we could create a new test suite that tests only the upgrade path. It would assume that the old operator is already installed in the cluster, and all the old CRDs exist.

omeryahud:
> Code looks good. Only two minor comments. Please run `go mod vendor` and `go mod tidy` to remove all the changes in `vendor` directory, since this PR does not need them. And another comment is inline.
> 
> Now only tests are missing. Writing functional tests for this can be tricky.
> Maybe we could create a new test suite that tests only the upgrade path. It would assume that the old operator is already installed in the cluster, and all the old CRDs exist.

Yes, I think a new test lane is required IMO, my idea was to create a new lane in OCP CI `e2e-upgrade-functests`, that follows this flow:

- Deploys Kubevirt and CDI (a bash script that does that can be extracted from this deployment script that had served us well: https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/hack/deploy-operator.sh)
- Deploys the latest version of the old operator (similarly to the way it is done for Kubevirt and CDI in the above script)
- Deploys this operator from source and waits for it to be in the `Deployed` state
- Runs functional tests as usual

I will create the OCP CI job in the meantime

akrejcir:
It makes sense. 
Do we need to run functional tests as usual? Most of them do not depend on the upgrade and we would be running them twice. I would suggest we create a separate directory in `tests` with a test suite only for upgrade tests.

omeryahud:
> It makes sense.
> Do we need to run functional tests as usual? Most of them do not depend on the upgrade and we would be running them twice. I would suggest we create a separate directory in `tests` with a test suite only for upgrade tests.

I think running the exact same tests is exactly what we need to do, because we want to ensure that a fresh operator and an upgraded operator behave in the exact same way.

jean-edouard:
Added a commit for a functional test

akrejcir:
Looks good. After the upgrade test lane is merged, we can merge this.
https://github.com/openshift/release/pull/13867

akrejcir:
This needs to be rebased, for the upgrade tests CI to pass.

PR #43, opened 2020-11-18
Change default namespace for operator pod

vatsalparekh:

Change default namespace for operator pod
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -1,26 +1,24 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  labels:
-    control-plane: controller-manager
-  name: system
+  name: kubevirt
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
-  namespace: system
+  name: ssp-operator
+  namespace: kubevirt
   labels:
-    control-plane: controller-manager
+    control-plane: ssp-operator
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
+      control-plane: ssp-operator
   replicas: 1
   template:
     metadata:
       labels:
-        control-plane: controller-manager
+        control-plane: ssp-operator
     spec:
       containers:
       - command:
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,12 +1,12 @@
 # Adds namespace to all resources.
-namespace: ssp-operator-update-system
+namespace: kubevirt
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: ssp-operator-update-
+namePrefix: ""
 
 # Labels to add to all resources and selectors.
 #commonLabels:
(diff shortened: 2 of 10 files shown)


config/manager/manager.yaml
```diff
@@ -2,25 +2,25 @@ apiVersion: v1
...
-  name: system
+    control-plane: ssp-controller-manager
+  name: kubevirt
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
-  namespace: system
+  name: ssp-controller-manager
```
akrejcir, line 11: The deployment can have name `ssp-operator`. It would be clearer.
The labels can also be changed to `control-plane: ssp-operator`.
I'm not sure why the key is `control-plane` and if we can change it.

omeryahud: I agree, I think the `control-plane` key is there to integrate with OLM in some way, not sure though

config/manager/manager.yaml
```diff
@@ -2,25 +2,25 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
-  name: system
+    control-plane: ssp-controller-manager
```
omeryahud, line 5: this should be `ssp-operator` as well? (looking at the same label on the resources below)

vatsalparekh: corrected

config/manager/manager.yaml
```diff
@@ -2,25 +2,25 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: controller-manager
-  name: system
+    control-plane: ssp-operator
```
akrejcir, line 5: I think we should not add this label to the `kubevirt` namespace. At least until we know how is this label used. I will try to find out.

go.sum
```diff
@@ -198,6 +198,7 @@ github.com/evanphx/json-patch v4.0.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLi
 github.com/evanphx/json-patch v4.2.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
 github.com/evanphx/json-patch v4.5.0+incompatible h1:ouOWdg56aJriqS0huScTkVXPC5IcNrDCXZ6OoTAWu7M=
 github.com/evanphx/json-patch v4.5.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
+github.com/evanphx/json-patch v4.9.0+incompatible h1:kLcOMZeuLAJvL2BPWLMIj5oaZQobrkAqrL+WFZwQses=
```
akrejcir, line 201: This PR does not edit `go.mod` or anything in `vendor` directory, so changing this file is not needed.

akrejcir (changes requested):

config/manager/manager.yaml
omeryahud: Right, good point

vatsalparekh: Removed the whole label

go.sum
vatsalparekh: Corrected

akrejcir:
The label needs to be changed also in: `config/prometheus/monitor.yaml`.

vatsalparekh:
> The label needs to be changed also in: `config/prometheus/monitor.yaml`.

@akrejcir  We don't use that https://github.com/kubevirt/ssp-operator/blob/master/config/default/kustomization.yaml#L25

akrejcir:
Ok, I missed that.
/lgtm

PR #44, opened 2020-11-19
rewrite csv generator in golang

ksimon1:
rewrite csv generator in golang

**Release note**:
```NONE

```


.gitignore
```diff
@@ -6,4 +6,7 @@ _out/*
 build/_output/*
 bin/*
 bundle/*
+data/olm-catalog/*
```
omeryahud, line 9: Contents of the `data` directory need to be committed for ds use

hack/csv-generator.go
```diff
@@ -0,0 +1,263 @@
...
+	rootCmd.Flags().StringVar(&f.cpuPlugin, "cpu-plugin-image", "", "Link to cpu-nfd-plugin image")
+	rootCmd.Flags().BoolVar(&f.dumpCRDs, "dump-crds", false, "Dump crds to stdout")
+
+	rootCmd.MarkFlagRequired("csv-version")
+	rootCmd.MarkFlagRequired("namespace")
+	rootCmd.MarkFlagRequired("operator-image")
+}
+
+func runGenerator() error {
+	csvFile, err := ioutil.ReadFile("/data/olm-catalog/ssp-operator.clusterserviceversion.yaml")
```
omeryahud, line 110: The path to the `data` directory should be relative and not obsolete (that's how common-templates use it)

go.mod
```diff
@@ -3,19 +3,27 @@ module kubevirt.io/ssp-operator
...
 	k8s.io/apimachinery v0.18.6
 	k8s.io/client-go v8.0.0+incompatible
 	k8s.io/utils v0.0.0-20200821003339-5e75c0163111
 	kubevirt.io/controller-lifecycle-operator-sdk v0.1.0
 	sigs.k8s.io/controller-runtime v0.6.2
 )
 
-replace k8s.io/client-go => k8s.io/client-go v0.18.6
+replace (
+	github.com/operator-framework/operator-lifecycle-manager => github.com/operator-framework/operator-lifecycle-manager v0.0.0-20190128024246-5eb7ae5bdb7a
```
omeryahud, line 27: Can you `go get` the latest tagged version of this package instead?

go.mod
```diff
@@ -3,19 +3,27 @@ module kubevirt.io/ssp-operator
...
+	k8s.io/client-go => k8s.io/client-go v0.18.6
```
omeryahud, line 28: Could you bump the version of the package as well?

go.mod
akrejcir: It looks like the OLM API was moved to this repo: `https://github.com/operator-framework/api/tree/v0.3.21`

akrejcir (changes requested):

hack/csv-generator.go
```diff
@@ -0,0 +1,263 @@
...
+	ServiceAccountName string              `json:"serviceAccountName"`
+	Rules              []rbacv1.PolicyRule `json:"rules"`
+}
+
+type csvDeployments struct {
+	Name string                `json:"name"`
+	Spec appsv1.DeploymentSpec `json:"spec,omitempty"`
+}
+
+type csvStrategySpec struct {
```
akrejcir, line 67: The new API package has definitions for all these types. The `ClusterServiceVersion.Spec.InstallStrategy.StrategySpec` is no longer `json.RawMessage`.

hack/csv-generator.go
```diff
@@ -0,0 +1,69 @@
+/*
```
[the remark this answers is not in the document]
omeryahud: I would vote not to do that, because `operator-sdk` might expect `main.go` in the root directory if we try to generate webhooks (or anything else that adds code) in the future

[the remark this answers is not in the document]
akrejcir: Ok, good point.

hack/csv-generator.go
ksimon1: `manager` binary is located in `/` but csv-generator is located in `/usr/bin/` so I have to use absolute path

hack/csv-generator.go
```diff
@@ -0,0 +1,263 @@
...
+	strategySpecJSONBytes, err = json.Marshal(strategySpec)
+	if err != nil {
+		return err
+	}
+
+	csv.Spec.InstallStrategy.StrategySpecRaw = json.RawMessage(strategySpecJSONBytes)
+	return nil
+}
+
+func marshallObject(obj interface{}, writer io.Writer) error {
```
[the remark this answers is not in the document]
ksimon1: But then it wouldn't be possible to delete atributes (like timestamps) from csv

go.mod
ksimon1: I am not able to do that. For some unknown reasons, `k8s.io/client-go` always return to `v8.0.0+incompatible` version.

hack/csv-generator.go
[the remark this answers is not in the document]
akrejcir: True, I didn't notice that.

go.mod
akrejcir: The version returns to `v8.0.0+incompatible` because the other module `kubevirt.io/controller-lifecycle-operator-sdk v0.1.0` requires this version. That is why the `k8s.io/client-go` version is overridden back to `v0.18.6` in `replace` block.

hack/csv-generator.go
omeryahud: So let's please align both of them to be in `/`

hack/csv-generator.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+	}
+)
+
+func main() {
+	rootCmd.Execute()
+}
+
+func init() {
+	rootCmd.Flags().StringVar(&f.csvVersion, "csv-version", "", "Version of csv manifest (required)")
```
omeryahud, line 73: `operator-version` is missing here and should be added as an ENV var to the operator deployment manifest

ksimon1: isn't `operator-version` part of `operator-image`? Where should be `operator-version` used (e.g. `spec.version`?)?

omeryahud: No, `operator-version` should be passed to the `OPERATOR_VERSION` env var and is used by the operator to indicate upgrade status (see the old operator implementation)

data/crd/ssp.kubevirt.io_ssps.yaml
```diff
@@ -0,0 +1,862 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
akrejcir, line 1: Do we want to commit this file, or is it enough that it is generated by `make container-build`?
@omeryahud, You have mentioned that these files are needed for downstream, but do they need to be commited in the repo?

omeryahud: It should be committed because we don't have `operator-sdk` downstream to generate manifests, we can think of a way to resolve it in the future, but to quickly resolve this for now, they should be committed

akrejcir: Ok.

go.mod
```diff
@@ -3,14 +3,19 @@ module kubevirt.io/ssp-operator
 go 1.13
 
 require (
+	github.com/blang/semver v3.5.1+incompatible
```
akrejcir, line 6: Why not use a compatible version? For example the newest one `v4.0.0`, or if that is too new, maybe: `v3.8.0`?

hack/csv-generator.go
```diff
@@ -0,0 +1,238 @@
...
+var (
+	f       generatorFlags
+	rootCmd = &cobra.Command{
+		Use:   "csv-generator",
+		Short: "csv-generator for ssp operator",
+		Long:  `csv-generator generates deploy manifest for ssp operator`,
+		Run: func(cmd *cobra.Command, args []string) {
+			err := runGenerator()
+			if err != nil {
+				fmt.Println(err.Error())
```
akrejcir, line 62: It may be better to print the error to `stderr`, because `stdout` is used for output of generated resources. It would be strange if there was an error message in the middle of valid resources.

hack/csv-generator.go
```diff
@@ -0,0 +1,238 @@
...
+			}
+			updatedContainer.Env = updatedVariables
+			csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Containers[i] = updatedContainer
+			break
+		}
+	}
+	return nil
+}
+
+func marshallObject(obj interface{}, writer io.Writer) error {
```
akrejcir, line 178: The return value of this method is never checked.

hack/csv-generator.go
```diff
@@ -0,0 +1,238 @@
...
+		return err
+	}
+
+	// remove status and metadata.creationTimestamp
+	unstructured.RemoveNestedField(r.Object, "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(r.Object, "template", "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(r.Object, "spec", "template", "metadata", "creationTimestamp")
+	unstructured.RemoveNestedField(r.Object, "status")
+
+	deployments, exists, err := unstructured.NestedSlice(r.Object, "spec", "install", "spec", "deployments")
```
akrejcir, line 195: The `err` returned value from this line is never checked. If an error can never happen, please use `_` to ignore the returned value.

akrejcir (commented):
A few minor comments. If you think they are not important, then the patch looks good.

go.mod
ksimon1: Because according to https://pkg.go.dev/github.com/blang/semver?tab=versions
the latest v1 version is `v3.5.1+incompatible` and v4 is not compatible with `github.com/operator-framework/api`

akrejcir: Ok

akrejcir:
For some reason on my machine the CRDs are generated with newlines in `description` fields. Are you using `controller-gen` version `v0.4.0` ?

ksimon1:
> ine the CRDs are generated with newlines in

yes 
```
[ksimon@localhost ssp-operator]$ controller-gen --version
Version: v0.4.0

```

Makefile
```diff
@@ -94,18 +94,15 @@ operator-sdk:
...
-	mkdir -p data/olm-catalog
-	./hack/update-csv.py bundle/manifests/ssp-operator.clusterserviceversion.yaml > data/olm-catalog/ssp-operator.clusterserviceversion.yaml
-	cp bundle/manifests/ssp.kubevirt.io_ssps.yaml data/crd/ssp.kubevirt.io_ssps.yaml
-	rm -r bundle.Dockerfile
+build-csv-generator:
+	go build -v hack/csv-generator.go
 
 # Build the container image
-container-build: unittest olm-catalog
+container-build: unittest build-csv-generator bundle
```
omeryahud, line 101: No need to `build-csv-generator` here, as it is being built inside a container

omeryahud (commented):
One small thing and then lgtm

PR #46, opened 2020-11-20
Release container and manifests on tag

akrejcir:
When a commit is tagged TravisCI pushes the image to quay.io
and creates a github release.

```release-note
NONE
```


akrejcir:
@omeryahud, are we going to change the image repository to something like: `quay.io/kubevirt/ssp-operator` ?

omeryahud:
@akrejcir Yes, the repo already exists, you can update it in the makefile

PR #47, opened 2020-11-23
Remove the extra proxy sidecar

vatsalparekh:

Remove the extra proxy sidecar
Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -3,8 +3,8 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
-  namespace: system
+  name: ssp-operator
+  namespace: kubevirt
 spec:
   template:
     spec:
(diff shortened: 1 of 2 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -9,16 +9,6 @@ spec:
   template:
     spec:
       containers:
-      - name: kube-rbac-proxy
```
omeryahud, line 12: I think commenting out the entire patch is a better idea https://github.com/kubevirt/ssp-operator/blob/master/config/default/kustomization.yaml#L31

vatsalparekh: But we still need the manager container here

omeryahud: The manager container will remain as it is present in the deployment base resource https://github.com/kubevirt/ssp-operator/blob/master/config/manager/manager.yaml#L25

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -3,23 +3,23 @@
 apiVersion: apps/v1
```
omeryahud, line 3: Please fix indentation in this file and I think that's it

vatsalparekh: Corrected

PR #50, opened 2020-11-25
Remove workload, flavor, os labels from templates which have older version

ksimon1:
remove workload, flavor, os labels from templates which have older version

With labels removed, old templates will no longer be visible in UI.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1859235

**Release note**:
```
SSP operator now removes workload, flavor, os labels from templates which have older version
```

### DIFF
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -2,17 +2,21 @@ package common_templates
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
+	"sync"
 
 	templatev1 "github.com/openshift/api/template/v1"
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/operands"
-	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sync"
 )
 
 var (
@@ -61,6 +65,7 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ResourceS
 		reconcileViewRole,
 		reconcileViewRoleBinding,
 		reconcileEditRole,
+		updateOldTemplates,
 	}
 	funcs = append(funcs, reconcileTemplatesFuncs(request)...)
 	return common.CollectResourceStatus(request, funcs...)
@@ -122,6 +127,57 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
 			foundRole := foundRes.(*rbac.ClusterRole)
 			foundRole.Rules = newRole.Rules
 		})
+}
+func updateOldTemplates(request *common.Request) (common.ResourceStatus, error) {
+	baseRequirement, err := labels.NewRequirement("template.kubevirt.io/type", selection.Equals, []string{"base"})
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error during preparing of type label requirement, %v", err))
+		panic(err)
+	}
+	var versionRequirement *labels.Requirement
+	versionRequirement, err = labels.NewRequirement("template.kubevirt.io/version", selection.NotEquals, []string{Version})
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error during preparing of version label requirement, %v", err))
+		panic(err)
+	}
+
+	labelsSelector := labels.NewSelector().Add(*baseRequirement, *versionRequirement)
+
+	opts := client.ListOptions{
+		LabelSelector: labelsSelector,
+	}
+
+	var oldTemplates templatev1.TemplateList
+	err = request.Client.List(request.Context, &oldTemplates, &opts)
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error during listing templates, %v", err))
+		return common.ResourceStatus{}, err
+	}
+
+	updatedTemplates := make([]templatev1.Template, 0)
+	for _, template := range oldTemplates.Items {
+		updated := false
+		for key := range template.Labels {
+			if strings.HasPrefix(key, "os.template.kubevirt.io/") || strings.HasPrefix(key, "flavor.template.kubevirt.io/") || strings.HasPrefix(key, "workload.template.kubevirt.io/") {
+				updated = true
+				delete(template.Labels, key)
+			}
+		}
+		if updated {
+			updatedTemplates = append(updatedTemplates, template)
+		}
+	}
+	oldTemplates.Items = updatedTemplates
+
+	for _, template := range oldTemplates.Items {
+		err = request.Client.Update(request.Context, &template)
+		if err != nil {
+			request.Logger.Error(err, fmt.Sprintf("Error during updating of templates, %v", err))
+			return common.ResourceStatus{}, err
+		}
+	}
+
+	return common.ResourceStatus{}, nil
 }
 
 func reconcileTemplatesFuncs(request *common.Request) []common.ReconcileFunc {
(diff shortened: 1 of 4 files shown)


internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +130,53 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+
+	for _, template := range oldTemplates.Items {
+		err = request.Client.Update(request.Context, &template)
+		if err != nil {
+			request.Logger.Error(err, fmt.Sprintf("Error during updating of templates, %v", err))
+			panic(err)
+		}
+	}
+
+	return common.ResourceStatus{}, nil
```
ksimon1, line 178: @akrejcir, @vatsalparekh what should be returned in this case?

akrejcir: For other resources, this return value signals if the resource is still being updated, or if it is in the desired state. If any resource is still updating, the operator will not be in phase `deployed`.

In this case however,  the operator does not own any of the old templates, so I think they should not block the operator from being in `deployed` state. The return value can be left as is now.

internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +130,53 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error during listing templates, %v", err))
+		panic(err)
+	}
+
+	updatedTemplates := make([]templatev1.Template, 0)
+	for _, template := range oldTemplates.Items {
+		for key := range template.Labels {
+			if strings.HasPrefix(key, "os.template.kubevirt.io/") || strings.HasPrefix(key, "flavor.template.kubevirt.io/") || strings.HasPrefix(key, "workload.template.kubevirt.io/") {
+				delete(template.Labels, key)
```
akrejcir, line 163: I didn't know go allows removing elements from map while iterating over it. Good to know it does, in other languages it is usually not a good idea.

internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +130,53 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+		}
+		updatedTemplates = append(updatedTemplates, template)
+	}
+	oldTemplates.Items = updatedTemplates
+
+	for _, template := range oldTemplates.Items {
+		err = request.Client.Update(request.Context, &template)
+		if err != nil {
+			request.Logger.Error(err, fmt.Sprintf("Error during updating of templates, %v", err))
+			panic(err)
```
akrejcir, line 174: Same comment as above, it is better to return the error.

internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +130,53 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+			if strings.HasPrefix(key, "os.template.kubevirt.io/") || strings.HasPrefix(key, "flavor.template.kubevirt.io/") || strings.HasPrefix(key, "workload.template.kubevirt.io/") {
+				delete(template.Labels, key)
+			}
+		}
+		updatedTemplates = append(updatedTemplates, template)
+	}
+	oldTemplates.Items = updatedTemplates
+
+	for _, template := range oldTemplates.Items {
+		err = request.Client.Update(request.Context, &template)
```
akrejcir, line 171: This update is called for all retrieved templates, which is inefficient. Can you change it so that update is only called for modified templates?

akrejcir (changes requested):

internal/operands/common-templates/reconcile.go
ksimon1: This should be safe operation.
See https://golang.org/ref/spec#For_statements
```
The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next. If a map entry that has not yet been reached is removed during iteration, the corresponding iteration value will not be produced. If a map entry is created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next. If the map is nil, the number of iterations is 0. 
```

akrejcir: Yes, it is safe. I just didn't know that before reading this code. I always learn something new.

akrejcir:
Looks good. Do you think this needs a functional test?

ksimon1:
> Looks good. Do you think this needs a functional test?

Yes, but there is already one https://github.com/kubevirt/ssp-operator/blob/master/tests/e2e-test-csv-generator.sh

akrejcir:
That file only tests CSV generator, not old templates.

ksimon1:
ah you are right, sorry I posted wrong link

tests/tests_suite_test.go
```diff
@@ -88,6 +88,19 @@ func (s *newSspStrategy) Init() {
...
+	}, timeout, time.Second).ShouldNot(HaveOccurred())
+
+	Eventually(func() error {
+		return apiClient.Create(ctx, getTemplateWithOldVersion("1", strategy.GetTemplatesNamespace(), "v0.0.1"))
+	}, timeout, time.Second).ShouldNot(HaveOccurred())
+
+	Eventually(func() error {
+		return apiClient.Create(ctx, getTemplateWithOldVersion("2", strategy.GetTemplatesNamespace(), "v0.0.1"))
+	}, timeout, time.Second).ShouldNot(HaveOccurred())
+
```
akrejcir, line 103: It may be more clear if these lines that create templates were moved somewhere closer to the test case. Maybe to `BeforeEach()` or directly inside the test. But then the test would need to trigger reconciliation, for example by modifying a watched resource. I leave it up to you to decide.

In any case, these templates needs to be removed either after the test or on cleanup.

ksimon1: I would prefer to create in `BeforeSuite`, because they should represent templates, which were in cluster before ssp operator was deployed.
They will be deleted during namespace deletion https://github.com/kubevirt/ssp-operator/blob/master/tests/tests_suite_test.go#L109

akrejcir: Ok, then please add to code to create the templates also to the `existingSspStrategy`.

tests/commonTemplates_test.go
```diff
@@ -20,6 +24,8 @@ var _ = Describe("Common templates", func() {
 		editClusterRole testResource
 		goldenImageNS   testResource
 		testTemplate    testResource
+		currentVersion  = "v0.11.3"
```
akrejcir, line 27: Instead of using this variable, the tests should use `Version` from `internal/operand/common-templates/resources.go`.

tests/commonTemplates_test.go
```diff
@@ -122,4 +128,86 @@ var _ = Describe("Common templates", func() {
...
+				"flavor.template.kubevirt.io/test":     "true",
+				"template.kubevirt.io/type":            "base",
+				"workload.template.kubevirt.io/server": "true",
+				"template.kubevirt.io/version":         version,
+			},
+		},
+	}
+}
+
+func getTemplateWithOldVersion(indexStr, namespace, version string) *templatev1.Template {
```
akrejcir, line 195: These two functions `getTemplateWithCurrentVersion()` and `getTemplateWithOldVersion()` can be a single function. Their implementation is the same except for template name.

tests/tests_suite_test.go
```diff
@@ -88,6 +88,19 @@ func (s *newSspStrategy) Init() {
 	Eventually(func() error {
 		return apiClient.Create(ctx, newSsp)
 	}, timeout, time.Second).ShouldNot(HaveOccurred())
+
+	Eventually(func() error {
+		return apiClient.Create(ctx, getTemplateWithCurrentVersion(strategy.GetTemplatesNamespace(), "v0.11.3"))
```
akrejcir, line 93: Please use `Version` constant here too.

akrejcir (changes requested):
Looks good, just a few minor comments.

tests/commonTemplates_test.go
```diff
@@ -122,4 +127,66 @@ var _ = Describe("Common templates", func() {
 			table.Entry("[test_id:4770]golden image NS", &goldenImageNS),
 		)
 	})
+
+	FContext("update old templates", func() {
```
akrejcir, line 131: You forgot to unfocus the test. Th CI will fail if any tests are focused.

vatsalparekh:
Cloned this PR to #71 
/close

PR #52, opened 2020-11-26
Add scripts to deploy kubevirt, cdi  and the old operator

akrejcir:
Added scripts used by openshift CI

```release-note
NONE
```

### DIFF
--- a/hack/deploy-kubevirt-and-cdi.sh
+++ b/hack/deploy-kubevirt-and-cdi.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+set -e
+
+NAMESPACE=${1:-kubevirt}
+
+oc apply -f - <<EOF
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ${NAMESPACE}
+EOF
+
+# Deploying kuebvirt
+LATEST_KUBEVIRT=$(curl -L https://api.github.com/repos/kubevirt/kubevirt/releases | \
+            jq '.[] | select(.prerelease==false) | .name' | sort -V | tail -n1 | tr -d '"')
+
+oc apply -n $NAMESPACE -f "https://github.com/kubevirt/kubevirt/releases/download/${LATEST_KUBEVIRT}/kubevirt-operator.yaml"
+
+# Using KubeVirt CR from version v0.35.0
+oc apply -n $NAMESPACE -f - <<EOF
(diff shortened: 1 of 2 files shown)


akrejcir:
/cc @omeryahud, @vatsalparekh 
This PR is needed by: https://github.com/openshift/release/pull/13867

vatsalparekh (changes requested):
Can you also add a make target?

akrejcir:
These scripts are only helpers needed by the CI. I don't think they will be useful as makefile targets.

hack/deploy-kubevirt-and-cdi.sh
```diff
@@ -0,0 +1,52 @@
...
+EOF
+
+# Deploying kuebvirt
+LATEST_KUBEVIRT=$(curl -L https://api.github.com/repos/kubevirt/kubevirt/releases | \
+            jq '.[] | select(.prerelease==false) | .name' | sort -V | tail -n1 | tr -d '"')
+
+oc apply -n $NAMESPACE -f "https://github.com/kubevirt/kubevirt/releases/download/${LATEST_KUBEVIRT}/kubevirt-operator.yaml"
+oc apply -n $NAMESPACE -f "https://github.com/kubevirt/kubevirt/releases/download/${LATEST_KUBEVIRT}/kubevirt-cr.yaml"
+
+oc apply -n $NAMESPACE -f - <<EOF
```
jean-edouard, line 20: Nit: I believe the ConfigMap is the "legacy" way of configuring kubevirt. Feature gates are now configured in the CR, for example:
```
 spec:
  configuration:
    developerConfiguration:
      featureGates:
      - DataVolumes
      - CPUManager
      - LiveMigration
```

jean-edouard (approved):

hack/deploy-kubevirt-and-cdi.sh
vatsalparekh: @akrejcir can you confirm this?
otherwise, /lgtm

ksimon1: It is like Jed saying https://pkg.go.dev/kubevirt.io/client-go/api/v1#DeveloperConfiguration

akrejcir: Changed to use `featureGates`.

jean-edouard (approved):

PR #37, opened 2020-11-11
[WIP] Provided defaults for non-required spec fields

omeryahud:
Signed-off-by: Omer Yahud <oyahud@redhat.com>

**What this PR does / why we need it**:

Some of the fields in our spec are not strictly required, so I made them optional and provided defaults for them so the CR can be used as follows:

```
kind: SSP
metadata:
  name: ssp-sample
  namespace: kubevirt
spec:
  commonTemplates:
    namespace: kubevirt
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #56, opened 2020-12-04
Made some CR fields optional and set defaults for the optional fields

akrejcir:
Most fields are optional and the default value for template validator replicas is 2.

The minimal PR looks like:
```
apiVersion: ssp.kubevirt.io/v1beta1
kind: SSP
metadata:
  name: example-ssp
  namespace: kubevirt
spec:
  commonTemplates:
    namespace: kubevirt

```

**Release note**:
```release-note
NONE
```

This PR takes over code from: #37


internal/operands/node-labeller/defaults.go
```diff
@@ -0,0 +1,8 @@
+package node_labeller
+
+const (
+	kubevirtNodeLabellerDefaultImage = "quay.io/kubevirt/node-labeller:v0.2.0"
+	kvmInfoNfdDefaultImage           = "quay.io/kubevirt/kvm-info-nfd-plugin:v0.5.8"
+	kvmCpuNfdDefaultImage            = "quay.io/kubevirt/cpu-nfd-plugin:v0.1.1"
+	libvirtDefaultImage              = "docker.io/kubevirt/virt-launcher:v0.21.0"
```
ksimon1, line 7: I think we can update this image to version `v0.35.0`

akrejcir: Ok, I will open another PR for it.

PR #59, opened 2020-12-07
Break infinite reconciliation loop

akrejcir:
Since Openshift `4.6` and Kubernetes `v1.19.0`, a status change on SSP CR causes reconciliation.
This patch adds a predicate to ignore the status change, otherwise the operator would never stop with reconciliation.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -34,9 +34,12 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	lifecycleapi "kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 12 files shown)


akrejcir:
/hold

There are some race conditions that I need to fix.

kwiesmueller:
I'm curious on what caused the loop, like which field. More specifically if it was caused by .metadata.managedFields?

akrejcir:
> I'm curious on what caused the loop, like which field. More specifically if it was caused by .metadata.managedFields?

Possibly. I compared the difference between SSP CRs, when the operator received the update event.
They differed in:
- `resourceVersion`
- timestamp in `.metadata.managedFields`
- timestamp in `status.conditions[*].LastHeartbeatTime`

I suspect the update event was caused by updating the status, but maybe also `managedFields`.

kwiesmueller:
The managedFields timestamp changes on an Update to the resource. There was a bug that should be fixed where a no-op update can still update the timestamp in some cases.
A status change sounds like a good explanation for the event.
It could be tested if the loop also occurs with ServerSideApply disabled, but I don't know if there is time for more investigating.
Just my curiosity getting a hold on me.

akrejcir:
Interesting. The exact k8s version I found this on is: `v1.19.0+9f84db3`, maybe it was not yet fixed there.

PR #62, opened 2020-12-09
Add duplicate template test

kwiesmueller:

Adds a functional test to ensure there is only one template for every (os, workload, flavor, version) combination.
Derived from https://github.com/kubevirt/common-templates/blob/master/automation/test_duplicate_templates.sh

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -114,11 +114,19 @@ var _ = Describe("Common templates", func() {
 		})
 
 		It("[test_id:5545]did not create duplicate templates", func() {
-			liveTemplates := &templatev1.TemplateList{}
 			// TODO: the template path is relative pointing to the /data directory right now.
 			expectedTemplates, err := commonTemplates.ReadTemplates(filepath.Join("../"+commonTemplates.BundleDir, "common-templates-"+commonTemplates.Version+".yaml"))
 			Expect(err).NotTo(HaveOccurred())
 			Expect(expectedTemplates).NotTo(HaveLen(0))
+
+			liveTemplates := &templatev1.TemplateList{}
+			err = apiClient.List(ctx, liveTemplates,
+				client.InNamespace(strategy.GetTemplatesNamespace()),
+				client.MatchingLabels{
+					"template.kubevirt.io/version": commonTemplates.Version,
+				},
+			)
+			Expect(err).ToNot(HaveOccurred())
 
 			for _, template := range expectedTemplates {
 				Expect(template.ObjectMeta).NotTo(BeNil())
@@ -148,17 +156,16 @@ var _ = Describe("Common templates", func() {
 				for _, os := range oss {
 					for _, workload := range workloads {
 						for _, flavor := range flavors {
-							err = apiClient.List(ctx, liveTemplates,
-								client.InNamespace(strategy.GetTemplatesNamespace()),
-								client.MatchingLabels{
-									"template.kubevirt.io/version": commonTemplates.Version,
-								},
-								client.HasLabels{
-									os, workload, flavor,
-								},
-							)
-							Expect(err).ToNot(HaveOccurred())
-							Expect(liveTemplates.Items).To(HaveLen(1))
+							matchingLiveTemplates := 0
+							for _, liveTemplate := range liveTemplates.Items {
+								_, osMatch := liveTemplate.Labels[os]
+								_, workloadMatch := liveTemplate.Labels[workload]
+								_, flavorMatch := liveTemplate.Labels[flavor]
+								if osMatch && workloadMatch && flavorMatch {
+									matchingLiveTemplates++
+								}
+							}
+							Expect(matchingLiveTemplates).To(BeNumerically("==", 1))
 						}
 					}
 				}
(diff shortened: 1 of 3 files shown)


tests/commonTemplates_test.go
```diff
@@ -75,6 +79,57 @@ var _ = Describe("Common templates", func() {
...
+								client.InNamespace(allTemplates.Namsespace),
+								client.MatchingLabels{
+									"template.kubevirt.io/version": commonTemplates.Version,
+								},
+								client.HasLabels{
+									os, workload, flavor,
+								},
+							)
+							Expect(err).ToNot(HaveOccurred())
+							Expect(liveTemplates).To(HaveLen(1))
```
vatsalparekh, line 127: shouldn't this be
```suggestion
							Expect(liveTemplates.Items).To(HaveLen(1))
```

vatsalparekh (changes requested):
Just a small nit, looks good otherwise

vatsalparekh:
@kwiesmueller there seems to be a path error 
```
open data/common-templates-bundle/common-templates-v0.12.2.yaml: no such file or directory```

File: tests/commonTemplates_test.go (hunk 1 above)
Comment (kwiesmueller) [reply tests/commonTemplates_test.go]: Thanks, good catch!
Fixed.

**Comment by akrejcir** [comment]
/retest

Sometimes the CI fails with issue not related to the PR.

**Comment by akrejcir** [comment]
The PR looks ok. It seems correct and behaves like the original test.

I don't like the 4 nested loops with an API call from the performance side. It looks slow but since this is a test, it is not a priority.
If you want, maybe you can optimize it, or at least add a "TODO" comment so we don't forget about it.

**Comment by kwiesmueller** [comment]
My thought exactly. I wanted to copy it as close to the original first and get the test green and then change this.
I really don't like to make so many list calls to the apiserver. I'd probably pull the entire list for version and then check the labels on it (or sth. like that).
Will see, first I need to get the test working. Right now OpenShift seemed to hate me, finally got a working cluster.
And the data/ folder is not available to the test. Have to duplicate it another time, trying to see if there's a better option than having the yaml in there three times.

**Comment by kwiesmueller** [comment]
@akrejcir @vatsalparekh  PTAL
Also moved the API call outside. We might still be able to do better, but this way the test is fairly fast.

File: tests/commonTemplates_test.go [hunk 2]
```diff
@@ -75,6 +78,65 @@ var _ = Describe("Common templates", func() {
 			err := apiClient.Get(ctx, testTemplate.GetKey(), testTemplate.NewResource())
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		It("[test_id:TODO]did not create duplicate templates", func() {
```
kwiesmueller, line 82: /hold 
Still needs test_id

kwiesmueller: Done.
/hold remove

kwiesmueller: Done.
/hold cancel

tests/commonTemplates_test.go
```diff
@@ -75,6 +78,65 @@ var _ = Describe("Common templates", func() {
...
+			liveTemplates := &templatev1.TemplateList{}
+			err = apiClient.List(ctx, liveTemplates,
+				client.InNamespace(strategy.GetTemplatesNamespace()),
+				client.MatchingLabels{
+					"template.kubevirt.io/version": commonTemplates.Version,
+				},
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			for _, template := range expectedTemplates {
```
omeryahud, line 97: I would prefer the test not read the templates bundle file, and work only with the deployed templates,.
This keeps this test consistent for an upgrade scenario as well, when templates from a previous bundle (that do not exist in the current bundle) are deployed (consistency should be kept in that scenario as well)

omeryahud (changes requested):

tests/commonTemplates_test.go
kwiesmueller: I thought so too, bit wanted to replicate the test first.
It seemed to me, that the bundle was used to make sure every template in the bundle is present as well which is why I kept it.
I'm happy to change it though.

vatsalparekh: @kwiesmueller I would also go with Omer's comment about maintaining consistency and listing the templates from already deployed ones.
But also have an separate test for checking each template in the bundle is present

vatsalparekh (changes requested):

tests/commonTemplates_test.go
kwiesmueller: https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/common-templates/reconcile_test.go#L79
Looks like we already have a test checking all templates are there.
So this should be done now.

tests/commonTemplates_test.go
```diff
@@ -111,6 +111,60 @@ var _ = Describe("Common templates", func() {
...
+						flavors = append(flavors, labelKey)
+						continue
+					}
+				}
+
+				for _, os := range oss {
+					for _, workload := range workloads {
+						for _, flavor := range flavors {
+							matchingLiveTemplates := 0
+							for _, liveTemplate := range liveTemplates.Items {
```
vatsalparekh, line 154: 4 nested loops should be avoided but as there are no other calls made in the loops, I won't mind it
/lgtm
/approve

PR #63, opened 2020-12-10
Add test ensuring one default template per OS

kwiesmueller:
Signed-off-by: Kevin Wiesmueller <kwiesmul@redhat.com>

Adds a functional test to ensure there is only one template marked as default variation per os.
Derived from https://github.com/kubevirt/common-templates/blob/master/automation/test_defaults.sh

This test will currently fail until the common-template bundle in this repository is updated to have the `template.kubevirt.io/default-os-variant` present. This will be done in a separate PR so this can be rebased.
/hold

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -1,7 +1,9 @@
 package tests
 
 import (
+	"fmt"
 	"reflect"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
@@ -9,6 +11,7 @@ import (
 	templatev1 "github.com/openshift/api/template/v1"
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	commonTemplates "kubevirt.io/ssp-operator/internal/operands/common-templates"
 )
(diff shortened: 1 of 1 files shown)


kwiesmueller:
Seems like centos8 does not have a default specified in the templates bundle

tests/commonTemplates_test.go
```diff
@@ -75,6 +78,39 @@ var _ = Describe("Common templates", func() {
...
+			for _, liveTemplate := range liveTemplates.Items {
+				_, isDefaultOSVariant := liveTemplate.Labels["template.kubevirt.io/default-os-variant"]
+
+				for labelKey := range liveTemplate.Labels {
+					if strings.HasPrefix(labelKey, "os.template.kubevirt.io/") {
+						if isDefaultOSVariant {
+							osDefaultCounts[labelKey]++
+							continue
+						}
+						if _, knownOSVariant := osDefaultCounts[labelKey]; !knownOSVariant {
```
omeryahud, line 102: Map values are already zero-valued, so this block can be dropped (unless I am missing something)

tests/commonTemplates_test.go
```diff
@@ -75,6 +78,39 @@ var _ = Describe("Common templates", func() {
 			err := apiClient.Get(ctx, testTemplate.GetKey(), testTemplate.NewResource())
 			Expect(err).ToNot(HaveOccurred())
 		})
+
+		It("[test_id:5352]creates only one default variant per OS", func() {
```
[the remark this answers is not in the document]
kwiesmueller: I thought this is how tests are named in ginkgo.
So basically forming assertions through the tree.

In this case "resource creation only creates one default variant..."

rnetser, line 82: test_id: 5545 (test id 5352 is verify that only one template per os-flavor-workload has "template.kubevirt.io/default-os-variant=true")

kwiesmueller: Then this should be right, no?
5545 should be https://github.com/kubevirt/ssp-operator/pull/62

tests/commonTemplates_test.go
kwiesmueller: In theory yes, here specifically this line is required though.
In the loop bellow where the counts get checked, it iterates over the map. If this line is missing it would never add OSs that do not have a default so the test would not fail if there is a an OS without a default variant.
I can clarify this in a comment as it's easy to overlook.

tests/commonTemplates_test.go
rnetser: sorry; was looking at the worng pr

PR #67, opened 2020-12-17
Fix CSV for OLM deployments

omeryahud:

This commit addresses two OLM issues we are experiencing:
- [Temporary solution] Our CSV was missing the csv.webhookdefinitions[0].containerPort field, which prevented OLM from creating the appropriate service for our webhook server.
      For a reason I still don't know, operator-sdk does not populate the 'containerPort' field, so I had to add it manually using 'yq'
- On OLM deployments, the certificates secret is mounted in a different directory than what operator-sdk expects by default.
      To solve this, I have copied the certificates from OLM's volume directory to operator-sdk's certificate directory

**Release note**:

```release-note
None
```


main.go
```diff
@@ -56,6 +62,62 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
+	// If certificates are generated by OLM, we should use OLM certificates mount path
```
zcahana, line 65: The copy from `olmCertDir` to `certDir` is redundant, as OLM already mounts the certificate secret to both.
Here's an excerpt from the `ssp-operator` pod running in HCO CI (with a manual CSV patch to remove the `certs` volume):
```json
"volumeMounts": [
	{
		"mountPath": "/apiserver.local.config/certificates",
		"name": "apiservice-cert"
	},
	{
		"mountPath": "/tmp/k8s-webhook-server/serving-certs",
		"name": "webhook-cert"
	},
	{
		"mountPath": "/var/run/secrets/kubernetes.io/serviceaccount",
		"name": "ssp-operator-token-455fh",
		"readOnly": true
	}
]
```

And:
```json
"volumes": [
	{
		"name": "apiservice-cert",
		"secret": {
			"defaultMode": 420,
			"items": [
				{
					"key": "tls.crt",
					"path": "apiserver.crt"
				},
				{
					"key": "tls.key",
					"path": "apiserver.key"
				}
			],
			"secretName": "ssp-operator-service-cert"
		}
	},
	{
		"name": "webhook-cert",
		"secret": {
			"defaultMode": 420,
			"items": [
				{
					"key": "tls.crt",
					"path": "tls.crt"
				},
				{
					"key": "tls.key",
					"path": "tls.key"
				}
			],
			"secretName": "ssp-operator-service-cert"
		}
	},
	{
		"name": "ssp-operator-token-455fh",
		"secret": {
			"defaultMode": 420,
			"secretName": "ssp-operator-token-455fh"
		}
	}
]
```

omeryahud: I seem to get different behavior with OLM on OCP 4.6.6, when I patch the CSV to remove the `volumes` and `volumeMounts` from the operator pod:
```
      volumeMounts:
      - mountPath: /apiserver.local.config/certificates
        name: apiservice-cert
      - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
        name: ssp-operator-token-xqhsb
        readOnly: true
...
    volumes:
    - name: apiservice-cert
      secret:
        defaultMode: 420
        items:
        - key: tls.crt
          path: apiserver.crt
        - key: tls.key
          path: apiserver.key
        secretName: ssp-operator-service-cert
```

zcahana: Interesting. @orenc1 do you have any idea why that differs? And what OCP/OLM version is running in our CI?

omeryahud: I remember seeing a different behavior on a CRC 4.5.x cluster, it might have changed after I've upgraded to 4.6.6, I'll take another look

zcahana: It appears that this was added in OCP 4.7 (see [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/release-4.7/pkg/controller/install/certresources.go#L516), but was missing in 4.6.

zcahana: The CNV release we're targeting is aligned with OCP 4.7 right?

omeryahud: Yes, you are correct, 4.6 does not mount the certs to the webhook cert path.
Another issue is that operator-sdk expects `tls.crt` and `tls.key`, but OLM in 4.6 mounts them to a read-only volume as `apiserver.crt` and `apiserver.key`, so they cannot be renamed/copied in the same directory, so I have to copy them to operator-sdk's expected directory.
I would like to keep this copy operation to allow us to develop on latest CRC (4.6.6 as of now)

zcahana: well, copying the files to `/tmp/k8s-webhook-server/serving-certs`, even when they already exists there (OCP 4.7+), shouldn't cause any problems - it'll just overwrite them with the same files. If you want to be more righteous, you can test if the files exist before copying, but the end result should be the same.

omeryahud: Yes, checking if the files exist makes sense, I will add a condition for that

akrejcir:
Copying certs is only ok as a temporary solution. Openshift does certificate rotation by changing the contents of the mounted secret. We will need to fix this later.

PR #70, opened 2020-12-21
Migrate template validation tests

vatsalparekh:
Migrate template-validator tests from template-validator [repository](https://github.com/kubevirt/kubevirt-template-validator/tree/master/functests) to ssp-operator golang based tests

**Release note**:

```release-note
NONE
```


tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		It("[test_id:5033]: Template with validations, VM without validations", func() {
+			vmi.Spec.Domain.CPU = &v1.CPU{
+				Cores: 2,
+			}
+			vmi.Spec.Domain.Machine.Type = "q35"
+			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{
+				k8sv1.ResourceMemory: resource.MustParse("128M"),
+			}
```
akrejcir, line 381: These values are always assigned to create a VM that needs to pass validation. It would be clearer to extract them to a function with descriptive name like `CreateVmiToPassValidation`, and use is in every test  that needs it.

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+				Sockets: 1,
+			}
+			vm = test_utils.NewRandomVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           templateWithRules.Name,
+				"vm.kubevirt.io/template-namespace": templateWithRules.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		It("[test_id:5173]: should create a VM that passes validation", func() {
```
akrejcir, line 530: Please create a new `Context` to group all the tests for validation rules inside the VM's annotations.

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+														"message": "Core amount not within range",
+														"rule": "integer",
+														"min": 1,
+														"max": 4
+        										 }
+												]`,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Failed to create VM")
+		})
+		It("[test_id:5036]: should successfully create a VM based on the VM validation rules", func() {
```
akrejcir, line 605: Can you change the description to mention that VM's rules should have priority over template's rules? It took me a wile to notice the difference from test `[test_id:5173]`

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+			vmi.Spec.Domain.Machine.Type = "q35"
+			vmi.Spec.Domain.Resources.Requests = k8sv1.ResourceList{
+				k8sv1.ResourceMemory: resource.MustParse("64M"),
+			}
+			vmi.Spec.Domain.CPU = &v1.CPU{
+				Cores: 3,
+			}
+			vm = test_utils.NewRandomVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           "fedora-desktop-small-with-rules",
```
akrejcir, line 640: If the template does not exist, use a name that obviously does not exist, like `nonexisting-vm-template`

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+														"message": "Core amount not within range",
+														"rule": "integer",
+														"min": 1,
+														"max": 4
+        										 }
+												]`,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Failed to create VM")
+		})
+		It("[test_id:5046]: should fail to create a VM based on the VM validation rules", func() {
```
akrejcir, line 680: This test has the same description and ID as the one above.

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+														"message": "Core amount not within range",
+														"rule": "integer",
+														"min": 1,
+														"max": 4
+        										 }
+												]`,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Failed to create VM")
+		})
+		It("[test_id:5174]: VM with validations and deleted template", func() {
```
akrejcir, line 703: This is the same test as the one above.

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+														"message": "Core amount not within range",
+														"rule": "integer",
+														"min": 1,
+														"max": 4
+        										 }
+												]`,
+			}
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		AfterEach(func() {
```
akrejcir, line 726: Please move `AfterEach` block to the top of `Context`, right after `BeforeEach`. It will be easier to find.

akrejcir (changes requested):
Can you add more details to the commit message?

internal/test-utils/test_utils.go
```diff
@@ -23,3 +33,280 @@ func ExpectResourceNotExists(resource controllerutil.Object, request common.Requ
...
+		k8sv1.ResourceMemory: resource.MustParse("64M"),
+	}
+	vmi.Spec.Networks = []v1.Network{*v1.DefaultPodNetwork()}
+	return vmi
+}
+
+func NewRandomVirtualMachine(vmi *v1.VirtualMachineInstance, running bool) *v1.VirtualMachine {
+	name := vmi.Name
+	namespace := vmi.Namespace
+	labels := map[string]string{"kubevirt.io/vm": name}
```
ksimon1, line 69: I think more efficient would be:
```
labels := vmi.Labels
labels["kubevirt.io/vm"] = name
```

internal/test-utils/test_utils.go
```diff
@@ -23,3 +33,280 @@ func ExpectResourceNotExists(resource controllerutil.Object, request common.Requ
...
+	running     = false
+	liveMigrate = v1.EvictionStrategyLiveMigrate
+)
+
+func addObjectstoTemplate(name, validation string) templatev1.Template {
+	template := templatev1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Annotations: map[string]string{
+				"openshift.io/display-name":             "Fedora 23+ VM",
```
ksimon1, line 114: Do we need all of these annotations and labels?
AFAIK only UI uses them

internal/test-utils/test_utils.go
```diff
@@ -23,3 +33,280 @@ func ExpectResourceNotExists(resource controllerutil.Object, request common.Requ
...
+			Raw: []byte(runtime.EncodeOrDie(corev1Codec, &v1.VirtualMachine{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "${NAME}",
+					Namespace: "ssp-operator-functests",
+					Labels: map[string]string{
+						"vm.kubevirt.io/template": "fedora-desktop-small",
+						"app":                     "${NAME}",
+					},
+				},
+				Spec: v1.VirtualMachineSpec{
```
ksimon1, line 166: can't you reuse VM which is returned by `NewRandomVirtualMachine`?

tests/validator_test.go
```diff
@@ -320,13 +324,415 @@ var _ = Describe("Template validator", func() {
...
+		})
+
+		It("be created without template", func() {
+			vm = test_utils.NewRandomVirtualMachine(vmi, false)
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		It("be created from template without rules", func() {
+			vm = test_utils.NewRandomVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           "fedora-desktop-small",
```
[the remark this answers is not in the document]
akrejcir: I think that is an unnecessary dependency between template validator functional tests and common templates. Do you think we can use a template that was created by this test suite instead of a common template?

tests/tests_common_test.go
```diff
@@ -235,3 +236,59 @@ func expectUserCannot(sars *authv1.SubjectAccessReviewSpec) {
 			sars.ResourceAttributes.Subresource, sars.ResourceAttributes.Name, sars.ResourceAttributes.Group,
 			sars.ResourceAttributes.Version, sars.ResourceAttributes.Namespace))
 }
+
+func GetVMIPodLogs(vmi *v1.VirtualMachineInstance) (string, error) {
```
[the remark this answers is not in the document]
akrejcir: Ok, in this case it is not important to move it.

But for future PRs, there is a guideline in software engineering, that says it is better to write specific code and generalize only when needed. It is not necessary to create generic functions that may be used in the future, because we don't know for certain if they will. When they will be needed, we can easily create them. (I don't remember the name of this guideline)

In this case it is not that important, but please keep it in mind, when you think about future tests that may use it.
If you don't have a specific test in mind that would use it, then it is probably not needed now.

tests/tests_common_test.go
```diff
@@ -235,3 +236,59 @@ func expectUserCannot(sars *authv1.SubjectAccessReviewSpec) {
...
+	}
+	if len(pods.Items) == 0 {
+		return nil, fmt.Errorf("failed to find pod with the label %s", label)
+	}
+
+	var readyPod *k8sv1.Pod
+	for _, pod := range pods.Items {
+		ready := true
+		for _, status := range pod.Status.ContainerStatuses {
+			if status.Name == "kubevirt-infra" {
```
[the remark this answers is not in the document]
akrejcir: I have tried searching for the string "kubevirt-infra" in the kubevirt repository, but there is no container with this name. Maybe this code is old and the check is not needed anymore.

Please add a comment that explains that this check is also done in kubevirt, and a TODO note, that we should figure out if it is still needed.

tests/validator_test.go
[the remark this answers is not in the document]
vatsalparekh: Created a minimal template for it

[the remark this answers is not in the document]
vatsalparekh: Created a minimal template for it

tests/validator_test.go
```diff
@@ -329,4 +337,594 @@ var _ = Describe("Template validator", func() {
...
+												]`,
+				}
+				Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+			})
+		})
+	})
+
+	Context("Certificates", func() {
+		var (
+			secretName = "virt-template-validator-certs"
```
akrejcir, line 686: You can use `validator.secretName` directly, when you export it.

tests/validator_test.go
```diff
@@ -329,4 +337,594 @@ var _ = Describe("Template validator", func() {
...
+		)
+		It("[test_id:4375] Test refreshing of certificates", func() {
```
akrejcir, line 688: This test is not the same as the one in `kubevirt-template-validator` repository. This one does not actually test certificate rotation inside the template validator.

The original test may be more complicated to port into golang. If you don't want to do it in this PR, then please mark this test as pending, write a TODO comment, and open a new issue, so we don't forget about it.

akrejcir (changes requested):
Can you update the PR's description and the commit message to include the link to the `functests` directory in the template validator repository? https://github.com/kubevirt/kubevirt-template-validator/tree/v0.9.0/functests

Please remove unnecessary text from the PR's description.

akrejcir:
Looks good. Only thing that is left is to update the PR's description.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			template.Namespace = strategy.GetNamespace()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "128M")
+			vm = NewVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           template.Name,
+				"vm.kubevirt.io/template-namespace": template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Should have given error failing to fulfill validations")
```
akrejcir, line 435: A typo: `ToNot(` -> `To(`. The rules are not fulfilled, so the validation should fail. Strange that the CI passed for this one. Maybe there is a bug in validator.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			template.Namespace = strategy.GetNamespace()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "32M")
+			vm = NewVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           template.Name,
+				"vm.kubevirt.io/template-namespace": template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Should have given error failing to fulfill validations")
```
akrejcir, line 461: The same comment as above, you need to check the error.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			template.Namespace = strategy.GetNamespace()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "128M")
+			vm = NewVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           template.Name,
+				"vm.kubevirt.io/template-namespace": template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Should have given error failing to fulfill validations")
```
akrejcir, line 448: The same comment as above, you need to check the error.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "32M")
+			vm = NewVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           template.Name,
+				"vm.kubevirt.io/template-namespace": template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Should have given error failing to fulfill validations")
+		})
+		It("[test_id:2959] Start a VM with memory restrictions violation that succeeds with a warning", func() {
```
akrejcir, line 463: Inaccurate description, VM is not started.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			template.Namespace = strategy.GetNamespace()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "1G")
+			vm = NewVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           template.Name,
+				"vm.kubevirt.io/template-namespace": template.Namespace,
+			}
+			// TODO: Investigate why this test consistently fails without Eventually
```
akrejcir, line 474: Probably the same reason as the other test - race condition.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+														"name": "LimitCores",
+														"path": "jsonpath::.spec.domain.cpu.cores",
+														"message": "Core amount not within range",
+														"rule": "integer",
+														"min": 1,
+														"max": 4
+        										 }
+												]`,
+				}
+				Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Failed to create VM")
```
akrejcir, line 586: Check the error type and fix the text.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
```
akrejcir, line 609: Check the error type and fix the text.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
```
akrejcir, line 670: Check the error type and fix the text.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
```
akrejcir, line 687: Check the error type and fix the text.

akrejcir (changes requested):

internal/operands/template-validator/resources.go
```diff
@@ -19,20 +19,20 @@ import (
 
 const (
 	containerPort          = 8443
-	kubevirtIo             = "kubevirt.io"
-	secretName             = "virt-template-validator-certs"
-	virtTemplateValidator  = "virt-template-validator"
+	KubevirtIo             = "kubevirt.io"
```
ksimon1, line 22: will these variables be used in another package?

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+						},
+					},
+				},
+			})),
+		})
+	return template
+}
+
+func TemplateWithRules() templatev1.Template {
+	validations := `[
```
ksimon1, line 852: can you please use rule struct (e.g. https://github.com/kubevirt/kubevirt-template-validator/blob/1c714f043f18b5957bf3ebebce0226e756adbb75/pkg/validation/rules.go#L32) and then use json.Marshal() to serialize it.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			// set value unfulfilling validation
+			vmi = addDomainResourcesToVMI(vmi, 2, "test", "128M")
+			vm = NewVirtualMachine(vmi, false)
+			vm.ObjectMeta.Annotations = map[string]string{
+				"vm.kubevirt.io/template":           template.Name,
+				"vm.kubevirt.io/template-namespace": template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).To(HaveOccurred(), "Should have given error failing to fulfill validations")
```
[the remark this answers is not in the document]
vatsalparekh: Added a check for error type

tests/validator_test.go
vatsalparekh: The only rule not fulfilled here is the CPU number, but I see that if it's 0, the validator mutates it to 1, which makes the rule fulfilling
Changed the description to match the test

tests/validator_test.go
vatsalparekh: Added error type check

tests/validator_test.go
vatsalparekh: Added error type check

tests/validator_test.go
vatsalparekh: Corrected

tests/validator_test.go
vatsalparekh: But not sure why only on this test, consistently

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+				template = TemplateWithoutRules()
+				template.Namespace = strategy.GetNamespace()
+				Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+				vmi = addDomainResourcesToVMI(vmi, 1, "q35", "64M")
+				vm = NewVirtualMachine(vmi, false)
+				vm.ObjectMeta.Annotations = map[string]string{
+					"vm.kubevirt.io/template":           template.Name,
+					"vm.kubevirt.io/template-namespace": template.Namespace,
+					"vm.kubevirt.io/validations": `[
```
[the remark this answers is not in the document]
vatsalparekh: I will do it in the part when we move template-validator code under ssp-operator after these tests are merged

tests/validator_test.go
vatsalparekh: Added

tests/validator_test.go
vatsalparekh: Added

tests/validator_test.go
vatsalparekh: Added

tests/validator_test.go
vatsalparekh: Will do it in the part when we move template-validator here in the ssp-operator

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			vm       *kubevirtv1.VirtualMachine
+			vmi      *kubevirtv1.VirtualMachineInstance
+			template templatev1.Template
+		)
+
+		BeforeEach(func() {
+			vmi = NewRandomVMIWithBridgeInterface(strategy.GetNamespace())
+		})
+		AfterEach(func() {
+			err := apiClient.Delete(ctx, &template)
```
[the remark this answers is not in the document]
akrejcir: I meant:
```
if template != nil {
    err := apiClient.Delete(ctx, &template)
    ...
}
```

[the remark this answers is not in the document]
akrejcir: I missed that `template` variable is not a pointer. Please change it to a pointer.

tests/validator_test.go
```diff
@@ -329,4 +338,618 @@ var _ = Describe("Template validator", func() {
...
+			}
+		})
+
+		It("should create VM without template", func() {
+			template = TemplateWithMinimalRule()
+			template.Namespace = strategy.GetNamespace()
+			vm = NewVirtualMachine(vmi, false)
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		It("be created from template without rules", func() {
```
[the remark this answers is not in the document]
akrejcir: Now the this test and the one below are the same, no? They are only different in the number of rules used in the template, but they test the same functionality. I think we should change the code to match the description, and test the case when the template has no rules.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
 			Expect(err).ToNot(HaveOccurred(), "SSP status should be deployed.")
 		})
 	})
+
+	Context("Validation rules tests", func() {
+		var (
+			vm       *kubevirtv1.VirtualMachine
+			vmi      *kubevirtv1.VirtualMachineInstance
+			template templatev1.Template
```
akrejcir, line 345: Please use `template *templatev1.Template`.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+				Expect(err).ToNot(HaveOccurred(), "Failed to Delete VM")
+			}
+		})
+
+		It("should create VM without template", func() {
+			vm = NewVirtualMachine(vmi)
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		It("be created from template with minimal rules", func() {
+			template = TemplateWithMinimalRule()
```
akrejcir, line 376: For this test, the original test case in the template validator repository is probably wrong. Can you change the code so that it tests creating a VM from template with no rules? Instead of changing the description.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to create VM")
+		})
+		It("test with template cpu count 0 ", func() {
+			template = TemplateWithRules()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "128M")
```
akrejcir, line 430: The original test case for this was different. It checked that the validation fails if the VM resource does not contain the jsonpath that the rule needs to check.
For example this VM spec does not contain CPU informataion:
```
spec:
  domain:
    devices:
      interfaces:
      - name: default
        bridge: {}
    machine:
      type: "q35"
    resources:
      requests:
        memory: 128M
  networks:
  - name: default
    pod: {}
  terminationGracePeriodSeconds: 0
```
So a rule that checks `jsonpath::.spec.domain.cpu.cores` will fail, because the path does not exist.

In this new test case, the VM contains CPU information `vmi.Spec.Domain.CPU` , it is not `nil`. So this new test does not test the same functionality as the old test.

Please set the CPU to `nil`, change the the `Expect()` to expect and error, and rewrite the test's description to mention that one of the jsonpaths needed for a rule does not exist in the VM.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			Expect(errors.IsInvalid(apiClient.Create(ctx, vm))).To(BeTrue(), "Should match error type because of unfulfilled validations")
+		})
+		It("test with template optional rules unfulfilled", func() {
+			template = TemplateWithRulesOptional()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "128M")
```
akrejcir, line 418: The VM in the original test does not have CPU information. Please remove it from `vmi`: `vmi.Spec.Domain.CPU = nil`. See my other comment for explanation why.

Since the rules are optional, the validation should pass even if no CPU info is provided.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			Expect(apiClient.Create(ctx, vm)).ToNot(HaveOccurred(), "Failed to Create VM")
+		})
+		It("Test template with incorrect rules satisfied", func() {
+			template = TemplateWithIncorrectRules()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "128M")
```
akrejcir, line 442: The VM in original test case does not have CPU info, please remove it. See my other comments why.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			Expect(errors.IsInvalid(apiClient.Create(ctx, vm))).To(BeTrue(), "Should have given the invalid error failing to fulfill validations")
+		})
+		It("Test template with incorrect rules unfulfilled", func() {
+			template = TemplateWithIncorrectRules()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "32M")
```
akrejcir, line 454: The VM in original test case does not have CPU info, please remove it. See my other comments why.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			Expect(errors.IsInvalid(apiClient.Create(ctx, vm))).To(BeTrue(), "Should have given the invalid error failing to fulfill validations")
+		})
+		It("[test_id:2959] Create a VM with memory restrictions violation that succeeds with a warning", func() {
+			template = TemplateWithIncorrectRulesJustWarning()
+			Expect(apiClient.Create(ctx, &template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+
+			vmi = addDomainResourcesToVMI(vmi, 0, "q35", "1G")
```
akrejcir, line 466: The VM in original test case does not have CPU info, please remove it. See my other comments why.

tests/validator_test.go
```diff
@@ -329,4 +337,611 @@ var _ = Describe("Template validator", func() {
...
+			vm = NewVirtualMachine(vmi)
+			vm.ObjectMeta.Annotations = map[string]string{
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			// TODO: Investigate why this test consistently fails without Eventually
+			Eventually(func() error {
+				return apiClient.Create(ctx, vm)
+			}, 180*time.Second).Should(BeNil(), "Failed to create VM")
```
akrejcir, line 475: Use `shortTimeout`.

tests/validator_test.go
akrejcir: You have missed this one.

akrejcir (changes requested):

tests/validator_test.go
vatsalparekh: Corrected

tests/validator_test.go
vatsalparekh: Corrected

tests/validator_test.go
vatsalparekh: Corrected

tests/validator_test.go
vatsalparekh: Corrected

tests/validator_test.go
```diff
@@ -329,4 +337,606 @@ var _ = Describe("Template validator", func() {
...
+		const (
+			TemplateNameAnnotation      = "vm.kubevirt.io/template"
+			TemplateNamespaceAnnotation = "vm.kubevirt.io/template-namespace"
+		)
+
+		BeforeEach(func() {
+			vmi = NewRandomVMIWithBridgeInterface(strategy.GetNamespace())
+			// Create a dummy template to make sure there's no value from an earlier test
+			// And dummy name & namespace for a valid object in Cleanup
+			template = &templatev1.Template{}
```
akrejcir, line 356: Set `template` and `vm` to nil.

tests/validator_test.go
```diff
@@ -329,4 +337,606 @@ var _ = Describe("Template validator", func() {
...
+			template.Namespace = "testDummy"
+		})
+		AfterEach(func() {
+			if template != nil {
+				err := apiClient.Delete(ctx, template)
+				if !errors.IsNotFound(err) {
+					Expect(err).ToNot(HaveOccurred(), "Failed to delete Template")
+				}
+			}
+			err := apiClient.Delete(ctx, vm)
```
akrejcir, line 367: Check for `vm != nil` here. Similarly as above with `template`.

akrejcir (changes requested):
Last comments. Otherwise looks good.

akrejcir:
Looks good. @ksimon1 can you take a quick look too?

ksimon1:
some tests are failing, can you please check them?

PR #71, opened 2020-12-21
Remove workload, flavor, os labels from templates which have older version

vatsalparekh:
remove workload, flavor, os labels from templates which have older version
Fixes # https://bugzilla.redhat.com/show_bug.cgi?id=1859235

**Release note**:

```release-note
SSP operator now removes workload, flavor, os labels from templates which have older version
```

### DIFF
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -2,6 +2,7 @@ package common_templates
 
 import (
 	"fmt"
+	"strings"
 
 	"path/filepath"
 	"sync"
@@ -169,10 +170,15 @@ func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, e
 	for i := range existingTemplates.Items {
 		template := &existingTemplates.Items[i]
 		funcs = append(funcs, func(*common.Request) (common.ResourceStatus, error) {
-			// An empty update func is sufficient because the generic reconcile
-			// logic would add owner annotations and remove ownerReferences
 			return common.CreateOrUpdateClusterResource(request, template,
-				func(newRes, foundRes controllerutil.Object) {})
+				func(_, foundRes controllerutil.Object) {
+					foundTemplate := foundRes.(*templatev1.Template)
+					for key := range foundTemplate.Labels {
+						if strings.HasPrefix(key, "os.template.kubevirt.io/") || strings.HasPrefix(key, "flavor.template.kubevirt.io/") || strings.HasPrefix(key, "workload.template.kubevirt.io/") {
+							delete(foundTemplate.Labels, key)
+						}
+					}
+				})
 		})
 	}
 
--- a/internal/operands/common-templates/reconcile_test.go
+++ b/internal/operands/common-templates/reconcile_test.go
@@ -2,12 +2,15 @@ package common_templates
 
 import (
 	"context"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"strings"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	templatev1 "github.com/openshift/api/template/v1"
-	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	. "kubevirt.io/ssp-operator/internal/test-utils"
@@ -26,16 +29,17 @@ var (
 	operand = GetOperand()
 )
 
+const (
+	namespace = "kubevirt"
+	name      = "test-ssp"
+)
+
 func TestTemplates(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Common Templates Suite")
 }
 
 var _ = Describe("Common-Templates operand", func() {
-	const (
-		namespace = "kubevirt"
-		name      = "test-ssp"
-	)
 
 	var request common.Request
 
@@ -56,11 +60,11 @@ var _ = Describe("Common-Templates operand", func() {
 			Scheme:  s,
 			Context: context.Background(),
 			Instance: &ssp.SSP{
-				TypeMeta: meta.TypeMeta{
+				TypeMeta: metav1.TypeMeta{
 					Kind:       "SSP",
 					APIVersion: ssp.GroupVersion.String(),
 				},
-				ObjectMeta: meta.ObjectMeta{
+				ObjectMeta: metav1.ObjectMeta{
 					Name:      name,
 					Namespace: namespace,
 				},
@@ -107,7 +111,7 @@ var _ = Describe("Common-Templates operand", func() {
 
 	Context("old templates", func() {
 		var (
-			parentTpl *templatev1.Template
+			parentTpl, oldTpl *templatev1.Template
 		)
 
 		BeforeEach(func() {
@@ -116,7 +120,7 @@ var _ = Describe("Common-Templates operand", func() {
 			// might be deployed in a different namespace than the CR, and will be immediately
 			// removed by the GC, the choice to use a template as an owner object was arbitrary
 			parentTpl = &templatev1.Template{
-				ObjectMeta: meta.ObjectMeta{
+				ObjectMeta: metav1.ObjectMeta{
 					Name:      "parent-test-template",
 					Namespace: request.Instance.Spec.CommonTemplates.Namespace,
 				},
@@ -125,23 +129,20 @@ var _ = Describe("Common-Templates operand", func() {
 			key, err := client.ObjectKeyFromObject(parentTpl)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(request.Client.Get(request.Context, key, parentTpl)).ToNot(HaveOccurred())
-		})
 
-		AfterEach(func() {
-			Expect(request.Client.Delete(request.Context, parentTpl)).ToNot(HaveOccurred(), "deletion of parent tempalte failed")
-		})
-
-		It("should replace ownerReferences with owner annotations for older templates", func() {
-			oldTpl := &templatev1.Template{
-				ObjectMeta: meta.ObjectMeta{
+			oldTpl = &templatev1.Template{
+				ObjectMeta: metav1.ObjectMeta{
 					Name:      "test-tpl",
 					Namespace: request.Instance.Spec.CommonTemplates.Namespace,
 					Labels: map[string]string{
-						"template.kubevirt.io/version": "not-latest",
-						"template.kubevirt.io/type":    "base",
+						"template.kubevirt.io/version":         "not-latest",
+						"template.kubevirt.io/type":            "base",
+						"os.template.kubevirt.io/some-os":      "true",
+						"flavor.template.kubevirt.io/test":     "true",
+						"workload.template.kubevirt.io/server": "true",
 					},
 					Annotations: map[string]string{},
-					OwnerReferences: []meta.OwnerReference{{
+					OwnerReferences: []metav1.OwnerReference{{
 						APIVersion: parentTpl.APIVersion,
 						Kind:       parentTpl.Kind,
 						UID:        parentTpl.UID,
@@ -150,12 +151,20 @@ var _ = Describe("Common-Templates operand", func() {
 				},
 			}
 
-			err := request.Client.Create(request.Context, oldTpl)
+			err = request.Client.Create(request.Context, oldTpl)
 			Expect(err).ToNot(HaveOccurred(), "creation of old template failed")
+		})
+
+		AfterEach(func() {
+			Expect(request.Client.Delete(request.Context, oldTpl)).ToNot(HaveOccurred(), "deletion of parent tempalte failed")
+			Expect(request.Client.Delete(request.Context, parentTpl)).ToNot(HaveOccurred(), "deletion of parent tempalte failed")
+		})
+
+		It("should replace ownerReferences with owner annotations for older templates", func() {
 
 			Expect(oldTpl.GetOwnerReferences()).ToNot(BeNil(), "template should have owner reference before reconciliation")
 
-			_, err = operand.Reconcile(&request)
+			_, err := operand.Reconcile(&request)
 			Expect(err).ToNot(HaveOccurred(), "reconciliation in order to update old template failed")
 
 			key, err := client.ObjectKeyFromObject(oldTpl)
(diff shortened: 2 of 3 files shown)


internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +127,55 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+				delete(template.Labels, key)
+			}
+		}
+		if updated {
+			updatedTemplates = append(updatedTemplates, template)
+		}
+	}
+	oldTemplates.Items = updatedTemplates
+	for _, template := range oldTemplates.Items {
+		err = request.Client.Update(request.Context, &template)
```
kwiesmueller, line 171: I think instead of Updates I'd do Patches to the list of resources.
You can build a JSON Patch from the fields that should be removed:
```
[ { "op": "remove", "path": "/metadata/labels/os.template.kubevirt.io\/centos8" }]
```
So I'd collect all the labels that should be removed for each resource (or just all of them) and then apply the patch.
This should be a bit less error prone as it does not care about the previous object. I think Update "could" fail if there was a modification since the `get`.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -100,4 +102,61 @@ var _ = Describe("Common-Templates operand", func() {
...
+		err = request.Client.Create(request.Context, getTestTemplate(oldVersion, "2", "test-namespace"))
+		Expect(err).ToNot(HaveOccurred())
+		err = request.Client.Create(request.Context, getTestTemplate(Version, "3", namespace))
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		var oldTemplates templatev1.TemplateList
+		Expect(request.Client.List(request.Context, &oldTemplates)).ToNot(HaveOccurred())
```
kwiesmueller, line 119: I usually write this with `err := ...` because the wording sounds off (Expect List ToNot HaveOccurred) but that's just my obsession with wording.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -100,4 +102,61 @@ var _ = Describe("Common-Templates operand", func() {
...
 		ExpectResourceExists(newEditRole(), request)
 	})
+
+	It("should remove labels only from old templates", func() {
+		oldVersion := "v0.0.1"
+		err := request.Client.Create(request.Context, getTestTemplate(oldVersion, "1", namespace))
+		Expect(err).ToNot(HaveOccurred())
+		err = request.Client.Create(request.Context, getTestTemplate(oldVersion, "2", "test-namespace"))
+		Expect(err).ToNot(HaveOccurred())
+		err = request.Client.Create(request.Context, getTestTemplate(Version, "3", namespace))
```
kwiesmueller, line 112: I think this is what makes the test fail. The testTemplate does not define any default variant which the other test checks for. Either you should put the creation into a beforeEach and clean them up in aftereach or do the cleanup in here.

tests/commonTemplates_test.go
```diff
@@ -212,4 +218,66 @@ var _ = Describe("Common templates", func() {
 			table.Entry("[test_id:4770]golden image NS", &goldenImageNS),
 		)
 	})
+
+	Context("update old templates", func() {
+		It("should remove labels from old templates", func() {
```
kwiesmueller, line 223: I'm not sure if this test actually does anything in the default CI environment where no old labels are present.
Is there an upgrade test environment? If yes nevermind.
Maybe still fail or at least warn this test if no old templates are found, as it indicates that the case is not actually tested?

tests/tests_suite_test.go
```diff
@@ -89,6 +90,19 @@ func (s *newSspStrategy) Init() {
 	Eventually(func() error {
 		return apiClient.Create(ctx, newSsp)
 	}, timeout, time.Second).ShouldNot(HaveOccurred())
+
+	Eventually(func() error {
+		return apiClient.Create(ctx, getTestTemplate("3", strategy.GetTemplatesNamespace(), commonTemplates.Version))
```
kwiesmueller, line 95: Ah, I think that answers my previous comment and also explains the failed default-variant test mentioned above.

kwiesmueller:
Also, is there a test ensuring that there is no old version of the operator that races with this change?
Like a botched update or even during an update?

tests/tests_suite_test.go
omeryahud: Why do we need to create these templates at test suite init? If they are only relevant for a specific test, they should be created and cleaned up in the context of the test

vatsalparekh:
> Also, is there a test ensuring that there is no old version of the operator that races with this change?
> Like a botched update or even during an update?

Good point
I think HCO is supposed to remove the old Operator & CR when it is doing the upgrade
But not sure where this test should be
@omeryahud Any ideas?

omeryahud:
> > Also, is there a test ensuring that there is no old version of the operator that races with this change?
> > Like a botched update or even during an update?
> 
> Good point
> I think HCO is supposed to remove the old Operator & CR when it is doing the upgrade
> But not sure where this test should be
> @omeryahud Any ideas?

Yes, I think the upgrade test covers this, as it pauses the CRs of the old operator in order to take ownership of the resources.
One thing I am unsure of is if the new operator takes ownership of existing templates during an upgrade, I will double check and propose a PR to fix it if not

internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +127,52 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+	var oldTemplates templatev1.TemplateList
+	err = request.Client.List(request.Context, &oldTemplates, &opts)
+	if err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error during listing templates, %v", err))
+		return common.ResourceStatus{}, err
+	}
+
+	for _, template := range oldTemplates.Items {
+		needsUpdate := false
+		newTemplate := template.DeepCopy()
```
kwiesmueller, line 158: I don't think the DeepCopy and MergeFrom are necessary.
You could just store a list of labels that need to be removed and then go over them and build the json patch directly.
Would be a few less/smaller allocations than DeepCopy.
But it should be okay this way.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -69,35 +75,67 @@ var _ = Describe("Common-Templates operand", func() {
...
+		var oldTemplates templatev1.TemplateList
+		versionRequirement, err := labels.NewRequirement("template.kubevirt.io/version", selection.Equals, []string{Version})
+		Expect(err).ToNot(HaveOccurred())
+		labelsSelector := labels.NewSelector().Add(*versionRequirement)
+		opts := client.ListOptions{
+			LabelSelector: labelsSelector,
+		}
+		Expect(request.Client.List(request.Context, &oldTemplates, &opts)).ToNot(HaveOccurred())
+		for _, template := range oldTemplates.Items {
+			Expect(template.Labels["os.template.kubevirt.io/some-os"]).To(Equal("true"), "os.template.kubevirt.io should not be empty")
```
omeryahud, line 134: Does this test pass? I don't see where you create a new template with those labels

tests/commonTemplates_test.go
```diff
@@ -212,4 +218,60 @@ var _ = Describe("Common templates", func() {
...
+			opts := client.ListOptions{
+				LabelSelector: labelsSelector,
+			}
+
+			var oldTemplates templatev1.TemplateList
+			err = apiClient.List(ctx, &oldTemplates, &opts)
+			Expect(err).To(BeNil())
+
+			for _, template := range oldTemplates.Items {
+				Expect(template.Labels["os.template.kubevirt.io/some-os"]).To(Equal("true"), "os.template.kubevirt.io should not be empty")
```
omeryahud, line 269: Same question as in the unit-tests, no template has `some-os` as a label

omeryahud:
> > > Also, is there a test ensuring that there is no old version of the operator that races with this change?
> > > Like a botched update or even during an update?
> > 
> > 
> > Good point
> > I think HCO is supposed to remove the old Operator & CR when it is doing the upgrade
> > But not sure where this test should be
> > @omeryahud Any ideas?
> 
> Yes, I think the upgrade test covers this, as it pauses the CRs of the old operator in order to take ownership of the resources.
> One thing I am unsure of is if the new operator takes ownership of existing templates during an upgrade, I will double check and propose a PR to fix it if not

Here's a PR to replace the owner reference with owner annotations on older templatesL https://github.com/kubevirt/ssp-operator/pull/72

internal/operands/common-templates/reconcile_test.go
vatsalparekh: It was created here https://github.com/kubevirt/ssp-operator/pull/71/files#diff-6c919a0ae6f12f061dafb6d58a1d1c60cf6717bdfc2ab7c924ebf6a18a439939R39
But was redundant because those labels do not apply to latest and actual templates
Updated now

tests/commonTemplates_test.go
vatsalparekh: Corrected now like unit-tests

internal/operands/common-templates/reconcile.go
omeryahud: `DeepCopy` is redundant here, since you intend to `Patch` anyway

internal/operands/common-templates/reconcile.go
```diff
@@ -122,6 +127,52 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+		for key := range template.Labels {
+			if strings.HasPrefix(key, "os.template.kubevirt.io/") || strings.HasPrefix(key, "flavor.template.kubevirt.io/") || strings.HasPrefix(key, "workload.template.kubevirt.io/") {
+				needsUpdate = true
+				delete(newTemplate.Labels, key)
+			}
+		}
+		if needsUpdate {
+			err = request.Client.Patch(request.Context, newTemplate, client.MergeFrom(&template))
```
omeryahud, line 166: Will a `merge-patch` remove labels? I am not sure if it will just merge the label maps

omeryahud (commented):
Tests look good, let's review the code again once #72 is merged and this is rebased

internal/operands/common-templates/reconcile.go
```diff
@@ -172,7 +173,14 @@ func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, e
 			// An empty update func is sufficient because the generic reconcile
```
omeryahud, line 173: This comment of mine can be removed now (the update func is no longer empty)

tests/commonTemplates_test.go
```diff
@@ -289,9 +296,54 @@ var _ = Describe("Common templates", func() {
...
+				Expect(err).ToNot(HaveOccurred(), "failed to read template object key")
+				err = apiClient.Get(ctx, key, updatedTpl)
+				if err != nil {
+					return false
+				}
+				if updatedTpl.Labels["os.template.kubevirt.io/some-os"] == "" && updatedTpl.Labels["flavor.template.kubevirt.io/test"] == "" && updatedTpl.Labels["workload.template.kubevirt.io/server"] == "" && updatedTpl.Labels["template.kubevirt.io/type"] == "base" && updatedTpl.Labels["template.kubevirt.io/version"] == "non-latest" {
+					return true
+				}
+				return false
+			}, shortTimeout).Should(BeTrue(), "ownerReference was not replaced by owner annotations on the old template")
```
omeryahud, line 315: Description text should be changed

tests/commonTemplates_test.go
```diff
@@ -289,9 +296,54 @@ var _ = Describe("Common templates", func() {
...
+			Eventually(func() bool {
+				updatedTpl := &templatev1.Template{}
+				key, err := client.ObjectKeyFromObject(oldTpl)
+				Expect(err).ToNot(HaveOccurred(), "failed to read template object key")
+				err = apiClient.Get(ctx, key, updatedTpl)
+				if err != nil {
+					return false
+				}
+				if updatedTpl.Labels["os.template.kubevirt.io/some-os"] == "" && updatedTpl.Labels["flavor.template.kubevirt.io/test"] == "" && updatedTpl.Labels["workload.template.kubevirt.io/server"] == "" && updatedTpl.Labels["template.kubevirt.io/type"] == "base" && updatedTpl.Labels["template.kubevirt.io/version"] == "non-latest" {
+					return true
```
akrejcir, line 312: The value of label `template.kubevirt.io/version` should be `not-latest`.

akrejcir (changes requested):
After rebasing on latest master and fixing the test as commented, this PR works as expected.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -169,5 +178,47 @@ var _ = Describe("Common-Templates operand", func() {
...
+			err = request.Client.Get(request.Context, key, updatedTpl)
+			Expect(err).ToNot(HaveOccurred(), "failed fetching updated template")
+
+			Expect(updatedTpl.Labels["os.template.kubevirt.io/some-os"]).To(Equal(""), "os.template.kubevirt.io should be empty")
+			Expect(updatedTpl.Labels["flavor.template.kubevirt.io/test"]).To(Equal(""), "flavor.template.kubevirt.io should be empty")
+			Expect(updatedTpl.Labels["workload.template.kubevirt.io/server"]).To(Equal(""), "workload.template.kubevirt.io should be empty")
+			Expect(updatedTpl.Labels["template.kubevirt.io/type"]).To(Equal("base"), "template.kubevirt.io/type should equal base")
+			Expect(updatedTpl.Labels["template.kubevirt.io/version"]).To(Equal("not-latest"), "template.kubevirt.io/version should equal not-latest")
+		})
+		It("should not remove labels from latest templates", func() {
```
akrejcir, line 198: You forgot to call `_, err := operand.Reconcile(&request)` in this test.

tests/commonTemplates_test.go
```diff
@@ -299,9 +306,54 @@ var _ = Describe("Common templates", func() {
...
+				if err != nil {
+					return false
+				}
+				if updatedTpl.Labels["os.template.kubevirt.io/some-os"] == "" && updatedTpl.Labels["flavor.template.kubevirt.io/test"] == "" && updatedTpl.Labels["workload.template.kubevirt.io/server"] == "" && updatedTpl.Labels["template.kubevirt.io/type"] == "base" && updatedTpl.Labels["template.kubevirt.io/version"] == "not-latest" {
+					return true
+				}
+				return false
+			}, shortTimeout).Should(BeTrue(), "ownerReference was not replaced by owner annotations on the old template")
+		})
+		It("should continue to have labels on latest templates", func() {
```
akrejcir, line 327: This test needs to trigger reconciliation: `triggerReconciliation()`.

tests/commonTemplates_test.go
```diff
@@ -299,9 +306,54 @@ var _ = Describe("Common templates", func() {
...
+			baseRequirement, err := labels.NewRequirement("template.kubevirt.io/type", selection.Equals, []string{"base"})
+			Expect(err).To(BeNil())
+			var versionRequirement *labels.Requirement
+			versionRequirement, err = labels.NewRequirement("template.kubevirt.io/version", selection.Equals, []string{commonTemplates.Version})
```
akrejcir, line 331: No need to separately declare the variable `versionRequirement`, you can use:
`versionRequirement, err := labels.NewRequirement(...)`

tests/commonTemplates_test.go
```diff
@@ -299,9 +306,54 @@ var _ = Describe("Common templates", func() {
...
+			Expect(err).To(BeNil())
+			labelsSelector := labels.NewSelector().Add(*baseRequirement, *versionRequirement)
+			opts := client.ListOptions{
+				LabelSelector: labelsSelector,
+			}
```
akrejcir, line 336: To make the test more efficient, it could list the templates only in one namespace.
```
opts := client.ListOptions{
    LabelSelector: labelsSelector,
    Namespace: strategy.GetTemplatesNamespace(),
}
```

akrejcir (changes requested):

omeryahud (approved):
/lgtm
/approve

akrejcir:
I gave LGTM too soon, the test is not correct. I will post a fix.

omeryahud:
> I gave LGTM too soon, the test is not correct. I will post a fix.

Which test are you referring to?

akrejcir:
Functional test: "should continue to have labels on latest templates".

akrejcir:
The fix is in this PR: #78

PR #72, opened 2020-12-27
Replace owner references with owner annotations for previously deployed templates during an upgrade

omeryahud:
The operator will now replace the ownerReferences on older templates with owner annotations pointing to the SSP CR that is reconciling them to prevent them from being removed when their old owner is deleted.

We need this PR to make sure that existing templates are not deleted from the cluster when the old CR is being removed

**Release note**:

```release-note
None
```

### DIFF
--- a/internal/common/resource.go
+++ b/internal/common/resource.go
@@ -5,8 +5,11 @@ import (
 	"reflect"
 
 	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 
 	libhandler "github.com/operator-framework/operator-lib/handler"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
@@ -93,8 +96,41 @@ func createOrUpdate(request *Request, resource controllerutil.Object, isClusterR
 
 func setOwner(request *Request, resource controllerutil.Object, isClusterRes bool) error {
 	if isClusterRes {
-		return libhandler.SetOwnerAnnotations(request.Instance, resource)
+		err := libhandler.SetOwnerAnnotations(request.Instance, resource)
+		if err != nil {
+			return err
+		}
+
+		// Removing ownerReferences for objects prior to creation
+		resource.SetOwnerReferences(nil)
+
+		// Removing ownerReferences for existing objects
+		existingRes := newEmptyResource(resource)
+		key, err := client.ObjectKeyFromObject(resource)
+		if err != nil {
+			return err
+		}
+
+		err = request.Client.Get(request.Context, key, existingRes)
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		} else if err == nil {
+			if len(existingRes.GetOwnerReferences()) > 0 {
+				request.Logger.Info(fmt.Sprintf("Patching %s to remove ownerReferences", key))
+				patch := client.RawPatch(types.JSONPatchType,
+					[]byte(`[{"op": "replace", "path": "/metadata/ownerReferences", "value": []}, {"op": "remove", "path": "/metadata/ownerReferences"}]`))
+				err = request.Client.Patch(request.Context, existingRes, patch)
+
+				if err != nil {
+					return err
+				}
+			}
+		}
+
+		return nil
 	} else {
+		delete(resource.GetAnnotations(), libhandler.NamespacedNameAnnotation)
+		delete(resource.GetAnnotations(), libhandler.TypeAnnotation)
 		return controllerutil.SetControllerReference(request.Instance, resource, request.Scheme)
 	}
 }
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -10,9 +10,12 @@ import (
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/operands"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
@@ -63,7 +66,15 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ResourceS
 		reconcileViewRoleBinding,
 		reconcileEditRole,
 	}
+
+	oldTemplateFuncs, err := reconcileOlderTemplates(request)
+	if err != nil {
+		return nil, err
+	}
+
+	funcs = append(funcs, oldTemplateFuncs...)
 	funcs = append(funcs, reconcileTemplatesFuncs(request)...)
+
 	return common.CollectResourceStatus(request, funcs...)
 }
 
@@ -121,6 +132,51 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
 			foundRole := foundRes.(*rbac.ClusterRole)
 			foundRole.Rules = newRole.Rules
 		})
+}
+
+func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, error) {
+	// Append functions to take ownership of previously deployed templates during an upgrade
+	existingTemplates := &templatev1.TemplateList{}
+	templatesSelector := func() labels.Selector {
+		tplSelector := labels.NewSelector()
+		tplBaseRequirement, err := labels.NewRequirement("template.kubevirt.io/type", selection.Equals, []string{"base"})
+		if err != nil {
+			panic("Failed creating label selector for 'template.kubevirt.io/type=base")
+		}
+
+		// Only fetching older templates  to prevent duplication of API calls
+		tplVersionRequirement, err := labels.NewRequirement("template.kubevirt.io/version", selection.NotEquals, []string{Version})
+		if err != nil {
+			panic("Failed creating label selector for 'template.kubevirt.io/version")
+		}
+
+		tplSelector = tplSelector.Add(*tplBaseRequirement, *tplVersionRequirement)
+
+		return tplSelector
+	}()
+
+	err := request.Client.List(request.Context, existingTemplates, &client.ListOptions{
+		LabelSelector: templatesSelector,
+		Namespace:     request.Instance.Spec.CommonTemplates.Namespace,
+	})
+
+	// There might not be any templates (in case of a fresh deployment), so a NotFound error is accepted
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	}
+
+	funcs := make([]common.ReconcileFunc, 0, len(existingTemplates.Items))
+	for i := range existingTemplates.Items {
+		template := &existingTemplates.Items[i]
+		funcs = append(funcs, func(*common.Request) (common.ResourceStatus, error) {
+			// An empty update func is sufficient because the generic reconcile
+			// logic would add owner annotations and remove ownerReferences
+			return common.CreateOrUpdateClusterResource(request, template,
+				func(newRes, foundRes controllerutil.Object) {})
+		})
+	}
+
+	return funcs, nil
 }
 
 func reconcileTemplatesFuncs(request *common.Request) []common.ReconcileFunc {
--- a/internal/operands/common-templates/reconcile_test.go
+++ b/internal/operands/common-templates/reconcile_test.go
@@ -6,6 +6,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	templatev1 "github.com/openshift/api/template/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -14,8 +15,10 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	libhandler "github.com/operator-framework/operator-lib/handler"
 	ssp "kubevirt.io/ssp-operator/api/v1beta1"
 	"kubevirt.io/ssp-operator/internal/common"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var (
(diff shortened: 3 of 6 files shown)


internal/common/resource.go
```diff
@@ -93,8 +97,45 @@ func createOrUpdate(request *Request, resource controllerutil.Object, isClusterR
...
+				return err
+			}
+		} else {
+			if existingRes.GetOwnerReferences() != nil && len(existingRes.GetOwnerReferences()) > 0 {
+				patch := client.RawPatch(types.JSONPatchType,
+					[]byte(`[{"op": "replace","path": "/metadata/ownerReferences", "value": []}, {"op": "remove","path": "/metadata/ownerReferences"}]`))
+				err = request.Client.Patch(request.Context, existingRes, patch)
+
+				// If the OwnerReferences field does not exist for 'existingRes' that should be accepted as well,
+				// I couldn't find the typed error for this case (at least for unit-tests), so I went with
```
kwiesmueller, line 127: There is no explicit typed error for this, but the jsonPatch wraps the error to this form here:
https://github.com/kubernetes/kubernetes/blob/0a839c6c3b9489716e216b4fc9dd2f9c17948c16/vendor/github.com/evanphx/json-patch/patch.go#L401
You could use that ErrMissing here, but I don't think if it should be a dependency for us as it's not a dep directly on K8s but a third_party library.

internal/common/resource.go
```diff
@@ -93,8 +97,45 @@ func createOrUpdate(request *Request, resource controllerutil.Object, isClusterR
...
+		// Removing ownerReferences for existing objects
+		existingRes := newEmptyResource(resource)
+		key, err := client.ObjectKeyFromObject(resource)
+		if err != nil {
+			return err
+		}
+
+		err = request.Client.Get(request.Context, key, existingRes)
+		if err != nil {
+			if !errors.IsNotFound(err) {
```
kwiesmueller, line 117: You could combine those checks into one line `err != nil && !errors.IsNotFound(err)` to remove one level of intending.

internal/common/resource.go
```diff
@@ -93,8 +97,45 @@ func createOrUpdate(request *Request, resource controllerutil.Object, isClusterR
...
+				return err
+			}
+		} else {
+			if existingRes.GetOwnerReferences() != nil && len(existingRes.GetOwnerReferences()) > 0 {
```
kwiesmueller, line 121: Same here, using `else if` would move the code one level left and make it a bit easier to read.

internal/common/resource.go
```diff
@@ -93,8 +97,45 @@ func createOrUpdate(request *Request, resource controllerutil.Object, isClusterR
...
+				patch := client.RawPatch(types.JSONPatchType,
+					[]byte(`[{"op": "replace","path": "/metadata/ownerReferences", "value": []}, {"op": "remove","path": "/metadata/ownerReferences"}]`))
```
kwiesmueller, line 123: Do we need to replace and remove?
It seems redundant or am I missing something here?

internal/operands/common-templates/reconcile.go
```diff
@@ -123,7 +132,59 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
 		})
 }
 
-func reconcileTemplatesFuncs(request *common.Request) []common.ReconcileFunc {
+func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, error) {
+	// Append functions to take ownership of previously deployed templates during an upgrade
+	var existingTemplates templatev1.TemplateList
+	templatesSelector := func() labels.Selector {
```
kwiesmueller, line 138: You could move this into its own function instead of an inline definition.
Like a `buildTemplateSelector`, then you can also handle the errors without panics (but they seem like assertions so it should be fine).
Tbh. I don't understand right now how this actually works, as LabelSelector takes an interface but you pass it a `func() labels.Selector`.

tests/commonTemplates_test.go
```diff
@@ -212,4 +215,63 @@ var _ = Describe("Common templates", func() {
...
+					OwnerReferences: []v1.OwnerReference{{
+						APIVersion: "ssp.kubevirt.io/v1beta1",
+						Kind:       "SSP",
+						Name:       ssp.Name,
+						UID:        ssp.UID,
+					}},
+				},
+			}
+
+			err := apiClient.Create(ctx, &oldTpl)
```
kwiesmueller, line 239: Could you delete the resource once the test is done?
So it doesn't collide with the defaults etc.

internal/common/resource.go
omeryahud: Done

internal/operands/common-templates/reconcile.go
omeryahud: I don't pass a function, I define it an immediately execute it, notice the `()` in line 161. I do it just to scope the creation and definition of the label selector (makes it more readable IMO).

To you other point, I don't think creating a function is required, because the label selectors here are specific enough

tests/commonTemplates_test.go
omeryahud: Done

internal/common/resource.go
omeryahud: Simply removing gave inconsistent errors (both when `ownerReferences` exist and not), so the workaround I chose here is to first `replace` with an empty slice (create if doesn't exist), then `remove`, this seemed to give a consistent error flow that I can follow

omeryahud:
@vatsalparekh Would appreciate a review, thanks!

omeryahud:
Having an issue with a testcase, moving back to WIP until it is resolved

internal/operands/common-templates/reconcile.go
kwiesmueller: Ah, I overlooked the ().
In general having functionality inlined like this and immediately executed is something I rarely see in Golang so far, I think that's more a JavaScript thing.
To me it's more confusing because I didn't expect it, but if it's a common thing in the project I don't disagree.

I'm not sure right now if there is a runtime difference between the two (how the compiler optimizes inline functions), but it shouldn't matter. Mainly I was surprised because I'm only used to inline functions like this for defers or to pass the function.

internal/common/resource.go
kwiesmueller: Makes sense, thanks.

internal/operands/common-templates/reconcile.go
```diff
@@ -123,7 +139,52 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+	}()
+
+	err := request.Client.List(request.Context, existingTemplates, &client.ListOptions{
+		LabelSelector: templatesSelector,
+		Namespace:     request.Instance.Spec.CommonTemplates.Namespace,
+	})
+
+	// There might not be any templates (in case of a fresh deployment), so a NotFound error is accepted
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
```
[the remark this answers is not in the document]
omeryahud: IIRC `NotFound` is returned if `List` does not find any resources

tests/commonTemplates_test.go
```diff
@@ -212,4 +215,83 @@ var _ = Describe("Common templates", func() {
...
+
+			// Add dummy annotation to SSP resource to initiate reconciliation
+			updateSsp(func(foundSsp *sspv1beta1.SSP) {
+				if foundSsp.GetAnnotations() == nil {
+					foundSsp.SetAnnotations(map[string]string{})
+				}
+
+				foundSsp.GetAnnotations()["dummy"] = "dummy"
+			})
+
```
vatsalparekh, line 269: The part in this `It` block till now seems good to be moved to a `BeforeEach` for this context
All the tests in this `Context` are going to need a dummy template and a reconciliation trigger

tests/commonTemplates_test.go
```diff
@@ -212,4 +215,83 @@ var _ = Describe("Common templates", func() {
...
+			// Template should eventually be updated by the operator
+			Eventually(func() bool {
```
vatsalparekh, line 271: I would rely on `waitUntilDeployed()` in a `BeforeEach` block than `Eventually`

tests/commonTemplates_test.go
```diff
@@ -212,4 +215,83 @@ var _ = Describe("Common templates", func() {
...
+				}
+
+				return false
+			}, shortTimeout).Should(BeTrue(), "ownerReference was not replaced by owner annotations on the old template")
+
+			// Cleanup
+			Expect(apiClient.Delete(ctx, oldTpl)).ToNot(HaveOccurred(), "deletion of dummy old template failed")
+			Expect(apiClient.Delete(ctx, ownerTemplate)).ToNot(HaveOccurred(), "deletion of dummy owner template failed")
+			updateSsp(func(foundSsp *sspv1beta1.SSP) {
+				delete(foundSsp.GetAnnotations(), "dummy")
```
vatsalparekh, line 293: Like `BeforeEach`, this seems like good to be moved to `AfterEach`

internal/operands/common-templates/reconcile.go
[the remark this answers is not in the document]
vatsalparekh: Also, if we don't care about this template anymore, wouldn't passing an error reque it to reconciliation
Confused here between `nil` and `NotFound` err

vatsalparekh (changes requested):

internal/operands/common-templates/reconcile.go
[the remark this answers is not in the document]
omeryahud: Sorry, I couldn't follow what you wanted to change here, this condition returns an error when an `err` which was not `NotFound` occurred, if a `NotFound` error occurred we should keep going and return an empty `funcs` slice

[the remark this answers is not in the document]
omeryahud: And if a different error happened, it means that templates might still exist and we have to try again, hence returning `err` unless we are confident there are none (`NotFound` error)

internal/operands/common-templates/reconcile_test.go
```diff
@@ -100,4 +104,51 @@ var _ = Describe("Common-Templates operand", func() {
...
+				Namespace: request.Instance.Spec.CommonTemplates.Namespace,
+				Labels: map[string]string{
+					"template.kubevirt.io/version": "not-latest",
+					"template.kubevirt.io/type":    "base",
+				},
+				Annotations: map[string]string{},
+				OwnerReferences: []meta.OwnerReference{{
+					APIVersion: parentTpl.APIVersion,
+					Kind:       parentTpl.Kind,
+					UID:        parentTpl.UID,
```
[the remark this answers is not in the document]
omeryahud: IIRC `Create` updates the pointer that was passed in, I will verify

tests/commonTemplates_test.go
```diff
@@ -212,4 +215,83 @@ var _ = Describe("Common templates", func() {
...
+				tpl := &templatev1.Template{
+					ObjectMeta: v1.ObjectMeta{
+						Name:      "owner-template",
+						Namespace: ssp.Spec.CommonTemplates.Namespace,
+					},
+				}
+				Expect(apiClient.Create(ctx, tpl)).ToNot(HaveOccurred(), "failed to create dummy owner for an old template")
+				// key, err := client.ObjectKeyFromObject(tpl)
+				// Expect(err).ToNot(HaveOccurred(), "failed to read template object key")
+				// Expect(apiClient.Get(ctx, key, tpl)).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
omeryahud: Yes, will remove

tests/commonTemplates_test.go
omeryahud: I didn't want to add `BeforeEach` and `AfterEach` until they are needed, in this case there is a single test in the `Context` so it is not needed, but it does make sense to make it more future proof, I will update

tests/commonTemplates_test.go
omeryahud: `waitUntilDeployed()` is already called by an upper-level `BeforaEach` in line 58

tests/commonTemplates_test.go
omeryahud: Done

vatsalparekh (approved):
/lgtm
/approve

internal/operands/common-templates/reconcile.go
```diff
@@ -123,6 +134,51 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
...
+	}
+
+	funcs := make([]common.ReconcileFunc, 0, len(existingTemplates.Items))
+	for i := range existingTemplates.Items {
+		template := &existingTemplates.Items[i]
+		funcs = append(funcs, func(*common.Request) (common.ResourceStatus, error) {
+			// An empty update func is sufficient because the generic reconcile
+			// logic would add owner annotations and remove ownerReferences
```
akrejcir, line 173: Adding owner annotations means that the SSP resource will own the old templates. Should the old templates be removed in `Cleanup()` method? So the meaning of owner annotations is the same for all resources.

internal/operands/common-templates/reconcile.go
akrejcir: I also find it a little unusual. No need to change this now, but in the future you can create a new scope using `{ }`.
```
var templatesSelector labels.Selector
{
    tplSelector := labels.NewSelector()
    ...
    templatesSelector = tplSelector
}
```

akrejcir (commented):
I have noticed that the old templates are not removed when SSP is removed. Should we open a new PR for it?

internal/operands/common-templates/reconcile.go
vatsalparekh: I'll include this as part of #71

internal/operands/common-templates/reconcile.go
omeryahud: You are right! I forgot that is even an option

internal/common/resource.go
```diff
@@ -93,8 +96,41 @@ func createOrUpdate(request *Request, resource controllerutil.Object, isClusterR
...
+
+		err = request.Client.Get(request.Context, key, existingRes)
+		if err != nil && !errors.IsNotFound(err) {
+			return err
+		} else if err == nil {
+			if len(existingRes.GetOwnerReferences()) > 0 {
+				request.Logger.Info(fmt.Sprintf("Patching %s to remove ownerReferences", key))
+				patch := client.RawPatch(types.JSONPatchType,
+					[]byte(`[{"op": "replace", "path": "/metadata/ownerReferences", "value": []}, {"op": "remove", "path": "/metadata/ownerReferences"}]`))
+				err = request.Client.Patch(request.Context, existingRes, patch)
```
akrejcir, line 122: Could you explain why this `Patch` call is needed? I thought it would be enough to clear owner references on the resource: `resource.SetOwnerReferences(nil)`.
Then in the `createOrUpdate` function the owner references of the found resource are rewritten.

omeryahud: I tried that before, but it did not work, I suspected the `patchStrategy` of `ownerReferences`, I will verify it again

akrejcir: Does Update API call use patch strategy?

omeryahud: After verifying again, it does not, `SetOwnerReferences(nil)` before an `Update` should be enough

omeryahud: I think @vatsalparekh and I experienced similar issue

PR #73, opened 2020-12-30
Adding important log messages

omeryahud:
Adding some more log messages to make operator logs more informative

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -91,6 +91,7 @@ type SSPReconciler struct {
 
 func (r *SSPReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	reqLogger := r.Log.WithValues("ssp", req.NamespacedName)
+	reqLogger.V(1).Info("Starting reconciliation...")
 
 	ctx := context.Background()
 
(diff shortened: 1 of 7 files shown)


akrejcir:
Some of these messages could probably use debug level. The logging library is confusing. Debug logging is done like: `Logger.V(1).Info(...)`.

We can change it in the future.

omeryahud:
> Some of these messages could probably use debug level. The logging library is confusing. Debug logging is done like: `Logger.V(1).Info(...)`.
> 
> We can change it in the future.

Yea, that's a good point, I will set them as debug

kwiesmueller:
Actually [I started defining operator names as well right now and am conflicting with this](https://github.com/kubevirt/ssp-operator/pull/92/files#diff-f6d56152bde2c23e8384a420483683cbfc8b9d3152bd9c104914c737b8a99cf3R46).
Wasn't sure so far if I should extend the Operand interface, so thanks for that bump.

The names I defined are e.g. `common-templates` which is more coherent with the naming style of our kubernetes resources.
Would you be good with changing the names or prefer me to use something else?
In the HCO PRs they also seem to go with the lower-case format.

omeryahud:
> Actually [I started defining operator names as well right now and am conflicting with this](https://github.com/kubevirt/ssp-operator/pull/92/files#diff-f6d56152bde2c23e8384a420483683cbfc8b9d3152bd9c104914c737b8a99cf3R46).
> Wasn't sure so far if I should extend the Operand interface, so thanks for that bump.
> 
> The names I defined are e.g. `common-templates` which is more coherent with the naming style of our kubernetes resources.
> Would you be good with changing the names or prefer me to use something else?
> In the HCO PRs they also seem to go with the lower-case format.

Hi Kevin, feel free to change them to the lower-case format

PR #77, opened 2021-01-07
Add observedGeneration to the SSP status

akrejcir:
The `observedGeneration` field can be used to check if the operator has seen latest changes to the SSP resource.

**Release note**:
```release-note
Added 'status.observed_generation' field.
```

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -1,0 +1,75 @@
+package tests
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	lifecycleapi "kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
+
+	sspv1beta1 "kubevirt.io/ssp-operator/api/v1beta1"
+)
+
+var _ = Describe("Observed generation", func() {
+	BeforeEach(func() {
+		strategy.SkipSspUpdateTestsIfNeeded()
+	})
+
+	AfterEach(func() {
+		strategy.RevertToOriginalSspCr()
+		waitUntilDeployed()
+	})
+
+	It("after deployment observedGeneration equals generation", func() {
+		ssp := getSsp()
+		Expect(ssp.Status.ObservedGeneration).To(Equal(ssp.Generation))
+	})
+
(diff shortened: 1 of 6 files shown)


omeryahud (commented):
/lgtm

tests/misc_test.go
```diff
@@ -0,0 +1,75 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		defer watch.Stop()
+
+		var newValidatorReplicas int32 = 0
+		updateSsp(func(foundSsp *sspv1beta1.SSP) {
+			foundSsp.Spec.TemplateValidator.Replicas = &newValidatorReplicas
+		})
+
+		// Watch changes until above change
+		err = WatchChangesUntil(watch, func(updatedSsp *sspv1beta1.SSP) bool {
```
kwiesmueller, line 38: Is this test stable enough? It feels like it could be source for flakes.
In cases where the operator wins and is faster than the test code, this would timeout and fail, right?

You could instead record the ObservedGeneration before sending the update and comparing it to the Observed after the upgrade.
But I'm not sure right now if that would be better.

akrejcir: The function `WatchChangesUntil` uses kubernetes watch API call. Is it possible that this watch skips an event? Or that returned events are out of order?

I assumed that the watch returns a stream of update events in order as they happen, so it does not matter how fast the operator is, the test should not miss an update.

kwiesmueller: Ah, I see. And because the ObservedGeneration gets changed as part of the status update which happens later it should be fine. Okay then, this should be fine.
Just to be sure, I'm trying to figure out if watch events are really guaranteed to be in order, so far I haven't found a source that explicitly says they are.
Thanks for clarifying.

akrejcir: Good point, I will also try to find out if the order is guaranteed.

omeryahud: Was this resolved?

akrejcir: I'm reasonably sure, that the events are ordered. But I did not ask anyone form k8s devs, nor checked the source code.
I think we can mark this as resolved. It is just a test and when it starts flaking, we will notice.

PR #83, opened 2021-01-11
[Bugfix] Give regular users permissions to view os images DataVolumes

omeryahud:
Allows regular users to view DataVolumes in the os-images namespace, main usecase is to follow the upload status of the image

**Which issue(s) this PR fixes** 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1913717

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/tests_common_test.go
+++ b/tests/tests_common_test.go
@@ -1,14 +1,18 @@
 package tests
 
 import (
-	"github.com/onsi/ginkgo"
+	"fmt"
 	"reflect"
 	"time"
+
+	"github.com/onsi/ginkgo"
 
 	. "github.com/onsi/gomega"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"github.com/operator-framework/operator-lib/handler"
+	authv1 "k8s.io/api/authorization/v1"
 	core "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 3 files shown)


tests/tests_common_test.go
```diff
@@ -204,3 +208,27 @@ func getResourceKey(obj controllerutil.Object) client.ObjectKey {
...
+				Group:       resGroup,
+				Version:     resVersion,
+				Resource:    resKind,
+				Subresource: resSubresource,
+				Name:        resName,
+			},
+		},
+	}, metav1.CreateOptions{})
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	ExpectWithOffset(1, sar.Status.Reason).ToNot(BeEmpty(),
```
vatsalparekh, line 229: I don't think this is an correct assertion
Reason field can be and sometimes is empty even if the Status is allowed

vatsalparekh (changes requested):
Just a small nit, looks good otherwise

tests/tests_common_test.go
omeryahud: Oh, good catch, I wanted to remove that and forgot, will remove now

omeryahud: Done

tests/tests_common_test.go
```diff
@@ -204,3 +208,25 @@ func getResourceKey(obj controllerutil.Object) client.ObjectKey {
 		Name:      obj.GetName(),
 	}
 }
+
+func expectUserCan(user string, groups []string, verb, resNamespace, resGroup, resVersion, resKind, resName, resSubresource string) {
```
akrejcir, line 212: It can be confusing to have a function with 9 parameters. Can you use one `*authv1.ResourceAttributes` instead of the last 7 string parameters?

akrejcir (commented):
Just a minor code style comment. You could fix it in #84 instead.

/lgtm

akrejcir:
Also, can you please remove the following text from the PR's description? 
```
 (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged)
```

tests/tests_common_test.go
omeryahud: Sure, will handle it in #84

PR #84, opened 2021-01-11
Make os images tests more aesthetic

omeryahud:
Reuse code in the os-images namespace permission tests, this is based on #83 , so it should only be merged after it

**Release note**:

```release-note
None
```


omeryahud:
/hold
Until #83 is merged

PR #92, opened 2021-01-13
Add app.kubernetes.io labels to resources provisioned by ssp-operator

kwiesmueller:
This PR adds the [kubernetes app labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) to all resources maintained by the operator.

We are adding:
```
app.kubernetes.io/name: <name of the operand>
app.kubernetes.io/version: <version of the CSV (more details below)>
app.kubernetes.io/component: <component in the overall architecture this belongs to>
app.kubernetes.io/part-of: <name of the entire deployment (more details below)>
app.kubernetes.io/managed-by: "ssp-operator"
```

The values for `app.kubernetes.io/version` and `app.kubernetes.io/part-of` should be provided by HCO as they are based on the CSV.
To get those values when reconciling resources, we look at the SSP CR and extract those two labels from it if available.

The kitchen recommends to consume this PR by commits. It should give you the changes grouped by operand.

**Release note**:

```release-note
Add app.kubernetes.io/* labels to resources managed by the operator.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/operator-framework/api v0.3.25
 	github.com/operator-framework/operator-lib v0.2.0
 	github.com/spf13/cobra v1.0.0
+	gomodules.xyz/jsonpatch/v2 v2.0.1
 	gopkg.in/yaml.v2 v2.3.0
 	k8s.io/api v0.19.3
 	k8s.io/apiextensions-apiserver v0.19.3
--- a/internal/common/labels.go
+++ b/internal/common/labels.go
@@ -1,0 +1,60 @@
+package common
+
+import "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+const (
+	AppKubernetesNameLabel      = "app.kubernetes.io/name"
+	AppKubernetesPartOfLabel    = "app.kubernetes.io/part-of"
+	AppKubernetesVersionLabel   = "app.kubernetes.io/version"
+	AppKubernetesManagedByLabel = "app.kubernetes.io/managed-by"
+	AppKubernetesComponentLabel = "app.kubernetes.io/component"
+)
+
+type AppComponent string
+
+func (a AppComponent) String() string {
+	return string(a)
+}
+
+const (
+	AppComponentMonitoring AppComponent = "monitoring"
+	AppComponentSchedule   AppComponent = "schedule"
+	AppComponentTemplating AppComponent = "templating"
+)
+
+// AddAppLabels to the provided obj
+// Name will translate into the AppKubernetesNameLabel
+// Component will translate into the AppKubernetesComponentLabel
+// Instance wide labels will be taken from the request if available
+func AddAppLabels(request *Request, name string, component AppComponent, obj controllerutil.Object) controllerutil.Object {
+	labels := getOrCreateLabels(obj)
+	addInstanceLabels(request, labels)
+
+	labels[AppKubernetesNameLabel] = name
+	labels[AppKubernetesComponentLabel] = component.String()
+	labels[AppKubernetesManagedByLabel] = "ssp-operator"
+
+	return obj
+}
+
+func getOrCreateLabels(obj controllerutil.Object) map[string]string {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+		obj.SetLabels(labels)
+	}
+	return labels
+}
+
+func addInstanceLabels(request *Request, to map[string]string) {
+	if request.Instance.Labels == nil {
+		return
+	}
+
+	copyLabel(request.Instance.Labels, to, AppKubernetesPartOfLabel)
+	copyLabel(request.Instance.Labels, to, AppKubernetesVersionLabel)
+}
+
+func copyLabel(from, to map[string]string, key string) {
+	to[key] = from[key]
+}
--- a/internal/common/labels.go
+++ b/internal/common/labels.go
@@ -1,6 +1,9 @@
 package common
 
-import "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+import (
+	"kubevirt.io/ssp-operator/api/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
 
 const (
 	AppKubernetesNameLabel      = "app.kubernetes.io/name"
@@ -26,9 +29,9 @@ const (
 // Name will translate into the AppKubernetesNameLabel
 // Component will translate into the AppKubernetesComponentLabel
 // Instance wide labels will be taken from the request if available
-func AddAppLabels(request *Request, name string, component AppComponent, obj controllerutil.Object) controllerutil.Object {
+func AddAppLabels(requestInstance *v1beta1.SSP, name string, component AppComponent, obj controllerutil.Object) controllerutil.Object {
 	labels := getOrCreateLabels(obj)
-	addInstanceLabels(request, labels)
+	addInstanceLabels(requestInstance, labels)
 
 	labels[AppKubernetesNameLabel] = name
 	labels[AppKubernetesComponentLabel] = component.String()
@@ -46,13 +49,13 @@ func getOrCreateLabels(obj controllerutil.Object) map[string]string {
 	return labels
 }
 
-func addInstanceLabels(request *Request, to map[string]string) {
-	if request.Instance.Labels == nil {
+func addInstanceLabels(requestInstance *v1beta1.SSP, to map[string]string) {
+	if requestInstance.Labels == nil {
 		return
 	}
 
-	copyLabel(request.Instance.Labels, to, AppKubernetesPartOfLabel)
-	copyLabel(request.Instance.Labels, to, AppKubernetesVersionLabel)
+	copyLabel(requestInstance.Labels, to, AppKubernetesPartOfLabel)
+	copyLabel(requestInstance.Labels, to, AppKubernetesVersionLabel)
 }
 
 func copyLabel(from, to map[string]string, key string) {
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -43,8 +43,13 @@ func GetOperand() operands.Operand {
 }
 
 func (c *commonTemplates) Name() string {
-	return "CommonTemplates"
+	return operandName
 }
+
+const (
+	operandName      = "common-templates"
+	operandComponent = common.AppComponentTemplating
+)
 
 func (c *commonTemplates) AddWatchTypesToScheme(s *runtime.Scheme) error {
 	return templatev1.Install(s)
@@ -106,13 +111,15 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
 }
 
 func reconcileGoldenImagesNS(request *common.Request) (common.ResourceStatus, error) {
-	return common.CreateOrUpdateClusterResource(request, newGoldenImagesNS(GoldenImagesNSname),
+	return common.CreateOrUpdateClusterResource(request,
+		common.AddAppLabels(request, operandName, operandComponent, newGoldenImagesNS(GoldenImagesNSname)),
 		func(_, _ controllerutil.Object) {
 			// Namespace spec only contains finalizers, which can be modified
 		})
 }
 func reconcileViewRole(request *common.Request) (common.ResourceStatus, error) {
-	return common.CreateOrUpdateClusterResource(request, newViewRole(GoldenImagesNSname),
+	return common.CreateOrUpdateClusterResource(request,
+		common.AddAppLabels(request, operandName, operandComponent, newViewRole(GoldenImagesNSname)),
 		func(newRes, foundRes controllerutil.Object) {
 			foundRole := foundRes.(*rbac.Role)
 			newRole := newRes.(*rbac.Role)
@@ -121,7 +128,8 @@ func reconcileViewRole(request *common.Request) (common.ResourceStatus, error) {
 }
 
 func reconcileViewRoleBinding(request *common.Request) (common.ResourceStatus, error) {
-	return common.CreateOrUpdateClusterResource(request, newViewRoleBinding(GoldenImagesNSname),
+	return common.CreateOrUpdateClusterResource(request,
+		common.AddAppLabels(request, operandName, operandComponent, newViewRoleBinding(GoldenImagesNSname)),
 		func(newRes, foundRes controllerutil.Object) {
 			newBinding := newRes.(*rbac.RoleBinding)
 			foundBinding := foundRes.(*rbac.RoleBinding)
@@ -131,7 +139,8 @@ func reconcileViewRoleBinding(request *common.Request) (common.ResourceStatus, e
 }
 
 func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
-	return common.CreateOrUpdateClusterResource(request, newEditRole(),
+	return common.CreateOrUpdateClusterResource(request,
+		common.AddAppLabels(request, operandName, operandComponent, newEditRole()),
 		func(newRes, foundRes controllerutil.Object) {
 			newRole := newRes.(*rbac.ClusterRole)
 			foundRole := foundRes.(*rbac.ClusterRole)
@@ -171,7 +180,8 @@ func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, e
 	for i := range existingTemplates.Items {
 		template := &existingTemplates.Items[i]
 		funcs = append(funcs, func(*common.Request) (common.ResourceStatus, error) {
-			return common.CreateOrUpdateClusterResource(request, template,
+			return common.CreateOrUpdateClusterResource(request,
+				common.AddAppLabels(request, operandName, operandComponent, template),
 				func(_, foundRes controllerutil.Object) {
 					foundTemplate := foundRes.(*templatev1.Template)
 					for key := range foundTemplate.Labels {
@@ -208,7 +218,8 @@ func reconcileTemplatesFuncs(request *common.Request) []common.ReconcileFunc {
 		template := &templatesBundle[i]
 		template.ObjectMeta.Namespace = namespace
 		funcs = append(funcs, func(request *common.Request) (common.ResourceStatus, error) {
-			return common.CreateOrUpdateClusterResource(request, template,
+			return common.CreateOrUpdateClusterResource(request,
+				common.AddAppLabels(request, operandName, operandComponent, template),
 				func(newRes, foundRes controllerutil.Object) {
 					newTemplate := newRes.(*templatev1.Template)
 					foundTemplate := foundRes.(*templatev1.Template)
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -112,14 +112,14 @@ func (c *commonTemplates) Cleanup(request *common.Request) error {
 
 func reconcileGoldenImagesNS(request *common.Request) (common.ResourceStatus, error) {
 	return common.CreateOrUpdateClusterResource(request,
-		common.AddAppLabels(request, operandName, operandComponent, newGoldenImagesNS(GoldenImagesNSname)),
+		common.AddAppLabels(request.Instance, operandName, operandComponent, newGoldenImagesNS(GoldenImagesNSname)),
 		func(_, _ controllerutil.Object) {
 			// Namespace spec only contains finalizers, which can be modified
 		})
 }
 func reconcileViewRole(request *common.Request) (common.ResourceStatus, error) {
 	return common.CreateOrUpdateClusterResource(request,
-		common.AddAppLabels(request, operandName, operandComponent, newViewRole(GoldenImagesNSname)),
+		common.AddAppLabels(request.Instance, operandName, operandComponent, newViewRole(GoldenImagesNSname)),
 		func(newRes, foundRes controllerutil.Object) {
 			foundRole := foundRes.(*rbac.Role)
 			newRole := newRes.(*rbac.Role)
@@ -129,7 +129,7 @@ func reconcileViewRole(request *common.Request) (common.ResourceStatus, error) {
 
 func reconcileViewRoleBinding(request *common.Request) (common.ResourceStatus, error) {
 	return common.CreateOrUpdateClusterResource(request,
-		common.AddAppLabels(request, operandName, operandComponent, newViewRoleBinding(GoldenImagesNSname)),
+		common.AddAppLabels(request.Instance, operandName, operandComponent, newViewRoleBinding(GoldenImagesNSname)),
 		func(newRes, foundRes controllerutil.Object) {
 			newBinding := newRes.(*rbac.RoleBinding)
 			foundBinding := foundRes.(*rbac.RoleBinding)
@@ -140,7 +140,7 @@ func reconcileViewRoleBinding(request *common.Request) (common.ResourceStatus, e
 
 func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
 	return common.CreateOrUpdateClusterResource(request,
-		common.AddAppLabels(request, operandName, operandComponent, newEditRole()),
+		common.AddAppLabels(request.Instance, operandName, operandComponent, newEditRole()),
 		func(newRes, foundRes controllerutil.Object) {
 			newRole := newRes.(*rbac.ClusterRole)
 			foundRole := foundRes.(*rbac.ClusterRole)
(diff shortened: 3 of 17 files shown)


tests/commonTemplates_test.go
```diff
@@ -191,6 +205,14 @@ var _ = Describe("Common templates", func() {
 				}
 			}
 		})
+
+		table.DescribeTable("should set app labels", expectAppLabels,
+			table.Entry("edit role", &editClusterRole),
```
kwiesmueller, line 210: Do we need testids for all those cases?
/cc @omeryahud @rnetser

internal/common/labels.go
```diff
@@ -0,0 +1,60 @@
...
+type AppComponent string
+
+func (a AppComponent) String() string {
+	return string(a)
+}
+
+const (
+	AppComponentMonitoring AppComponent = "monitoring"
+	AppComponentSchedule   AppComponent = "schedule"
+	AppComponentTemplating AppComponent = "templating"
```
kwiesmueller, line 22: This is the component I came up with for templating stuff :man_shrugging: 
Feel free to suggest a different component for `template-validator` and `common-templates`
Here is the list the HCO folks defined: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/994/files#diff-65ef3f8820df1f74c24a4326ce70629156558895a1d1afff1f8f857cdbc24638R56

internal/operands/common-templates/reconcile.go
```diff
@@ -42,6 +42,11 @@ func GetOperand() operands.Operand {
 	return &commonTemplates{}
 }
 
+const (
+	operandName      = "common-templates"
```
kwiesmueller, line 46: Those are currently conflicting with https://github.com/kubevirt/ssp-operator/pull/73/files#diff-f6d56152bde2c23e8384a420483683cbfc8b9d3152bd9c104914c737b8a99cf3R45.
I think I'd prefer it as part of the interface as well but didn't want to extend it just for that.
But the naming scheme is a bit different so we should agree on something, I think this format adheres more to the general style in kubernetes naming and labels.

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/operator-framework/api v0.3.20
 	github.com/operator-framework/operator-lib v0.2.0
 	github.com/spf13/cobra v1.0.0
+	gomodules.xyz/jsonpatch/v2 v2.0.1
```
kwiesmueller, line 17: - [ ] @kwiesmueller check this import, looks wrong :thinking:

kwiesmueller: This is what kubernetes uses: https://github.com/evanphx/json-patch
We should probably too.

kwiesmueller: Hm... I looked at the library and the one k8s uses wouldn't help really.
I only use jsonpatch for creating a test patch as having Go types is nicer, but can do without if anybody has concerns with having this dep in here.
Switching would not make a lot of sense, then I'd rather just build the patches with fmt.Sprint or use own types.

akrejcir: You can keep this dependency `gomodules.xyz/jsonpatch/v2`. We already import it transitively, because `sigs.k8s.io/controller-runtime` uses it.

tests/labels_test.go
```diff
@@ -0,0 +1,122 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	fmt.Fprintf(GinkgoWriter, "sending patch: %s", string(patchBytes))
+	return client.RawPatch(types.JSONPatchType, patchBytes)
+}
+
+func waitForLabelMatch(resource controllerutil.Object, key client.ObjectKey, expectedLabels map[string]string) {
+	Eventually(func() bool {
+		err := apiClient.Get(context.TODO(), key, resource)
+		if err != nil {
```
akrejcir, line 99: I think the test can expect that the `err` is `nil`, we do it in other places. The resource exists, and if the `Get()` call returns an error, it is probably some unexpected issue.

tests/labels_test.go
```diff
@@ -0,0 +1,122 @@
...
+	patchBytes, err := json.Marshal(operations)
+	Expect(err).NotTo(HaveOccurred())
+
+	fmt.Fprintf(GinkgoWriter, "sending patch: %s", string(patchBytes))
+	return client.RawPatch(types.JSONPatchType, patchBytes)
+}
+
+func waitForLabelMatch(resource controllerutil.Object, key client.ObjectKey, expectedLabels map[string]string) {
+	Eventually(func() bool {
+		err := apiClient.Get(context.TODO(), key, resource)
```
akrejcir, line 98: You can use global `ctx` variable instead of `context.TODO()`.
```
apiClient.Get(ctx, key, resource)
```

tests/commonTemplates_test.go
```diff
@@ -19,9 +19,18 @@ import (
...
 
+	"kubevirt.io/ssp-operator/internal/common"
 	commonTemplates "kubevirt.io/ssp-operator/internal/operands/common-templates"
 )
 
+var commonTemplateExpectedLabels = map[string]string{
+	common.AppKubernetesNameLabel:      "common-templates",
+	common.AppKubernetesManagedByLabel: "ssp-operator",
+	common.AppKubernetesPartOfLabel:    "test",
+	common.AppKubernetesVersionLabel:   "v0.0.0-test",
```
akrejcir, line 30: The labels `common.AppKubernetesPartOfLabel` and `common.AppKubernetesVersionLabel` depend on the SSP resource. Could you change this variable into a function where these 2 labels are copied from the SSP resource? Or maybe, can you add these values to `TestSuiteStrategy` interface, then the function can be written like:

```
func commonTemplateExpectedLabels() map[string]string {
	return map[string]string{
		common.AppKubernetesNameLabel:      "common-templates",
		common.AppKubernetesManagedByLabel: "ssp-operator",
		common.AppKubernetesPartOfLabel:    strategy.getPartOfLabel(),
		common.AppKubernetesVersionLabel:   strategy.getVersionLabel(),
		common.AppKubernetesComponentLabel: common.AppComponentTemplating.String(),
	}
}
```

akrejcir (changes requested):

tests/labels_test.go
kwiesmueller: Okay, thanks. Didn't know we had a global context.
What is the global ctx there for?

tests/labels_test.go
kwiesmueller: I'm not sure I understand what you mean.
Right now it expects `nil` but if it gets and error it retries until it's fine again.

kwiesmueller: Oh, you mean Assert it is nil.
Yeah this seems more resilient, but I can switch it if you like.

internal/common/labels.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// AddAppLabels to the provided obj
+// Name will translate into the AppKubernetesNameLabel
+// Component will translate into the AppKubernetesComponentLabel
+// Instance wide labels will be taken from the request if available
+func AddAppLabels(request *Request, name string, component AppComponent, obj controllerutil.Object) controllerutil.Object {
```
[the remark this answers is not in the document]
kwiesmueller: Yes. Though it makes consuming this somewhat longer as `&request` has to be changed to `request.Instance` and should there be a change to Request a lot more places have to be changed.
Technically it should not make any difference as both are pointers and I'd say it's nicer to read this way. But I don't have a hard opinion on that.

internal/common/labels.go
```diff
@@ -0,0 +1,60 @@
...
+func addInstanceLabels(request *Request, to map[string]string) {
+	if request.Instance.Labels == nil {
+		return
+	}
+
+	copyLabel(request.Instance.Labels, to, AppKubernetesPartOfLabel)
+	copyLabel(request.Instance.Labels, to, AppKubernetesVersionLabel)
+}
+
+func copyLabel(from, to map[string]string, key string) {
```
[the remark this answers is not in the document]
kwiesmueller: I changed it after having it inlined first, which looks like
```
to[AppKubernetesPartOfLabel] = instanceLabels[AppKubernetesPartOfLabel]
to[AppKubernetesVersionLabel] = instanceLabels[AppKubernetesVersionLabel]
```
My reason to change it, was the duplication or rather the opportunity for mixing something up in the labels.
But yes, it's not entirely necessary.

tests/commonTemplates_test.go
kwiesmueller: I just changed that and almost fell into a trap.
When doing that, I have to explicitly set the labels on the SSP CR defined in the strategies otherwise this will not actually test anything (as the functions return "").
This means I wouldn't have to patch the SSP resource, but also means we're not testing the labels get set when SSP changes.

akrejcir: Well yes, the default test strategy creates a new SSP CR and runs tests on it. So the labels need to be added to this CR. When running the tests on existing SSP CR, I would add code to `func (s *existingSspStrategy) Init()` that checks if the labels are present, and would fail the whole test suite if they are not.

tests/labels_test.go
akrejcir: Just to make the API call shorter than using the `context.TODO()`. It has no other function. Maybe we can remove it in the future.

tests/metrics_test.go
```diff
@@ -55,4 +56,26 @@ var _ = Describe("Metrics", func() {
 			expectRestoreAfterUpdateWithPause(&prometheusRuleRes)
 		})
 	})
+
+	Context("app labels", func() {
```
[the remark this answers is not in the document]
kwiesmueller: Thanks I forgot about that one.
It doesn't really cause any unused functions though.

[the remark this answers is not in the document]
kwiesmueller: Done

tests/commonTemplates_test.go
kwiesmueller: Cool, thanks. Will do

internal/common/labels.go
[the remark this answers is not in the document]
akrejcir: I understand, but I still think it is a good practice to create minimal APIs, so functions only take parameters that they need.

If there is a change to Request, many places would need to be changed, yes, but in modern IDEs renaming a field everywhere is very simple. For example GoLand has Refactor->Rename action.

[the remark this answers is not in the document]
kwiesmueller: Done

tests/commonTemplates_test.go
kwiesmueller: Have a look now, I moved this into a function and used it everywhere. Also the labels are now set and verified as part of the strategy.
I think we can do without the assertions for every label, but for now they are there.

tests/tests_suite_test.go
```diff
@@ -140,6 +151,19 @@ func (s *newSspStrategy) GetValidatorReplicas() int {
 	return templateValidatorReplicas
 }
 
+func (s *newSspStrategy) GetVersionLabel() string {
+	if s.ssp.Labels == nil {
```
akrejcir, line 155: No need to check for `nil` here, because it is set to a non-nil value in `Init()`.

tests/tests_suite_test.go
```diff
@@ -140,6 +151,19 @@ func (s *newSspStrategy) GetValidatorReplicas() int {
...
+		return ""
+	}
+	return s.ssp.Labels[common.AppKubernetesVersionLabel]
+}
+func (s *newSspStrategy) GetPartOfLabel() string {
+	if s.ssp.Labels == nil {
```
akrejcir, line 161: Similarly here, no need to check for `nil`.

tests/tests_suite_test.go
```diff
@@ -171,6 +195,22 @@ func (s *existingSspStrategy) Init() {
...
 	}
 
+	appLabels := map[string]string{
+		common.AppKubernetesNameLabel:      "ssp-cr",
+		common.AppKubernetesManagedByLabel: "ssp-test-strategy",
+		common.AppKubernetesPartOfLabel:    "hyperconverged-cluster",
+		common.AppKubernetesVersionLabel:   "v0.0.0-test",
+		common.AppKubernetesComponentLabel: common.AppComponentSchedule.String(),
+	}
+	patch := buildLabelsPatchAdding(appLabels)
```
akrejcir, line 205: It would be better if it is not necessary to patch the existing CR. The tests that modify the SSP CR revert it back to its original state in `AfterEach()`. That revert would remove these labels.

I have a few questions about this:
- Are these labels added by HCO in the current HCO version? If yes, we don't need to patch the SSP CR.
- If the labels are not present, what do we want to happen?
  -  Should the tests ignore the missing labels? The code in `labels.go` sets them to empty string.
  -  Should the tests fail? If yes, a good place for the check is in this function, to fail the whole test suite.

akrejcir (changes requested):

tests/tests_suite_test.go
kwiesmueller: The change is merged in hco, I don't know if that means they are already in a version.
`part-of` and `version` depend on them being set on the SSP CR and if they are not set there they are "allowed" but discouraged to be empty.
Technically the other labels are mandatory as the operator specifies them and if they are missing something is wrong.
But we could also drop them if we miss `part-of` and `version` in the form of "all or nothing".

Maybe @omeryahud has an opinion on this, I don't really mind either way. I think I'd prefer how it is and if `part-of` and `version` are missing on SSP we just keep them empty but still add the other labels.

Does that answer your question?

akrejcir: Yes, thank you. In that case, this patch API call is not needed.

kwiesmueller: Great, thanks. I'll remove it.

internal/common/labels.go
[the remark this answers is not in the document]
akrejcir: Ok, you can leave it as is.

tests/labels_test.go
```diff
@@ -0,0 +1,130 @@
...
+func expectAppLabelsRestoreAfterUpdate(res *testResource) {
+	testExpectedLabelsRestoreAfterUpdate(res.NewResource(), res.GetKey(), res.ExpectedLabels)
+}
+
+func testExpectedLabels(resource controllerutil.Object, key client.ObjectKey, expectedLabels map[string]string) {
+	patchAppLabelsIntoSSP(expectedLabels)
+	waitForLabelMatch(resource, key, expectedLabels)
+}
+
+func testExpectedLabelsRestoreAfterUpdate(resource controllerutil.Object, key client.ObjectKey, expectedLabels map[string]string) {
```
akrejcir, line 41: This function is only called from `expectAppLabelsRestoreAfterUpdate()` function, so can be inlined.

tests/labels_test.go
```diff
@@ -0,0 +1,130 @@
...
+	testExpectedLabels(resource, key, expectedLabels)
```
akrejcir, line 42: I think it is not necessary to test if labels are correct before changing them. It should be enough to test if they are restored after the change.

tests/labels_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	operations := newLabelOperations(expectedLabels)
+	for label := range expectedLabels {
+		operations = append(operations, labelPatchOperationFor("replace", label, "wrong"))
+	}
+	patch := encodePatch(operations)
+	err := apiClient.Patch(ctx, resource, patch)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(labelsMatch(expectedLabels, resource)).To(BeFalse())
```
akrejcir, line 51: If I understand the k8s API correctly, after PATCH is successfully executed, the returned object will have the new labels, right? In that case it is not necessary to check them here.

tests/labels_test.go
```diff
@@ -0,0 +1,130 @@
...
+}
```
akrejcir, line 52: This function `testExpectedLabelsRestoreAfterUpdate` is not actually testing that the labels are restored after update. Please add an `Eventually()` block that will check it.

tests/labels_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+func patchAppLabelsIntoSSP(expectedLabels map[string]string) {
```
akrejcir, line 54: This function is no longer needed. We can assume the SSP CR contains the labels, as we discussed in one of the other comments.

akrejcir (changes requested):
I have a few more comments. 
By the way, this is a little inconvenient to review with so many commits changing the same code. I need to squash them before reviewing.

tests/labels_test.go
kwiesmueller: Right, that was supposed to be waitForLabelsMatch must have gone lost in the refactor as it was tested before. Thanks for the catch!

tests/labels_test.go
kwiesmueller: This is/was there as it did the patch to SSP.
Not needed anymore in theory, but ensures it really tests a full restore.

kwiesmueller:
@akrejcir I'm just trying to debug this failed test.
I saw that locally as well (twice) but not every time. I think it's because of the removed patchAppLabelsIntoSSP.
I kind of expected this, as the function ensured they are really there as a prerequisite to the actual test.

Might there be something wrong with the init stuff?

internal/operands/common-templates/reconcile.go
```diff
@@ -171,7 +180,8 @@ func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, e
 	for i := range existingTemplates.Items {
 		template := &existingTemplates.Items[i]
 		funcs = append(funcs, func(*common.Request) (common.ResourceStatus, error) {
-			return common.CreateOrUpdateClusterResource(request, template,
+			return common.CreateOrUpdateClusterResource(request,
```
kwiesmueller, line 183: @akrejcir what would you think about adding
`CreateOrUpdateClusterResourceWithAppLabels(request *Request, name string, component AppComponent, resource controllerutil.Object, updateResource ResourceUpdateFunc)`
which would move the need for those AddAppLabels calls into a wrapped function call?
I added it locally and it looks good so far.

kwiesmueller: So it would become
```suggestion
			return common.CreateOrUpdateClusterResourceWithAppLabels(request, operandName, operandComponent, template,
```

akrejcir:
> @akrejcir I'm just trying to debug this failed test.
> I saw that locally as well (twice) but not every time. I think it's because of the removed patchAppLabelsIntoSSP.
> I kind of expected this, as the function ensured they are really there as a prerequisite to the actual test.
> 
> Might there be something wrong with the init stuff?

I'm not sure, I will need to debug it too.

internal/operands/common-templates/reconcile.go
akrejcir: I think we need to create a builder for these calls. Then each resource can use what it needs. For example:
```
return common.CreateOrUpdate(request, resource).
    ClusterResource().
    WithAppLabels(name, component).
    UpdateFunc(...).
    Execute()
```

Let's leave it for another PR.

tests/labels_test.go
akrejcir: After debugging, I found out why the tests fail when this function is removed.
The reason is that the `createOrUpdateSsp()` function in `tests_suite_test.go` does not recreate the SSP with labels.
Can you change that function to the following?
```
func createOrUpdateSsp(ssp *sspv1beta1.SSP) {
	key := client.ObjectKey{
		Name:      ssp.Name,
		Namespace: ssp.Namespace,
	}
	Eventually(func() error {
		foundSsp := &sspv1beta1.SSP{}
		err := apiClient.Get(ctx, key, foundSsp)
		if err == nil {
			isSspEqual := reflect.DeepEqual(foundSsp.Spec, ssp.Spec) &&
				reflect.DeepEqual(foundSsp.Annotations, ssp.Annotations) &&
				reflect.DeepEqual(foundSsp.Labels, ssp.Labels)
			if isSspEqual {
				return nil
			}
			foundSsp.Spec = ssp.Spec
			foundSsp.Annotations = ssp.Annotations
			foundSsp.Labels = ssp.Labels
			return apiClient.Update(ctx, foundSsp)
		}
		if errors.IsNotFound(err) {
			newSsp := &sspv1beta1.SSP{
					Name:        ssp.Name,
					Namespace:   ssp.Namespace,
					Annotations: ssp.Annotations,
					Labels:      ssp.Labels,
				},
				Spec: ssp.Spec,
			}
		}
		return err
}
```

And then, you can remove this one `patchAppLabelsIntoSSP`.

kwiesmueller: Thanks for spotting this, I updated the PR.

akrejcir:
There may be another problem with the upgrade tests, I'm looking into it.

akrejcir:
The upgrade test fails with:
```
2021-02-01T18:49:56.517Z	DEBUG	controllers.SSP	Operand reconciliation failed: DaemonSet.apps "kubevirt-node-labeller" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"kubevirt-node-labeller", "app.kubernetes.io/component":"schedule", "app.kubernetes.io/managed-by":"ssp-operator", "app.kubernetes.io/name":"node-labeler"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable	{"ssp": "ssp-operator-functests/ssp-test"}
2021-02-01T18:49:56.529Z	ERROR	controller	Reconciler error	{"reconcilerGroup": "ssp.kubevirt.io", "reconcilerKind": "SSP", "controller": "ssp", "name": "ssp-test", "namespace": "ssp-operator-functests", "error": "DaemonSet.apps \"kubevirt-node-labeller\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\"app\":\"kubevirt-node-labeller\", \"app.kubernetes.io/component\":\"schedule\", \"app.kubernetes.io/managed-by\":\"ssp-operator\", \"app.kubernetes.io/name\":\"node-labeler\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable"}
```

[Logs from ssp-operator pod in the failed job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/92/pull-ci-kubevirt-ssp-operator-master-e2e-upgrade-functests/1356269791043653632/artifacts/e2e-upgrade-functests/gather-extra/artifacts/pods/kubevirt_ssp-operator-ff8d9c66d-ml9nd_manager.log)

kwiesmueller:
Ah, yeah...
It's probably because the labelSelector for the DaemonSet is immutable and the operator tries to reconcile it with the new labels added. So it probably takes the labels from metadata and also adds them to the label selector.
That's kind of a problem :-/

kwiesmueller:
This means we either have to recreate the DS in this case or not copy labels from .metadata.labels.
In general I'd suspect that changes like this will fail the upgrade regularly and we'd need a strategy to handle this.

akrejcir:
The issue is that maps in go are passed by reference. When the daemon set is constructed in `node-labeller/reconcile.go` , the same instance of map is used for `metadata.labels` and `selector.matchLabels`.

Can you modify the function `AddAppLabels` so that it creates a new map instance, fills it with necessary labels and assigns it to `metadata.labels` ? I will check if there are any other cases where we unintentionally modify data.

kwiesmueller:
This would solve this specific problem, but in general I think the labels should be on the DaemonSet spec as well as those pass on to the pod and that's what we really care about in the end.

akrejcir:
The labels could be on the pod, but do they also need to be in the selector?

omeryahud:
We should update the `spec.selector` field as well, and it seems like we must delete the `DaemonSet` and recreate it, in the specific case of `node-labeller`, I think re-creating it should not pose an issue during an upgrade, because it will simply re-trigger the labeling process, without removing any existing labels.
Can you think of any reason not to re-create it?

kwiesmueller:
For node-labeler it should be okay.
Should we add a logic that tries to update and if that fails recreate?

I'm more concerned with this for actual workloads. Not sure how kubevirt solved that for VMIs and pods.

omeryahud:
We can try to directly label the pods of the DS update fails with this error (only happens of upgrades), and for fresh deployments, the DS will be properly created with an updates `spec.selector`.
WRT workloads, a VMI owns a pod using owner references, there is no other managing object for the pod, so I don't think there was an issue there

akrejcir:
The reconciliation method could check if any of the selector labels need to be changed, and if yes, it will remove the DS. Next it will call the usual `createOrUpdate` method to recreate it.

Soon the node labeller will be part of kubevirt, so this code is only temporary.

tests/labels_test.go
akrejcir: Please also update the `reflect.DeepEqual()` check on line 419 in `tests_suite_test.go`. See my previous comment, the deep equals need to compare labels and annotations, not only spec.

kwiesmueller: Ja, me blind... Thanks again.

kwiesmueller:
I suspect this fails on the upgrade test because the labels are not on the old version? Or am I still missing something

akrejcir:
> I suspect this fails on the upgrade test because the labels are not on the old version? Or am I still missing something

Probably, yes, but I think the tests should not fail in case the labels on SSP CR are missing. What do you think?
The log message says: 
```
expected label app.kubernetes.io/part-of=, got: 
```
It seems correct, that it expects value `""` and found `""`. Can you debug this?
Also there are a lot of these log messages, because the log is written in a loop in `Eventually()`, can you make the log less verbose?

kwiesmueller:
Will do

kwiesmueller:
I don't know if I agree on the test part.
If we don't fail if the label is missing from the CR we don't always test if they get propagated correctly and if we want to do that, it might make the tests more complicated.
I'm still trying to run it locally, but it might be enough to add the labels here in automation/ssp-cr-template.yaml

akrejcir:
Good idea to add labels to `automation/ssp-cr-template.yaml`.

/lgtm

PR #97, opened 2021-01-19
Generate the CSV using csv-generator

jean-edouard:
This PR switches to using csv-generator to generate the ClusterServiceVersion yaml.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ jobs:
     if: tag IS present
     script: |
       docker login -u="${QUAY_BOT}" -p="${QUAY_PASSWORD}" quay.io
-      make release IMG_TAG="${TRAVIS_TAG}"
+      make release IMG_TAG="${TRAVIS_TAG}" VERSION="`echo ${TRAVIS_TAG} | sed 's/^v//'`"
     deploy:
       provider: releases
       name: $TRAVIS_TAG
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,10 @@ functest: generate fmt vet manifests
 manager: generate fmt vet
 	go build -o bin/manager main.go
 
+# Build csv-generator binary
+csv-generator: generate fmt vet
+	go build -o csv-generator hack/csv-generator.go
+
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests
 	go run ./main.go
@@ -142,20 +146,16 @@ else
 KUSTOMIZE=$(shell which kustomize)
 endif
 
-.PHONY: yq
-yq:
-	test -f yq || wget https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64 -O yq
-	chmod +x yq
-
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: operator-sdk manifests yq
+bundle: operator-sdk manifests csv-generator
 	./operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | ./operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
-	./yq w -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.webhookdefinitions[0].containerPort 9443
-	./yq d -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.install.spec.deployments[0].spec.template.spec.volumes
-	./yq d -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.install.spec.deployments[0].spec.template.spec.containers[0].volumeMounts
+	./csv-generator --csv-version $(VERSION) --namespace kubevirt --operator-image $(IMG) --operator-version $(VERSION) \
+			--file bundle/manifests/ssp-operator.clusterserviceversion.yaml \
+			--webhook-port 9443 --webhook-remove-certs > bundle/manifests/ssp-operator.clusterserviceversion.yaml.new
+	mv bundle/manifests/ssp-operator.clusterserviceversion.yaml.new bundle/manifests/ssp-operator.clusterserviceversion.yaml
 	./operator-sdk bundle validate ./bundle
 	rm -rf _out
 	mkdir -p _out
--- a/hack/csv-generator.go
+++ b/hack/csv-generator.go
@@ -39,7 +39,10 @@ import (
 )
 
 type generatorFlags struct {
+	file              string
 	dumpCRDs          bool
+	removeCerts       bool
+	webhookPort       int32
 	csvVersion        string
 	namespace         string
 	operatorVersion   string
@@ -72,6 +75,7 @@ func main() {
 }
 
 func init() {
+	rootCmd.Flags().StringVar(&f.file, "file", "data/olm-catalog/ssp-operator.clusterserviceversion.yaml", "Location of the CSV yaml to modify")
 	rootCmd.Flags().StringVar(&f.csvVersion, "csv-version", "", "Version of csv manifest (required)")
 	rootCmd.Flags().StringVar(&f.namespace, "namespace", "", "Namespace in which ssp operator will be deployed (required)")
 	rootCmd.Flags().StringVar(&f.operatorImage, "operator-image", "", "Link to operator image (required)")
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -145,13 +149,11 @@ yq:
 
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: operator-sdk manifests yq
+bundle: operator-sdk manifests yq csv-generator
 	./operator-sdk generate kustomize manifests -q
+	./csv-generator --csv-version $(VERSION) --operator-image $(IMG) --operator-version $(VERSION) > bundle/manifests/ssp-operator.clusterserviceversion.yaml
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | ./operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
-	./yq w -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.webhookdefinitions[0].containerPort 9443
```
omeryahud, line 152: The functionality of these lines should be added to the csv-generator, requried by OLM

hack/csv-generator.go
```diff
@@ -83,7 +83,6 @@ func init() {
 	rootCmd.Flags().BoolVar(&f.dumpCRDs, "dump-crds", false, "Dump crds to stdout")
 
 	rootCmd.MarkFlagRequired("csv-version")
-	rootCmd.MarkFlagRequired("namespace")
```
omeryahud, line 86: I would leave this field as required, and pass the `kubevirt` namespace to the csv-generator

omeryahud:
An idea for the csv-generator usage is 
`csv-generator <existing-flags> --csv-template=<path-to-csv>`
`path-to-csv` points to the CSV generated by `./operator-sdk generate bundle` in `make bundle`
This way, csv-generator can pick up the csv template from within the container downstream, and we can override the csv template upstream with `--csv-template`

Makefile
jean-edouard: That seems to be there already, the CSV yaml coming out of csv-generator is virtually identical to the one from the latest release. (though the csv-generator code doesn't grep for "9443"...)

hack/csv-generator.go
jean-edouard: That namespace field doesn't seem to be used anywhere, the value passed to it doesn't end up in the final yaml, and csv-generator runs happily without it.
Maybe it's there for future use? In any case, it doesn't feel like it should be "required" just yet.

akrejcir:
@omeryahud , I'm not sure why is this change needed or useful. Can you please explain it? IIUC, `csv-generator` is a downstream "hack" to change some values in csv to correspond to downstream image paths and versions.
Is the upstream CSV generated by operator-sdk not good?

omeryahud:
> @omeryahud , I'm not sure why is this change needed or useful. Can you please explain it? IIUC, `csv-generator` is a downstream "hack" to change some values in csv to correspond to downstream image paths and versions.
> Is the upstream CSV generated by operator-sdk not good?

Yes, I probably should have elaborated on it.
The CSV generated by operator-sdk includes the certificate mounts that we use for non-OLM deployments.
OLM expects to manage the cert volumes itself, so we should remove ours to prevent collision.
And yes, csv-generator is sort of a 'hack' downstream, but it is useful upstream also if we don't have another automation (which we don't)

jean-edouard:
The necessary options were added to csv-generator so that the Makefile `bundle` rule can use it instead of `yq` to adjust yaml values.
It was verified that `make bundle` with this PR produced the exact same CSV yaml as `make bundle` in master.

Makefile
```diff
@@ -142,20 +146,14 @@ else
...
 .PHONY: bundle
-bundle: operator-sdk manifests yq
+bundle: operator-sdk manifests csv-generator
 	./operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | ./operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
-	./yq w -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.webhookdefinitions[0].containerPort 9443
-	./yq d -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.install.spec.deployments[0].spec.template.spec.volumes
-	./yq d -i bundle/manifests/ssp-operator.clusterserviceversion.yaml spec.install.spec.deployments[0].spec.template.spec.containers[0].volumeMounts
+	./csv-generator --csv-version $(VERSION) --namespace kubevirt --operator-image $(IMG) --file bundle/manifests/ssp-operator.clusterserviceversion.yaml --webhook-port 9443 --webhook-remove-certs > bundle/manifests/ssp-operator.clusterserviceversion.yaml.new
```
omeryahud, line 155: I know that the `VERSION` var exist here, but we actually use `IMG_TAG` in the travis job (where it counts in this context), so lets use `IMG_TAG` here as well and decide if we want to use `VERSION` or remove it in a different PR

omeryahud: Also, `--operator-version` is missing here, so the `OPERATOR_VERSION` env var comes out empty in the deployment defined in the CSV

omeryahud (commented):
Looks good, just some small stuff I noticed, let's link the operator image in this annotation as well:
```
metadata:
  annotations:
    containerImage: REPLACE_IMAGE:TAG
```

Makefile
jean-edouard: Thanks, I'll replace `VERSION` with `IMG_TAG`.

As for OPERATOR_VERSION, it comes out empty without this PR as well. I went for 0 diff between `make bundle` before and after this PR.
I was planning on fixing that along with other things as part of a separate effort, but I can do it here if preferred.

omeryahud: Yes, I would prefer to take care of `OPERATOR_VERSION` as well, as it is a very low hanging fruit in this context

jean-edouard: While adding OPERATOR_VERSION, I noticed I forgot to replace VERSION with IMG_TAG for `--csv-version` above.
However, when I do that, csv-generator fails because IMG_TAG defaults to "latest" and csv-generator expects X.Y.Z.
Should I keep using VERSION, or fix csv-generator, or change the default value?
Thanks

omeryahud: I think it would be safe to do the following changes:
```
IMG_TAG ?= v$(VERSION)
```
This way, there's no danger of overriding the `v0.0.1` image because we have no releases prior to `v0.1.0` WDYT?

jean-edouard: Switched to setting VERSION in travis-ci as discussed offline, force-pushed, let me know if that makes sense.

.travis.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     if: tag IS present
     script: |
       docker login -u="${QUAY_BOT}" -p="${QUAY_PASSWORD}" quay.io
-      make release IMG_TAG="${TRAVIS_TAG}"
+      make release IMG_TAG="${TRAVIS_TAG}" VERSION="${TRAVIS_TAG}"
```
omeryahud, line 18: You should trim the leading `v` from `TRAVIS_TAG`, because our release tags are `vx.y.z`

jean-edouard: Sorry, show be fixed now.
Note: I used `sed` instead of the bashism `${TRAVIS_TAG:1}`, not sure if travis-ci still runs `dash` instead of `bash` but just in case...

hack/csv-generator.go
```diff
@@ -177,10 +184,35 @@ func replaceVariables(flags generatorFlags, csv *csvv1.ClusterServiceVersion) er
 				updatedVariables = append(updatedVariables, envVariable)
 			}
 			updatedContainer.Env = updatedVariables
-			csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Containers[i] = updatedContainer
+			templateSpec.Containers[i] = updatedContainer
+			if flags.removeCerts {
```
kwiesmueller, line 188: Could you move this functionality into another function to reduce complexity?

hack/csv-generator.go
```diff
@@ -177,10 +184,35 @@ func replaceVariables(flags generatorFlags, csv *csvv1.ClusterServiceVersion) er
...
+					}
+				}
+				templateSpec.Containers[i].VolumeMounts = updatedVolumeMounts
+			}
 			break
 		}
 	}
+
+	if flags.removeCerts {
+		updatedVolumes := templateSpec.Volumes
```
kwiesmueller, line 203: Same here, would it be possible to move this into a separate function?

kwiesmueller (changes requested):

hack/csv-generator.go
jean-edouard: Done, pushed in a separate commit

kwiesmueller: Thanks!
/lgtm

PR #98, opened 2021-01-20
ssp operator deletes the node-labeller

ksimon1:

**Do not merge. Work in progress. It is missing upgrade test**
ssp operator deletes the node-labeller

This PR updates the reconcile function for node-labeller,
so now it instead of creating/updating deletes all components of labeller.
**Release note**:
```
SSP operator now deletes all node-labeller components, because labeller was moved into Kubevirt.
```

### DIFF
--- a/tests/nodeLabeller_test.go
+++ b/tests/nodeLabeller_test.go
@@ -1,22 +1,18 @@
 package tests
 
 import (
-	"fmt"
 	"reflect"
-	"strings"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	secv1 "github.com/openshift/api/security/v1"
-	yaml "gopkg.in/yaml.v2"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
-	cpunfd "kubevirt.io/cpu-nfd-plugin/pkg/config"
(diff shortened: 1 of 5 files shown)


shwetaap:
See some sonar cloud failures, otherwise it 
/lgtm

tests/nodeLabeller_test.go
```diff
@@ -1,222 +0,0 @@
-package tests
```
akrejcir, line 1: Instead of removing this file, can you modify it to test that the `node-labeller` resources do not exist? It will be important for upgrade test lane.

akrejcir (changes requested):
Looks good. It just needs functional tests.

tests/nodeLabeller_test.go
```diff
@@ -103,120 +98,25 @@ var _ = Describe("Node Labeller", func() {
...
-		},
-			table.Entry("[test_id:5193] cluster role", &clusterRoleRes),
-			table.Entry("[test_id:5196] cluster role binding", &clusterRoleBindingRes),
-			table.Entry("[test_id:5202] security context constraint", &securityContextConstraintRes),
-		)
-
-		table.DescribeTable("created namespaced resource", func(res *testResource) {
-			err := apiClient.Get(ctx, res.GetKey(), res.NewResource())
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).To(HaveOccurred())
```
akrejcir, line 105: Check that the error is: `errors.IsNotFound(err)`.

PR #104, opened 2021-01-26
Add readiness probe to the operator pod

akrejcir:
Added readiness probe to the operator pod to signal when the webhook is ready to receive requests.

**Release note**:
```release-note
None
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -26,8 +26,7 @@ spec:
       containers:
       - command:
         - /manager
-        args:
-        - --enable-leader-election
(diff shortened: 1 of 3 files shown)


ksimon1 (commented):
Is there any other way how to test if the webhook is available? Because 5s might be not enough.

akrejcir:
> Is there any other way how to test if the webhook is available? Because 5s might be not enough.

I can add readiness probe to the pod that will check if the webhook is ready. That will be a better solution.

/hold

akrejcir:
Added readiness probe instead of `sleep`.

/hold cancel

PR #105, opened 2021-01-29
replaced strings with constants

ksimon1:
replaced strings with constants
This change should decrease "code smell" in sonarcloud

**Release note**:
```
NONE
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -339,6 +339,8 @@ func preUpdateStatus(request *common.Request) error {
 	sspStatus.OperatorVersion = operatorVersion
 	sspStatus.TargetVersion = operatorVersion
 
+	reconcileSSPResourceMessage := "Reconciling SSP resources"
+
 	if sspStatus.Paused {
 		request.Logger.Info(fmt.Sprintf("Unpausing SSP operator on resource: %v/%v",
 			request.Instance.Namespace, request.Instance.Name))
@@ -350,7 +352,7 @@ func preUpdateStatus(request *common.Request) error {
 			Type:    conditionsv1.ConditionAvailable,
 			Status:  v1.ConditionFalse,
 			Reason:  "available",
-			Message: "Reconciling SSP resources",
+			Message: reconcileSSPResourceMessage,
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


internal/common/resource_test.go
```diff
@@ -24,8 +24,9 @@ import (
 var log = logf.Log.WithName("common_operand_package")
 
 const (
-	namespace = "kubevirt"
-	name      = "test-ssp"
+	namespace   = "kubevirt"
+	name        = "test-ssp"
+	changedName = "changed-name"
```
akrejcir, line 29: The same here. The point of the test is to change the name, not that the changed name is the same in all tests.

akrejcir (commented):
I think these are false positives. But this issue is not that important, so if others think it is ok, feel free to give lgtm.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -30,8 +31,15 @@ var (
 )
 
 const (
-	namespace = "kubevirt"
-	name      = "test-ssp"
+	namespace             = "kubevirt"
+	name                  = "test-ssp"
+	templateVersionLabel  = "template.kubevirt.io/version"
```
omeryahud, line 36: These labels should be defined in the `common-templates` package code instead of the tests

internal/common/resource_test.go
kwiesmueller: In a case like this, where the actual content does not matter the way to go seems to be a nameGenerator.
Same as with the IP addresses, generate an arbitrary name for testing.

controllers/ssp_controller.go
```diff
@@ -337,6 +337,8 @@ func preUpdateStatus(request *common.Request) error {
 	sspStatus.OperatorVersion = operatorVersion
 	sspStatus.TargetVersion = operatorVersion
 
+	reconcileSSPResourceMessage := "Reconciling SSP resources"
```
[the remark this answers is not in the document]
kwiesmueller: I don't fully agree. Strings like this can still be (or are) part of the public API and can be something people expect to be the same.
It's not only about code semantics, but consistency and possibility of human error.

Right now if somebody changes the wording, they have to find all occurrences of the string and also hope there is no typo in any of them. Having this cleaned up into either a constant like here or having a function for creating the condition reduces those risks and makes it easier for the future.

ksimon1:
Closing in favor of https://github.com/kubevirt/ssp-operator/pull/124/files

PR #108, opened 2021-02-01
Temporarily add sleep to upgrade test

akrejcir:
The master version of operator uses readiness probe, but the upgrade test deploys the latest released version, which does not yet have the probe.

**Release note**:
```release-note
None
```

### DIFF
--- a/automation/ocp-ci-run-e2e-upgrade-tests.sh
+++ b/automation/ocp-ci-run-e2e-upgrade-tests.sh
@@ -13,6 +13,11 @@ oc apply -n $NAMESPACE -f "https://github.com/kubevirt/ssp-operator/releases/dow
 # Wait for deployment to be available, otherwise the validating webhook would reject the SSP CR.
 oc wait --for=condition=Available --timeout=600s -n ${NAMESPACE} deployments/ssp-operator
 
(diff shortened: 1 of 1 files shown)


akrejcir:
This PR should temporarily fix the failing upgrade test.

/cc @omeryahud @ksimon1 @kwiesmueller

omeryahud:
Because it is only temporary, I would go with something like 1 minute, just because the OCP CI infra is unpredictable, WDYT?

akrejcir:
> Because it is only temporary, I would go with something like 1 minute, just because the OCP CI infra is unpredictable, WDYT?

It seems like too much. The line before the sleep waits for the deployment to be available. At the time the sleep is run, a pod with the operator is running, but the application inside may not yet be ready to receive webhook requests.

omeryahud:
> > Because it is only temporary, I would go with something like 1 minute, just because the OCP CI infra is unpredictable, WDYT?
> 
> It seems like too much. The line before the sleep waits for the deployment to be available. At the time the sleep is run, a pod with the operator is running, but the application inside may not yet be ready to receive webhook requests.

OK!
/lgtm
/approve

PR #109, opened 2021-02-02
csv-generator: populate spec.relatedImages

jean-edouard:
csv-generator now computes and populates proper `relatedImages` for the CSV

```release-note
The relatedImages spec field is now populated and image caching should be supported
```

### DIFF
--- a/hack/csv-generator.go
+++ b/hack/csv-generator.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/operands/node-labeller"
 )
 
 type generatorFlags struct {
@@ -117,7 +118,12 @@ func runGenerator() error {
 		removeCerts(f, &csv)
 	}
 
-	err = marshallObject(csv, os.Stdout)
+	relatedImages, err := buildRelatedImages(f)
+	if err != nil {
+		return err
+	}
+
+	err = marshallObject(csv, relatedImages, os.Stdout)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


hack/csv-generator.go
```diff
@@ -259,6 +337,8 @@ func marshallObject(obj interface{}, writer io.Writer) error {
 		unstructured.SetNestedSlice(r.Object, deployments, "spec", "install", "spec", "deployments")
 	}
 
+	unstructured.SetNestedSlice(r.Object, relatedImages, "spec", "relatedImages")
```
akrejcir, line 340: Is the `spec.relatedImages` part of the CSV? I cannot find it in the API definition:
https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go#L268

omeryahud: @nunnatsa Nahshon, when we spoke about this,  I remember `relatedImages` was supposed to be a part of the CSV spec, if not, where should this data be present?

nunnatsa: Thsis in not implemented in OLM and so this is not part of the API. Please see the HCO implementation here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tools/csv-merger/csv-merger.go#L55..L66

tiraboschi: OLM doesn't consume it, but `oc adm catalog mirror` does:
https://docs.openshift.com/container-platform/4.6/operators/admin/olm-restricted-networks.html#olm-mirror-catalog_olm-restricted-networks

akrejcir: Ok, then the code is good.
Do the `relatedImages` need to specify digest, or can they also use tag?

nunnatsa: > Ok, then the code is good.
> Do the `relatedImages` need to specify digest, or can they also use tag?

In HCO, We assumes digests. Please see the code of our `digester` here for reference: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tools/digester/main.go

hack/csv-generator.go
```diff
@@ -27,6 +27,8 @@ import (
 
 	"github.com/blang/semver"
 	gyaml "github.com/ghodss/yaml"
+	"github.com/google/go-containerregistry/pkg/name"
```
omeryahud, line 30: Does this package need a remote registry to be available? If so, I don't think it would be a good idea to rely on that, maybe instead, we can change our `defaults.go` files to use digests instead of tags

jean-edouard: Not the package per say, but yes, it is used to query the repos listed in `defaults.go` to translate their tag into digests.
I can definitely just replace the tags with digests in `defaults.go` if desired, that would simplify this PR by a lot!

omeryahud: Yes, I think it would be a better idea

akrejcir:
/lgtm

Can you remove unneeded text from the PR's description?

PR #112, opened 2021-02-10
Migrate TravisCI jobs to Github actions

omeryahud:
Migrate TravisCI jobs to Github actions because of TravisCI's rate limitations.

On Pull Request:
- build operator container
- test csv-generator

On Release:
- Build operator container
- Push operator container
- Generate and upload manifests as release assets

Release action tested on my fork: https://github.com/omeryahud/ssp-operator/runs/1870283941?check_suite_focus=true
Pull Request action tested on my fork: https://github.com/omeryahud/ssp-operator/runs/1870412857

**Release note**:
```release-note
None
```


.github/workflows/release.yaml
```diff
@@ -0,0 +1,59 @@
+on:
+  release:
+    types: [published]
```
akrejcir, line 3: Does this mean that this action is only triggered when a new github release is manually created? If yes, is there some way how to make it more like the travis flow, where a release was created automatically once a tagged commit was merged to master.

omeryahud: That's a good question, I will look into it, I haven't considered releases being created from a tagged commit

omeryahud: @akrejcir I think we should leave it as-is for now, until we have an automatic release automation, otherwise it makes no sense

akrejcir: Ok. Then we will need to always remember to create a tag before release.

omeryahud: No need, a tag is created automatically when you create a manual release on Github

akrejcir: I didn't know that. Then there is no problem.

PR #121, opened 2021-02-18
Remove global variables from resource objects

akrejcir:
Created resources could be modified during reconciliation, which would modify global instances of objects.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/internal/operands/node-labeller/resources.go
+++ b/internal/operands/node-labeller/resources.go
@@ -33,28 +33,6 @@ type nodeLabellerImages struct {
 	virtLauncher string
 }
 
-var commonLabels = map[string]string{
-	"app": "kubevirt-node-labeller",
-}
-
-var cpuPluginConfigmap = `obsoleteCPUs:
-  - "486"
-  - "pentium"
-  - "pentium2"
-  - "pentium3"
-  - "pentiumpro"
-  - "coreduo"
-  - "n270"
-  - "core2duo"
-  - "Conroe"
-  - "athlon"
-  - "phenom"
-minCPU: "Penryn"`
-
-var configMapData = map[string]string{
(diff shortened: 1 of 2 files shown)


internal/operands/template-validator/resources.go
```diff
@@ -60,7 +56,9 @@ func newServiceAccount(namespace string) *core.ServiceAccount {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ServiceAccountName,
 			Namespace: namespace,
-			Labels:    commonLabels,
```
kwiesmueller, line 63: The changes in this file should cause more duplication than before.
What's the reason/goal of this change? Reducing global vars?

internal/operands/node-labeller/resources.go
```diff
@@ -106,12 +84,28 @@ func newClusterRoleBinding(namespace string) *rbac.ClusterRoleBinding {
 }
 
 func newConfigMap(namespace string) *core.ConfigMap {
+	var cpuPluginConfigmap = `obsoleteCPUs:
```
kwiesmueller, line 87: This could be a const I think.
Then it wouldn't even need to be in there but could remain global.

internal/operands/template-validator/resources.go
akrejcir: Yes, the goal is to remove global variables, because they are mutable and shared between created resources.
Do you remember the problem caused by global variable in your PR: #92 ?

kwiesmueller: Right, I like that idea not a fan of global vars.
Just would also like to try and avoid duplication like this.

kwiesmueller:
Thanks!
/lgtm

PR #123, opened 2021-02-19
Create a builder for reconcile methods

akrejcir:
There are four `CreateAndUpdate` methods with various parameters. This patch creates a builder to use instead of these functions.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/internal/common/resource.go
+++ b/internal/common/resource.go
@@ -36,24 +36,87 @@ func CollectResourceStatus(request *Request, funcs ...ReconcileFunc) ([]Resource
 type ResourceUpdateFunc = func(expected, found controllerutil.Object)
 type ResourceStatusFunc = func(resource controllerutil.Object) ResourceStatus
 
-func CreateOrUpdateResource(request *Request, resource controllerutil.Object, updateResource ResourceUpdateFunc) (ResourceStatus, error) {
-	return createOrUpdate(request, resource, false, updateResource, statusOk)
+type ReconcileBuilder interface {
+	NamespacedResource(controllerutil.Object) ReconcileBuilder
+	ClusterResource(controllerutil.Object) ReconcileBuilder
+	WithAppLabels(name string, component AppComponent) ReconcileBuilder
+	UpdateFunc(ResourceUpdateFunc) ReconcileBuilder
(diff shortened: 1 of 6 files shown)


internal/common/resource.go
```diff
@@ -36,24 +36,87 @@ func CollectResourceStatus(request *Request, funcs ...ReconcileFunc) ([]Resource
...
+	operandName      string
+	operandComponent AppComponent
+
+	updateFunc ResourceUpdateFunc
+	statusFunc ResourceStatusFunc
+}
+
+var _ ReconcileBuilder = &reconcileBuilder{}
+
+func (r *reconcileBuilder) NamespacedResource(resource controllerutil.Object) ReconcileBuilder {
```
kwiesmueller, line 64: I would split that into:
`Resource(resource controllerutil.Object)...`
and
`Namespaced()...` or `Namespaced(bool)` instead.

internal/operands/template-validator/reconcile.go
```diff
@@ -89,55 +89,60 @@ const (
...
-	return common.CreateOrUpdateClusterResource(request,
-		common.AddAppLabels(request.Instance, operandName, operandComponent, newClusterRole()),
-		func(newRes, foundRes controllerutil.Object) {
+	return common.CreateOrUpdate(request).
+		ClusterResource(newClusterRole()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(newRes, foundRes controllerutil.Object) {
 			foundRes.(*rbac.ClusterRole).Rules = newRes.(*rbac.ClusterRole).Rules
-		})
+		}).Reconcile()
```
kwiesmueller, line 97: Maybe make the formatting consistent, right now sometimes the reconcile is on this line, sometimes on its own one.
I don't really have a preference to either one, but just so it's consistent.

kwiesmueller (commented):
Really cool, thanks for doing this!

internal/common/resource.go
akrejcir: I did not split it, because this way it is not possible to forget to specify if a resource is cluster or namespaced.
I think in a future patch we may remove the distinction between cluster and namespaced resources and use owner annotations for all of them to simplify the code..

PR #125, opened 2021-02-25
Add remark on kustomize version

borod108:
Currently the dnf installs v2 of kustomize by default which fails the build.
Added a note about installing a newer version.

Signed-off-by: borod108 <bodnopoz@redhat.com>

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,7 +148,7 @@ endif
 
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: operator-sdk manifests csv-generator
+bundle: kustomize operator-sdk manifests csv-generator
 	./operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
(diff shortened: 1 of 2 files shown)


akrejcir:
The makefile installs the right version of kustomize during build, but only if it is not yet installed on the system. Can you mention it in the readme too?

Maybe later we can update the makefile to download kustomize to the repository folder, and not use the system version.

borod108:
@akrejcir ok took me some time to understand :P so turns out if you just did container-build it did not install kustomize :)
so added that and edited the text, what do you think now?

Makefile
```diff
@@ -103,7 +103,7 @@ operator-sdk:
 	chmod 0755 operator-sdk
 
 # Build the container image
-container-build: unittest bundle
+container-build: kustomize unittest bundle
```
akrejcir, line 106: So I see we forgot to call the `kustomize` makefile target. Please move it inside the `bundle` target instead of `container-build`.

borod108: sure, done.

akrejcir:
Looks good. Last thing, you need to add release note to the PR's description, otherwise the kubevirt-bot will not merge this. It is enough to write `None` to release note. For example: https://github.com/kubevirt/ssp-operator/pull/123

/lgtm

PR #131, opened 2021-03-09
Update common-templates to v0.14.0-rc.1

omeryahud:
Update common-templates bundle to v0.14.0-rc.1 
**Release note**:
```release-note
Update common-templates bundle to v0.14.0-rc.1
```


omeryahud:
Found another issue with the release script, version label is rendered as `0.14.0-rc.1` instead of `v0.14.0-rc.1`.
I will introduce a fix and let the automation create a new PR
/close

PR #134, opened 2021-03-15
Add Template-validator code migrated from the old template-validaor repo

vatsalparekh:
Add Template-validator code migrated from the old template-validaor repo

Most of the code is copied from the old repository with minimum changes
If there's a larger design or functional change to the template-validator, I would like to do that in a separate PR

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -4,6 +4,8 @@ VERSION ?= 0.1.3
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
 #operator-sdk version
 OPERATOR_SDK_VERSION ?= v1.4.2
+#Options for Template-Validator
+VALIDATOR_VERSION ?= V0.9.0
```
akrejcir, line 8: The validator version could be the same as ssp-operator version. `${VERSION}`.
The version should also be used in `internal/operands/template-validator/defaults.go`, so that the SSP operator deploys the same validator version.

go.mod
```diff
@@ -5,27 +5,41 @@ go 1.15
 require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/coreos/prometheus-operator v0.41.1
+	github.com/davecgh/go-spew v1.1.1
+	github.com/fsnotify/fsnotify v1.4.9
 	github.com/ghodss/yaml v1.0.0
+	github.com/go-kit/kit v0.10.0
```
akrejcir, line 11: Is this logger needed?

go.mod
```diff
@@ -5,27 +5,41 @@ go 1.15
...
 	github.com/operator-framework/operator-lib v0.4.0
 	github.com/spf13/cobra v1.1.1
+	github.com/spf13/pflag v1.0.5
 	gomodules.xyz/jsonpatch/v2 v2.1.0
 	gopkg.in/yaml.v2 v2.3.0
-	k8s.io/api v0.20.2
-	k8s.io/apiextensions-apiserver v0.20.1
-	k8s.io/apimachinery v0.20.2
+	k8s.io/api v0.20.4
+	k8s.io/apiextensions-apiserver v0.20.2
```
akrejcir, line 25: Can this be: `k8s.io/apiextensions-apiserver v0.20.4`? To be the same version as other k8s.io api clackages.

internal/template-validator/webhooks/utils.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	_ "github.com/go-kit/kit/log"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubevirt "kubevirt.io/client-go/api/v1"
```
akrejcir, line 12: I have noticed that these imports are different from the ones in `kubevirt-template-validator` repository. They are ok, but did you change anything else from the original code?

akrejcir (changes requested):
It looks good, but I have noticed that you have changed some code from the original repository.
In the future, please, when you are copying a lot of code from another place, it would be very useful to create 2 commits. The first one would be direct copy-pase, it may not even compile, and the other one would contain any new changes. It will make reviewing the code much easier.

But now, it is hard to separate the changes, so you can keep it this way. Do you remember what else did you change?

Also please add more information to the commit message.

vatsalparekh:
> In the future, please, when you are copying a lot of code fr
Not changing anything was the idea
But the only changes were mainly the dependencies related, and nothing with the core functional part of the code

Makefile
vatsalparekh: Changed now

go.mod
vatsalparekh: Corrected

go.mod
vatsalparekh: Not needed anymore
removed now

internal/template-validator/webhooks/utils.go
vatsalparekh: For this specific file, I included the changes from https://github.com/kubevirt/kubevirt-template-validator/pull/60
But not anything in other files

hack/build-and-push-validator.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [ -z "$QUAY_BOT_PASS" ] || [ -z "$QUAY_BOT_USER" ]; then
+	echo "missing QUAY_BOT_{USER,PASS} env vars"
+	exit 1
+fi
+
+echo "$QUAY_BOT_PASS" | docker login -u="$QUAY_BOT_USER" --password-stdin quay.io
+docker build -t quay.io/kubevirt/kubevirt-template-validator:$VERSION ./internal/template-validator/
+docker push quay.io/kubevirt/kubevirt-template-validator:$VERSION
+docker logout quay.io
```
akrejcir, line 15: The logic in this file is not needed now that the validator is part of ssp-operator repository.

Can you create a new commit, that would move the `docker build` and `docker push` commands to new Makefile targets, and remove this file? It is no longer needed to login to quay, because our Makefile already expects that the user is logged in. The github release scripts login before running build.

It would also be useful to extract the path to an env varaible in the makefile. Similarly to how SSP operator is build.

hack/build-template-validator.sh
```diff
@@ -0,0 +1,29 @@
...
+COMPONENT="kubevirt-template-validator"
+LDFLAGS="\
+-X 'github.com/kubevirt/kubevirt-template-validator/internal/pkg/version.COMPONENT=$COMPONENT' \
+-X 'github.com/kubevirt/kubevirt-template-validator/internal/pkg/version.VERSION=$TAG' "
+if git rev-parse &>/dev/null; then
+    BRANCH=$( git rev-parse --abbrev-ref HEAD )
+    REVISION=$( git rev-parse --short HEAD )
+    LDFLAGS="${LDFLAGS}\
+-X 'github.com/kubevirt/kubevirt-template-validator/internal/pkg/version.BRANCH=$BRANCH' \
+-X 'github.com/kubevirt/kubevirt-template-validator/internal/pkg/version.REVISION=$REVISION' "
```
akrejcir, line 20: These paths are no longer valid. Change them to:
```
kubevirt.io/ssp-operator/internal/template-validator/version
```

internal/template-validator/validation/rule_test.go
```diff
@@ -0,0 +1,94 @@
+package validation
```
akrejcir, line 1: This file was renamed from `rules_test.go`, can you return the original name?

internal/template-validator/validation/rule_test.go
```diff
@@ -0,0 +1,94 @@
...
+			ok, err := r.IsAppliableOn(vm)
+
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(ok).To(BeFalse())
+		})
+
+	})
+})
+
+func TestValidation(t *testing.T) {
```
akrejcir, line 91: This function runs all tests from this package, all `*_test.go` files. Can you move it to a separate file, like in the original repository?

internal/template-validator/validation/specializaed_test.go
```diff
@@ -0,0 +1,335 @@
+package validation
```
akrejcir, line 1: There is a typo in the filename. Should be `specialized_test.go`.

internal/template-validator/validation/util_test.go
```diff
@@ -0,0 +1,101 @@
+package validation
```
akrejcir, line 1: Can you restore the original file name: `utils_test.go`? Just to be consistent with other files in the repo, were we use `utils`.

internal/template-validator/validator/app.go
```diff
@@ -0,0 +1,101 @@
...
+	"kubevirt.io/ssp-operator/internal/template-validator/version"
+	"kubevirt.io/ssp-operator/internal/template-validator/virtinformers"
+	validating "kubevirt.io/ssp-operator/internal/template-validator/webhooks"
+)
+
+const (
+	defaultPort = 8443
+	defaultHost = "0.0.0.0"
+)
+
```
akrejcir, line 23: This file is missing the `init()` method that was added to it in the latest PR: https://github.com/kubevirt/kubevirt-template-validator/pull/59

This is the method: 
https://github.com/kubevirt/kubevirt-template-validator/blob/master/pkg/template-validator/app.go#L45

It adds the templates to the scheme. Previously, templates were added by the `init()` method in the imported module `github.com/fromanirh/okdutil/okd`. But it was inefficient to import a whole module for just one line of code.

internal/template-validator/version/version.go
```diff
@@ -0,0 +1,11 @@
+// Don't edit this file:
+// the vars below are meant to be overwritten by ../../../hack/build/build.sh.
```
akrejcir, line 2: This path is no longer valid, please update it.

akrejcir (changes requested):
I have checked the changes from the previous repository using diff functionality in the IDE, and I have a few comments on files.

Also, can you post another commit to change the file `internal/operands/template-validator/defaults.go`, to contain the current path and version of the validator image, that was build from make? You could:
- Change the constant to a varaible
- Use linker flag `-X` to set the variable value from `Makefile`. Like it is done in `build-template-validator.sh`

go.mod
```diff
@@ -5,27 +5,40 @@ go 1.15
...
-	k8s.io/apiextensions-apiserver v0.20.1
-	k8s.io/apimachinery v0.20.2
+	k8s.io/api v0.20.4
+	k8s.io/apiextensions-apiserver v0.20.4
+	k8s.io/apimachinery v0.20.4
 	k8s.io/client-go v12.0.0+incompatible
+	k8s.io/klog v1.0.0
 	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009
-	kubevirt.io/client-go v0.33.0
+	kubevirt.io/client-go v0.39.0
```
akrejcir, line 29: I think the CI failure is caused by this update of `kubevirt.io/client-go`. Can you try changing the version to `v0.38.1`? We can update to the latest in another PR.

vatsalparekh:
> I have checked the changes from the previous repository using diff functionality in the IDE, and I have a few comments on files.
> 
> Also, can you post another commit to change the file `internal/operands/template-validator/defaults.go`, to contain the current path and version of the validator image, that was build from make? You could:
> 
> * Change the constant to a varaible
> * Use linker flag `-X` to set the variable value from `Makefile`. Like it is done in `build-template-validator.sh`

@akrejcir but we build the operator binary in Dockerfile, you meant to refer passing the value from Makefile to docker build?

akrejcir:
> @akrejcir but we build the operator binary in Dockerfile, you meant to refer passing the value from Makefile to docker build?

Yes, you can add `ARG` instruction to the `Dockerfile` and then specify its value from Makefile.
https://docs.docker.com/engine/reference/builder/#arg

Also the `make manager` target needs to be changed, because it builds the operator.

Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Build the manager binary
 FROM golang:1.15 as builder
 
+ARG GoLinkerArgs
```
akrejcir, line 4: This argument will need to be specified also in CI.
It would be easier to pass only the validator image URL and tag, instead of generic linker arguments.

Makefile
```diff
@@ -54,7 +54,7 @@ functest: generate fmt vet manifests
 
 # Build manager binary
 manager: generate fmt vet
-	go build -o bin/manager main.go
+	go build -ldflags="-X 'kubevirt.io/ssp-operator/internal/operands/template-validator/template_validator.DefaultTemplateValidatorImage=${VALIDATOR_IMG}'" -o bin/manager main.go
```
akrejcir, line 57: Can you extract the build command into a shell script? For example `hack/build-operator.sh`. Then call it here and from the Dockerfile?

internal/operands/template-validator/defaults.go
```diff
@@ -1,5 +1,5 @@
 package template_validator
 
-const (
-	defaultTemplateValidatorImage = "quay.io/kubevirt/kubevirt-template-validator:v0.9.0"
+var (
+	DefaultTemplateValidatorImage = "quay.io/kubevirt/kubevirt-template-validator:v0.9.0"
```
akrejcir, line 4: Does the variable need to be exported? can you change its default value to something obviously incorrect, like `_VALIDATOR_IMAGE_PLACEHOLDER_`.

akrejcir (changes requested):

akrejcir:
Now that I think about it, the last commit will not pass CI. We will need to change the CI scripts, to build the validator, and then pass its path to the linker flags of ssp operator.

Can you put the last commit in another PR? Then we can change the CI scripts after the first PR is merged, but before the second one.

akrejcir:
The code looks good.
Maybe the CI is failing, because the kubevirt API needs to be added to scheme, here:
https://github.com/kubevirt/ssp-operator/blob/v0.2.0-rc.1/tests/tests_suite_test.go#L313

But curiously, the error message says:
```
multiple group-version-kinds associated with type *v1.VirtualMachine
```
I would expect that there is no GVK associated with v1.VirtualMachine, if it is missing from the scheme.

go.mod
```diff
@@ -5,27 +5,38 @@ go 1.15
...
-	k8s.io/apiextensions-apiserver v0.20.1
-	k8s.io/apimachinery v0.20.2
+	k8s.io/api v0.20.4
+	k8s.io/apiextensions-apiserver v0.20.4
+	k8s.io/apimachinery v0.20.4
 	k8s.io/client-go v12.0.0+incompatible
+	k8s.io/klog v1.0.0
 	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009
-	kubevirt.io/client-go v0.33.0
+	kubevirt.io/client-go v0.38.1
```
akrejcir, line 28: Can you decrease the version of `kubevirt.io/client-go` to `v0.36.0` for now?
It should fix the CI, and then we can increase it in another PR.

tests/tests_suite_test.go
```diff
@@ -47,6 +48,7 @@ const (
 var (
 	shortTimeout = 1 * time.Minute
 	timeout      = 10 * time.Minute
+	TestScheme   *runtime.Scheme
```
akrejcir, line 51: The scheme does not need to be exported.

tests/tests_suite_test.go
```diff
@@ -306,14 +308,17 @@ func expectSuccessOrNotFound(err error) {
 }
 
 func setupApiClient() {
-	Expect(sspv1beta1.AddToScheme(scheme.Scheme)).ToNot(HaveOccurred())
-	Expect(promv1.AddToScheme(scheme.Scheme)).ToNot(HaveOccurred())
-	Expect(templatev1.Install(scheme.Scheme)).ToNot(HaveOccurred())
-	Expect(secv1.Install(scheme.Scheme)).ToNot(HaveOccurred())
+	Expect(sspv1beta1.AddToScheme(TestScheme)).ToNot(HaveOccurred())
```
akrejcir, line 311: Basic kubernetes types need to be added to scheme:
```
import clientgoscheme "k8s.io/client-go/kubernetes/scheme"

...

Expect(clientgoscheme.AddToScheme(TestScheme)).ToNot(HaveOccurred())
```

akrejcir (changes requested):

tests/tests_common_test.go
```diff
@@ -2,6 +2,7 @@ package tests
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
akrejcir, line 5: Moving the import line is not needed.

akrejcir (commented):
Looks good. Can you change the commit message of the last commit? It does more than just add kubevirt API to scheme.

tests/tests_common_test.go
vatsalparekh: I didn't
I think it was the editor's import tool

vatsalparekh: done @akrejcir

akrejcir:
> Looks good. Can you change the commit message of the last commit? It does more than just add kubevirt API to scheme.

I meant that the last commit also changes `Makefile`, fixes paths in build script and adds `init()` function to the template validator main. But, you can keep it as is.

/lgtm
/approve

PR #137, opened 2021-03-22
Remove unused hack files

vatsalparekh:
Initially it was tried to build and run the functional tests inside a container, but it was half baked and was commented out. Doesn't seem like we need it currently.

**Release note**:

```release-note
NONE
```


ksimon1:
Instead of removing whole work, what has to be done to finish it?

vatsalparekh:
> Instead of removing whole work, what has to be done to finish it?

@ksimon1 I'm not sure if we want to use it anymore
iiric, It would also be difficult to make it work with CI

@akrejcir thoughts?

akrejcir:
CI runs test just by calling `make functests`, so it can build the test any way we want.

I don't think it is necessary to build tests in a container. We use `vendor` to store all our dependencies, and golang compiler versions should be backward compatible, so even if there is a newer version in the test environment, the tests would build correctly.

ksimon1:
ok, thanks for explanation.
/lgtm

PR #140, opened 2021-03-31
Add a dry-run to verify placement fields

omeryahud:
This PR adds a deployment creation dry-run to the SSP webhook to reject any invalid placement fields.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1938241

- Still missing tests

**Release note**:

```release-note
None
```

### DIFF
--- a/api/v1beta1/ssp_webhook.go
+++ b/api/v1beta1/ssp_webhook.go
@@ -20,8 +20,13 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/pkg/errors"
+	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+	"kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -65,6 +70,10 @@ func (r *SSP) ValidateCreate() error {
 		return fmt.Errorf("creation failed, the configured namespace for common templates does not exist: %v", namespaceName)
 	}
 
+	if err = validatePlacement(r); err != nil {
+		return errors.Wrap(err, "placement api validation error")
+	}
+
 	return nil
 }
 
@@ -79,6 +88,10 @@ func (r *SSP) ValidateUpdate(old runtime.Object) error {
 			r.Spec.CommonTemplates.Namespace)
 	}
 
+	if err := validatePlacement(r); err != nil {
+		return errors.Wrap(err, "placement api validation error")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


api/v1beta1/ssp_webhook.go
```diff
@@ -65,6 +69,11 @@ func (r *SSP) ValidateCreate() error {
 		return fmt.Errorf("creation failed, the configured namespace for common templates does not exist: %v", namespaceName)
 	}
 
+	err = validatePlacement(r)
```
marceloamaral, line 72: For simplicity and coding size, I like the error testing in only one line:

if err = validatePlacement(r); err != nil {
	return err
}

api/v1beta1/ssp_webhook.go
```diff
@@ -91,3 +105,71 @@ func (r *SSP) ValidateDelete() error {
...
+	err = validateOperandPlacement(ssp.Spec.TemplateValidator.Placement)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateOperandPlacement(placement *api.NodePlacement) error {
+	if placement == nil {
```
marceloamaral, line 124: Shouldn't this return an error saying that there is not placement rules to be applied?

omeryahud: No. No placement rules is acceptable, it basically tells the k8s scheduler that we don't care where our workloads are placed, we just want them running

api/v1beta1/ssp_webhook.go
omeryahud: Done

api/v1beta1/ssp_webhook.go
```diff
@@ -91,3 +104,67 @@ func (r *SSP) ValidateDelete() error {
 func setClientForWebhook(c client.Client) {
 	clt = c
 }
+
+func validatePlacement(ssp *SSP) error {
+	if err := validateOperandPlacement(ssp.Spec.NodeLabeller.Placement); err != nil {
```
ksimon1, line 109: This is no longer needed

tests/webhook_test.go
```diff
@@ -39,55 +87,131 @@ var _ = Describe("Validation webhook", func() {
...
 				}
 				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf(
 					"creation failed, the configured namespace for common templates does not exist: %v",
-					foundSsp.Spec.CommonTemplates.Namespace,
+					newSsp.Spec.CommonTemplates.Namespace,
 				)))
 			})
+
+			Context("Placement API validation", func() {
+				Context("node-labeller", func() {
```
ksimon1, line 128: This is no longer needed

api/v1beta1/ssp_webhook.go
omeryahud: Removed

tests/webhook_test.go
omeryahud: Removed

api/v1beta1/ssp_webhook.go
```diff
@@ -91,3 +104,63 @@ func (r *SSP) ValidateDelete() error {
...
 	clt = c
 }
+
+func validatePlacement(ssp *SSP) error {
+	if err := validateOperandPlacement(ssp.Spec.TemplateValidator.Placement); err != nil {
+		return err
+	}
+
+	return nil
+}
```
akrejcir, line 114: This function can be inlined, or simplified to:
```
return validateOperandPlacement(ssp.Spec.TemplateValidator.Placement)
```

tests/webhook_test.go
```diff
@@ -39,55 +87,100 @@ var _ = Describe("Validation webhook", func() {
...
 				}
 				Expect(err.Error()).To(ContainSubstring(fmt.Sprintf(
 					"creation failed, the configured namespace for common templates does not exist: %v",
-					foundSsp.Spec.CommonTemplates.Namespace,
+					newSsp.Spec.CommonTemplates.Namespace,
 				)))
 			})
+
+			Context("Placement API validation", func() {
+				Context("template-validator", func() {
```
akrejcir, line 128: This second `Context` is not needed. Each `It` test case already mentions `template-validator`.

tests/webhook_test.go
```diff
@@ -39,55 +87,100 @@ var _ = Describe("Validation webhook", func() {
...
-			foundSsp := getSsp()
 			originalNs := foundSsp.Spec.CommonTemplates.Namespace
 			foundSsp.Spec.CommonTemplates.Namespace = originalNs + "-updated"
 			err := apiClient.Update(ctx, foundSsp)
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring("commonTemplates.namespace cannot be changed."))
 		})
+
+		Context("Placement API validation", func() {
+			Context("template-validator", func() {
```
akrejcir, line 171: The same comment as above.

akrejcir (commented):
Looks good. I only have a few optional code style comments. You don't have to change them if you want to save time.

/lgtm

Is this still on hold? The test case is there.

api/v1beta1/ssp_webhook.go
omeryahud: Yes, I agree, I will change that

PR #141, opened 2021-03-31
move deprecated label to annotations

ksimon1:
move deprecated label to annotations

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1925019

**Release note**:
```
NONE
```

### DIFF
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -327,6 +327,17 @@ var _ = Describe("Common templates", func() {
 					updatedTpl.Labels[commonTemplates.TemplateVersionLabel] == "not-latest"
 			}, shortTimeout).Should(BeTrue(), "labels were not removed from older templates")
 		})
+		It("[test_id:5969]: should add deprecated annotation to old templates", func() {
+			triggerReconciliation()
+
+			Eventually(func() bool {
+				updatedTpl := &templatev1.Template{}
+				key := client.ObjectKey{Name: oldTemplate.Name, Namespace: oldTemplate.Namespace}
+				err := apiClient.Get(ctx, key, updatedTpl)
+				return err == nil &&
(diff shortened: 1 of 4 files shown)


tests/commonTemplates_test.go
```diff
@@ -327,6 +327,17 @@ var _ = Describe("Common templates", func() {
 					updatedTpl.Labels[commonTemplates.TemplateVersionLabel] == "not-latest"
 			}, shortTimeout).Should(BeTrue(), "labels were not removed from older templates")
 		})
+		It("should add deprecated annotation to old templates", func() {
```
rnetser, line 330: Test ID CNV-5969

If we add "deprecated" annotation to a new template, will it be reconciled?

tests/commonTemplates_test.go
```diff
@@ -327,6 +327,17 @@ var _ = Describe("Common templates", func() {
...
+			triggerReconciliation()
+
+			Eventually(func() bool {
+				updatedTpl := &templatev1.Template{}
+				key := client.ObjectKey{Name: oldTemplate.Name, Namespace: oldTemplate.Namespace}
+				err := apiClient.Get(ctx, key, updatedTpl)
+				return err == nil &&
+					updatedTpl.Annotations[commonTemplates.TemplateDeprecatedAnnotation] == ""
```
akrejcir, line 338: The annotation should equal `"true"`, not `""`.

akrejcir (changes requested):

tests/commonTemplates_test.go
akrejcir: No. New annotations are not modified by the reconciler.

akrejcir: I misunderstood your question.
The new template will be reconciled, but the annotation will not be removed.

PR #147, opened 2021-04-14
Update API version in unused config file

akrejcir:
The config was generated by operator SDK and is currently not used, but it may be used in the future.
Updating its API in case we use it.

**Release note**:
```release-note
None
```


akrejcir:
CI can be ignored, because the changed file is not used.

/override ci/prow/e2e-upgrade-from-old-operator-functests

PR #150, opened 2021-04-19
Test template-validator certificate rotation

akrejcir:
The test checks the TLS certificate returned by the template validator webhook server.
The port-forwarding code was mostly inspired by the implementation in kubectl.

This PR implements an existing test from the old repository:
https://github.com/kubevirt/kubevirt-template-validator/blob/master/functests/13-test-certificate-rotation.sh

**Release note**:
```release-note
None
```


vatsalparekh:
@akrejcir Can you rebase this once #161 is merged, to make sure we're testing this against the correct validator code

PR #161, opened 2021-04-30
Changes to inject template-validator image to ssp-operator build

vatsalparekh:

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,11 @@ install: manifests kustomize
 uninstall: manifests kustomize
 	$(KUSTOMIZE) build config/crd | $(OC) delete -f -
 
+manager-envsubst:
+	cd config/manager && VALIDATOR_IMG=${VALIDATOR_IMG} envsubst < manager.template.yaml > manager.yaml
+
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
-deploy: manifests kustomize
+deploy: manifests kustomize manager-envsubst
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default | $(OC) apply -f -
 
@@ -152,7 +155,7 @@ endif
 
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
-bundle: kustomize operator-sdk manifests csv-generator
+bundle: kustomize operator-sdk manifests csv-generator manager-envsubst
 	./operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | ./operator-sdk generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
--- a/config/manager/manager.template.yaml
+++ b/config/manager/manager.template.yaml
@@ -30,6 +30,7 @@ spec:
         env:
           - name: KVM_INFO_IMAGE
           - name: VALIDATOR_IMAGE
+            value: "$VALIDATOR_IMG"
           - name: VIRT_LAUNCHER_IMAGE
           - name: NODE_LABELLER_IMAGE
           - name: CPU_PLUGIN_IMAGE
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Build the manager binary
 FROM golang:1.15 as builder
```
kwiesmueller, line 2: Builds fail because this is being pulled from dockerhub :-/

Makefile
```diff
@@ -74,6 +74,7 @@ uninstall: manifests kustomize
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests kustomize
+	cd config/manager && VALIDATOR_IMG=${VALIDATOR_IMG} envsubst < manager.template.yaml > manager.yaml
```
akrejcir, line 77: Can you extract this template substitution to a separate make target? It will be needed for `make bundle` too.

config/manager/manager.template.yaml
```diff
@@ -0,0 +1,45 @@
+kind: ServiceAccount
```
akrejcir, line 1: Can you remove the `manager.yaml` file and add it to `.gitignore`? It will be recreated in makefile using this template.

internal/operands/template-validator/reconcile.go
```diff
@@ -137,6 +138,11 @@ func reconcileService(request *common.Request) (common.ResourceStatus, error) {
 func reconcileDeployment(request *common.Request) (common.ResourceStatus, error) {
 	validatorSpec := request.Instance.Spec.TemplateValidator
 	image := getTemplateValidatorImage()
+	request.Logger.Info(fmt.Sprintf("Template-Validator image: %s", image))
+	request.Logger.Info(fmt.Sprintf("VALIDATOR_IMAGE Envirnoment variable: %s", os.Getenv(common.TemplateValidatorImageKey)))
```
akrejcir, line 142: These lines will pollute the log, because they will be written on each reconcilation iteration, but the image is not changing.

Can you move the logging somewhere where it will be called only once at the start of the operator?

akrejcir (changes requested):

internal/operands/template-validator/reconcile.go
vatsalparekh: These were put here to make sure we know for this PR which image is being used, once this works fine, I think having them inside the panic block only is fine

internal/operands/template-validator/reconcile.go
```diff
@@ -137,6 +138,10 @@ func reconcileService(request *common.Request) (common.ResourceStatus, error) {
 func reconcileDeployment(request *common.Request) (common.ResourceStatus, error) {
 	validatorSpec := request.Instance.Spec.TemplateValidator
 	image := getTemplateValidatorImage()
+	if image == "" {
+		request.Logger.Info(fmt.Sprintf("VALIDATOR_IMAGE Envirnoment variable: %s", os.Getenv(common.TemplateValidatorImageKey)))
```
akrejcir, line 142: If `image == ""` the env var `VALIDATOR_IMAGE` will always be `""`, so this log line is not needed.

akrejcir (changes requested):
Can you squash the commits? The change is small, so I think it is not needed to split it into multiple.

config/manager/manager.template.yaml
```diff
@@ -30,6 +30,7 @@ spec:
         env:
           - name: KVM_INFO_IMAGE
           - name: VALIDATOR_IMAGE
+            value: $VALIDATOR_IMG
```
akrejcir, line 33: I think the variable needs to be quoted `"$VALIDATOR_IMG"`. Otherwise, if it was empty, the yaml would not be valid.

internal/operands/template-validator/reconcile.go
vatsalparekh: Removed

config/manager/manager.template.yaml
vatsalparekh: Good catch, changed now

kwiesmueller:
/lgtm
/approve
@akrejcir feel free to approve as well (I can't)

akrejcir:
@vatsalparekh Can you write some description for this PR?

vatsalparekh:
@akrejcir done

PR #168, opened 2021-05-10
Improve code: Use constants instead of string literals

akrejcir:
Fixed code smells found by SonarCloud.

**Release note**:
```release-note
None
```


kwiesmueller:
/retest
/lgtm
The security hotspot could be resolved afterwards as it seems to be this IP intentionally right?

akrejcir:
> The security hotspot could be resolved afterwards as it seems to be this IP intentionally right?

Yes, that is why I changed the IP.

PR #177, opened 2021-06-02
Add negative service account test

vatsalparekh:
This adds a missed test from https://github.com/kubevirt/kubevirt-ssp-operator/blob/master/functests/12-test-os-imaes-namespace_rbac.sh#L224-L230

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/operands/common-templates/constants.go
+++ b/internal/operands/common-templates/constants.go
@@ -11,6 +11,6 @@ const (
 	TemplateWorkloadLabelPrefix  = "workload.template.kubevirt.io/"
 	TemplateDeprecatedAnnotation = "template.kubevirt.io/deprecated"
 
-	CdiApiGroup = "cdi.kubevirt.io"
+	CdiApiGroup   = "cdi.kubevirt.io"
 	CdiApiVersion = "v1beta1"
 )
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -656,6 +656,18 @@ var _ = Describe("Common templates", func() {
 							},
 						}),
 				)
+				It("[test_id:4878]should not create any other resurces than the ones listed in the Edit Cluster role", func() {
+					sars := &authv1.SubjectAccessReviewSpec{
+						ResourceAttributes: &authv1.ResourceAttributes{
+							Verb:      "create",
+							Namespace: commonTemplates.GoldenImagesNSname,
+							Version:   core.SchemeGroupVersion.Version,
+							Resource:  "pods",
+						},
+					}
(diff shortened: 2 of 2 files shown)


internal/operands/common-templates/constants.go
```diff
@@ -11,6 +11,6 @@ const (
 	TemplateWorkloadLabelPrefix  = "workload.template.kubevirt.io/"
 	TemplateDeprecatedAnnotation = "template.kubevirt.io/deprecated"
 
-	CdiApiGroup = "cdi.kubevirt.io"
+	CdiApiGroup   = "cdi.kubevirt.io"
```
sarahbx, line 14: Let's drop this change, unrelated to pr

kwiesmueller: +1
I think gofmt should drop this anyways, don't we have a generation step for formating like in Kubevirt?

kwiesmueller: Once this is fixed and CI is green, lgtm from me.

tests/commonTemplates_test.go
```diff
@@ -656,6 +656,18 @@ var _ = Describe("Common templates", func() {
 							},
 						}),
 				)
+				It("[test_id:4878]should not create any other resurces than the ones listed in the Edit Cluster role", func() {
```
shwetaap, line 659: We have duplicate entries of the test_id: 4876 - https://github.com/kubevirt/ssp-operator/blob/master/tests/commonTemplates_test.go#L638-L657. I thought the 4878 block was going to replace one of those duplicate blocks. But looks like it is not. Could you remove the duplicate entry?

internal/operands/common-templates/constants.go
akrejcir: > +1
> I think gofmt should drop this anyways, don't we have a generation step for formating like in Kubevirt?

We do: `make fmt` and `make vet`, they are also run on any other build target.

vatsalparekh: This was done by either `make fmt` I guess, and it keeps coming back everytime, so I think it should be fixed

tests/commonTemplates_test.go
vatsalparekh: Removed it now

internal/operands/common-templates/constants.go
sarahbx: Looked at this locally, I see what it is doing now, lining up the `=`. 👍

sarahbx (approved):
/lgtm

Issue #178, opened 2021-06-07
Modify build scripts to update the default version of template validator image.


akrejcir:
It would be convenient, if the build process updates the default template validator image version in the file [internal/operands/template-validator/defaults.go](https://github.com/kubevirt/ssp-operator/blob/v0.11.1/internal/operands/template-validator/defaults.go#L4)

It should be equal to the env variable [`VALIDATOR_IMG`](https://github.com/kubevirt/ssp-operator/blob/v0.11.1/Makefile#L25).

/kind enhancement

**Environment**:
- SSP Operator version: `v0.11.1`

vatsalparekh:
I think it can be easily done in github actions

akrejcir:
I think it would be better to do it in `Makefile`. It would be useful for development, that I could just write:
```
make build-template-validator-container push-template-validator-container
make container-build container-push
make deploy
```
And it would deploy latest ssp with the latest template validator.

akrejcir:
Currently, it also works, but with the "hack" that we define the value of env variable in the deployment manifest:
https://github.com/kubevirt/ssp-operator/blob/v0.11.1/config/manager/manager.template.yaml#L33

PR #179, opened 2021-06-07
csv-generator: Override environment variables only when needed

akrejcir:
The csv-generator does not override environment variables `VALIDATOR_IMAGE` and `OPERATOR_VERSION`, if no override flag is specified on the command line.

**Release note**:
```release-note
None
```


omeryahud:
We should also update https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/template-validator/defaults.go#L4 on every release, I will submit a PR to do just that now

akrejcir:
> We should also update https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/template-validator/defaults.go#L4 on every release, I will submit a PR to do just that now

I wanted to create some automation in `Makefile` for it: https://github.com/kubevirt/ssp-operator/issues/178

akrejcir:
I see your PR does it, so you can ignore my previous comment.

PR #180, opened 2021-06-08
Update the default template-validator image to the latest released one

omeryahud:
Update the template validator image deployed by the SSP operator to the
latest released one on every release.

The default template validator image should no longer be manually updated for each release

**Release note**:

```release-note
NONE
```
/cc @akrejcir @kwiesmueller

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,14 @@
 # Build the manager binary
 FROM golang:1.15 as builder
 
+# Consume required variables so we can work with make
+ARG IMG_REPOSITORY
+ARG IMG_TAG
+ARG IMG
+ARG VALIDATOR_REPOSITORY
+ARG VALIDATOR_IMG_TAG
+ARG VALIDATOR_IMG
+
 WORKDIR /workspace
 # Copy the Go Modules manifests and vendor directory
 COPY go.mod go.mod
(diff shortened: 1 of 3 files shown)


akrejcir:
Related to: https://github.com/kubevirt/ssp-operator/issues/178

akrejcir (changes requested):
The `Dockerfile` needs to be updated too.
The `make manager` target is probably not used anywhere. Maybe only for local building.

Maybe we can call makefile targets directly from the `Dockerfile` ?

omeryahud:
> The `Dockerfile` needs to be updated too.
Ah right, I assumed it was used in the Dockerfile, I will update it as well

> The `make manager` target is probably not used anywhere. Maybe only for local building.
> 
> Maybe we can call makefile targets directly from the `Dockerfile` ?
Sounds like a good idea,  I will check it out

omeryahud:
@akrejcir Can you take another look?

Dockerfile
```diff
@@ -8,24 +8,25 @@ COPY go.sum go.sum
...
 COPY api/ api/
 COPY controllers/ controllers/
 COPY internal/ internal/
 
-COPY hack/csv-generator.go csv-generator.go
+COPY hack/csv-generator.go hack/csv-generator.go
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on make manager
```
akrejcir, line 20: Maybe the docker argument needs to be passed here as en env variable to make? I haven't checked it, so I'm not sure.

Makefile
```diff
@@ -106,7 +108,14 @@ container-build: unittest bundle
 	mkdir -p data/crd
 	cp bundle/manifests/ssp-operator.clusterserviceversion.yaml data/olm-catalog/ssp-operator.clusterserviceversion.yaml
 	cp bundle/manifests/ssp.kubevirt.io_ssps.yaml data/crd/ssp.kubevirt.io_ssps.yaml
-	docker build -t ${IMG} .
+	docker build -t ${IMG} \
+		--build-arg IMG_REPOSITORY=${IMG_REPOSITORY} \
```
akrejcir, line 112: Do these build arguments need to be defined inside the Dockerfile using `ARG` instruction?

omeryahud: Yes, you are right, I thought it was implicit, but it isn't, fixed now

Dockerfile
omeryahud: It is enough that the env vars are consumed with `ARG`, so it should be good now

akrejcir: Ok

akrejcir (commented):
/lgtm

akrejcir:
CI fails, because there was an issue with latest kubevirt release. They are working on it.

akrejcir:
New kubevirt version was released. 

/retest

PR #182, opened 2021-06-13
Enabling monitoring and adding basic alerts

borod108:
This PR enables basic metrics functionality for the ssp-operator.
In particular it:
1) Exposes metrics endpoints on the ssp-operator and template-validator pods.
2) Adds some basic recording rules and alerts.

```release-note
Exposing metrics for the ssp-operator to be gathered by Prometheus and part of kubevirt metrics.
Adding "Critical Alerts" that will fire in case the ssp-operator or template-validator pods are down.
```


main.go
```diff
@@ -67,14 +71,62 @@ func init() {
...
+		}
+	default:
+	}
+
+	return nil
+}
+
+func (m *promMonitor) startMonitor() {
+	for {
+		time.Sleep(400 * time.Millisecond)
```
ksimon1, line 100: Is this sleep necessary?

internal/operands/template-validator/resources.go
```diff
@@ -34,6 +36,12 @@ func commonLabels() map[string]string {
 	}
 }
 
+func podLabels() map[string]string {
+	labels := commonLabels()
+	labels[PrometheusLabel] = ""
```
kwiesmueller, line 41: do you want the label empty or gone?
If gone you could do `delete(labels, PrometheusLabel)` instead.

kwiesmueller: Nvm, I think I misunderstood what you are trying to achieve here.

main.go
```diff
@@ -89,9 +141,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	promErrCh := make(chan error)
+	go runPrometheusServer(metricsAddr, promErrCh)
+	monitor := newPromMonitor(promErrCh)
```
kwiesmueller, line 146: Is the whole monitor required?
From what it does right now, I think you could also just log the error inside `runPrometheusServer` as nothing else happens when the server fails.

kwiesmueller (changes requested):

main.go
borod108: nope, and like @kwiesmueller  pointed out, it is all useless :) so now it is gone :)

main.go
borod108: You are right - this is useless. But it looked so smart and fancy :p

config/samples/ssp_v1beta1_ssp.yaml
```diff
@@ -8,3 +8,5 @@ spec:
     namespace: kubevirt
   templateValidator:
     replicas: 2
+  metrics:
+    namespace: kubevirt
```
kwiesmueller, line 12: missing newline at eof

internal/operands/template-validator/resources.go
```diff
@@ -17,6 +17,7 @@ import (
 
 const (
 	ContainerPort          = 8443
+	MetricsPort            = 8443
```
kwiesmueller, line 20: can they both run on the same port?

internal/operands/template-validator/resources.go
kwiesmueller: Honestly this part is still confusing.
Why are you setting this label to empty?

main.go
```diff
@@ -67,14 +70,30 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+func runPrometheusServer(metricsAddr string) {
+	setupLog.Info("Entered runPrometheusServer")
+	mux := restful.NewContainer()
+	mux.Handle("/metrics", promhttp.Handler())
+	setupLog.Info("Defining prom server")
```
kwiesmueller, line 77: maybe write Prometheus in full as the admin will see those logs

main.go
```diff
@@ -67,14 +70,30 @@ func init() {
...
+	server := http.Server{
+		Addr:    metricsAddr,
+		Handler: mux,
+	}
+	setupLog.Info("Starting prom server with TLS")
+	err := server.ListenAndServeTLS(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
+	if err != nil {
+		setupLog.Error(err, "Failed to start metrics server")
```
kwiesmueller, line 85: Or call it metrics everywhere, like here

main.go
```diff
@@ -89,9 +108,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	go runPrometheusServer(metricsAddr)
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
+		MetricsBindAddress:     "0",
```
kwiesmueller, line 115: That makes us loose existing metrics exported by controller-runtime :-/
Can we merge those metrics somehow so we add to them instead of replacing them?

internal/operands/template-validator/resources.go
borod108: Because this how kubevirt defined the Prometheus server to pick it up.

https://github.com/kubevirt/kubevirt/blob/099be903803556bfe3a85075a7b55f0a711d9ca7/pkg/virt-operator/resource/generate/components/crds.go#L506

https://github.com/kubevirt/kubevirt/blob/099be903803556bfe3a85075a7b55f0a711d9ca7/pkg/virt-operator/resource/generate/components/crds.go#L43

and it works :)

internal/operands/template-validator/resources.go
borod108: yes they can and they do :)

config/samples/ssp_v1beta1_ssp.yaml
borod108: adding

main.go
borod108: you are right

main.go
borod108: fixed

internal/operands/metrics/resources.go
```diff
@@ -20,13 +20,47 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
...
+						},
+						{
+							Alert: "SSPDown",
+							Expr:  intstr.FromString("num_of_running_ssp_operators == 0"),
+							For:   "5m",
+							Annotations: map[string]string{
+								"summary": "All SSP operator pods are down.",
+							},
+							Labels: map[string]string{
+								"severity": "Critical",
```
akrejcir, line 47: Where are the `severity` levels defined? Maybe `Critical` is too big severity?

For me, critical severity means: "Wake up the cluster admin in the middle of the night to fix this error." Maybe that is too much for missing SSP operator :).

internal/operands/template-validator/reconcile.go
```diff
@@ -2,6 +2,7 @@ package template_validator
 
 import (
 	"fmt"
+
```
akrejcir, line 5: This was probably changed by the IDE, but nothing else is changed in this file.
Can you remove this change?

main.go
```diff
@@ -67,14 +70,30 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+func runPrometheusServer(metricsAddr string) {
+	setupLog.Info("Entered runPrometheusServer")
+	mux := restful.NewContainer()
```
akrejcir, line 75: Could you use the `http.ServeMux` structure from `net/http` standard library, instead of importing `github.com/emicklei/go-restful` package?

It should be enough to just replace this line with:
```
mux := http.NewServeMux()
```

main.go
```diff
@@ -67,14 +70,30 @@ func init() {
...
+	mux.Handle("/metrics", promhttp.Handler())
+	setupLog.Info("Defining Prometheus metrics endpoint server")
```
akrejcir, line 77: I think this log output and the one above are unnecessary. They would probably not be useful for debugging.

If you want, you can move the log command from line 81, to the top of the function.

akrejcir (changes requested):

internal/operands/metrics/resources.go
akrejcir: We discussed this offline. You can ignore this comment.

main.go
borod108: done.

internal/operands/template-validator/reconcile.go
borod108: ok

main.go
borod108: ok left just one log entry

borod108:
@kwiesmueller @akrejcir thank you for the reviews, I made changes accordingly, does it look ok to you now?

main.go
borod108: yes definitely.

internal/operands/template-validator/resources.go
kwiesmueller: Alright, thanks for clarifying. Fascinating!

main.go
```diff
@@ -67,14 +71,38 @@ func init() {
 	// +kubebuilder:scaffold:scheme
 }
 
+func runPrometheusServer(metricsAddr string) {
+	setupLog.Info("Starting Prometheus metrics endpoint server with TLS")
+	// we need to unregister a couple of collectors, otherwise the metrics gathering will fail
+	// due to duplicate metrics collection
+	metrics.Registry.Unregister(prometheus.NewGoCollector())
```
kwiesmueller, line 78: I wonder where the duplicates come from.
Is it the DefaultGatherer below?

main.go
kwiesmueller: Thanks, that looks better!

main.go
borod108: yes

borod108:
@kwiesmueller @akrejcir do we need anything else?

akrejcir (changes requested):
Can you use our usual template for PR's description?
https://raw.githubusercontent.com/kubevirt/ssp-operator/master/.github/PULL_REQUEST_TEMPLATE.md

Also, there are typos in the commit message: `tmeplate` and `recodring`.

main.go
```diff
@@ -89,9 +117,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	go runPrometheusServer(metricsAddr)
```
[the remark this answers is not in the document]
kwiesmueller: Do we really only want those metrics on leader election?
I think the cache etc. runs anyways, no? So the metrics can be useful :thinking:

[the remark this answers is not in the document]
kwiesmueller: I mean the controler workers running all the watches etc. I thought they always run as well.
So workqueue length metrics still could matter.

If it's like in kubevirt I think we should also collect all pods. @marceloamaral or @nelsonspbr might have some more thoughts on that.

config/samples/ssp_v1beta1_ssp.yaml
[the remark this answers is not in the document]
borod108: If I remove this, a person trying out the sample will not see any of the metric rules and alerts, I thought it would be nice if they will.

main.go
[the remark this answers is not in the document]
borod108: "/metrics" is the end point that Prometheus queries to collect metrics from the pod, and we want to have such an endpoint on every pod, that way we can get the metrics from all of them (and count the number of pods running :))

main.go
borod108: ok, I will remove it and if we need anything from it in the future we can add it then.

config/samples/ssp_v1beta1_ssp.yaml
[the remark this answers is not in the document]
akrejcir: I don't understand. The API for SSP CR is defined here: https://github.com/kubevirt/ssp-operator/blob/ad54715f1183481d808d0dc80143758eb2265a28/api/v1beta1/ssp_types.go#L51

And it does not contain any `metrics` field. Where does this field come from?

main.go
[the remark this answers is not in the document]
akrejcir: Ok, I'm not familiar with how metrics and prometheus work in openshift. If you think it makes sense for a pod to report metrics even if it looses leader election, then a goroutine is fine.

When an operator pod looses leader election, it will only run webhook service.

If you use a goroutine, can you only wrap the `ListenAndServeTLS()` method in a goroutine, not the whole function `runPrometheusServer()` ?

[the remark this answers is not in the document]
marceloamaral: Regarding workqueue metrics for the SSP controller, it might be useful to track. But I am not sure if It could be enabled as we enabled for the internal Kubevirt controllers, as [here](https://github.com/marceloamaral/kubevirt/blob/46a3f061ecb8af5207f7cdc675ea8ec7a05fa79e/pkg/monitoring/workqueue/prometheus/prometheus.go). The workqueue must have a name, and I did not see in the SSP controller a transparent workqueue creation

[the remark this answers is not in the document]
borod108: @akrejcir I raised this question with the monitoring team, and it seems our approach in OCP is to gather metrics from all pods. Also - moved the goroutine as you requested. Would like to know why this is better though.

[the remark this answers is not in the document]
akrejcir: In my opinion, this is better for code readability. My reasoning is the following:

When I see a goroutine started like this in a function:
```go
func function() {
    // ... do things ...

    go runPrometheusServer()

    // ... do other things ...
}
```

It looks like a probable bug, because there is no visible way how to control the new goroutine. There is no way to wait for its completion or signal it to stop. I would need to read the implementation of the function `runPrometheusServer()` to know that it is intentional.

Another general reason, that maybe does not apply in this case, is that the caller of this function needs to know that it has to be started as a goroutine. So they need to read the documentation or implementation.


When the goroutine is started inside the function as closure, like in this case:
```go
func runPrometheusServer() {
    server := // ... setup server ...

    go func() {
        err := server.ListenAndServeTLS(...)
        // ... handle error ...
    }()
}
```

It is easier to see what it does, and why it needs to be a goroutine. Also, that it is intentional, that there is no way to control the goroutine.


But this is only my opinion, code readability is subjective.

config/samples/ssp_v1beta1_ssp.yaml
```diff
@@ -7,4 +7,4 @@ spec:
   commonTemplates:
     namespace: kubevirt
   templateValidator:
-    replicas: 2
+    replicas: 2
```
akrejcir, line 10: Missing new line at the end of the file.

data/olm-catalog/ssp-operator.clusterserviceversion.yaml
```diff
@@ -16,6 +16,9 @@ metadata:
             "commonTemplates": {
               "namespace": "kubevirt"
             },
+            "metrics": {
+              "namespace": "kubevirt"
+            },
```
akrejcir, line 21: This file needs to be regenerated. See my general comment.

internal/operands/template-validator/resources.go
```diff
@@ -159,7 +167,12 @@ func newDeployment(namespace string, replicas int32, image string) *apps.Deploym
 							Name:          "webhook",
 							ContainerPort: ContainerPort,
 							Protocol:      core.ProtocolTCP,
-						}},
+						},
+							{
+								Name:          "metrics",
+								ContainerPort: MetricsPort,
+								Protocol:      core.ProtocolTCP,
+							}},
```
akrejcir, line 175: Can you fix the indentation here to one of these:
```go
Ports: []core.ContainerPort{{
    Name:          "webhook",
    ContainerPort: ContainerPort,
    Protocol:      core.ProtocolTCP,
},{
    Name:          "metrics",
    ContainerPort: MetricsPort,
    Protocol:      core.ProtocolTCP,
}}
```
or
```go
Ports: []core.ContainerPort{
    {
        Name:          "webhook",
        ContainerPort: ContainerPort,
        Protocol:      core.ProtocolTCP,
    }, {
        Name:          "metrics",
        ContainerPort: MetricsPort,
        Protocol:      core.ProtocolTCP,
    },
},
```

internal/template-validator/validator/app.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	templatev1 "github.com/openshift/api/template/v1"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
akrejcir, line 8: This dependency is not listed in `go.mod` file and not downloaded in `vendor` directory. It will be added automatically if you run:
```
go mod tidy
go mod vendor
```

main.go
```diff
@@ -70,11 +73,32 @@ func init() {
...
+	go func() {
+		err := server.ListenAndServeTLS(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
+		if err != nil {
+			setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
+		}
+	}()
+}
+
+// Give permissions to use leases for leader election.
+// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch;create;update;patch;delete
```
akrejcir, line 95: This `kubebuilder` annotation is no longer needed. It is probably an artifact of rebasing this PR. You can remove it.

akrejcir (changes requested):
Functionality looks good, but the dependencies and one file need to be regenerated.
You can do that by running:
```bash
go mod tidy  # Adds the github.com/prometheus/client_golang dependency to go.mod file
go mod vendor # Downloads the dependency to vendor folder
make container-build # Regenerates various files. One of them is data/olm-catalog/ssp-operator.clusterserviceversion.yaml
```

akrejcir:
I noticed that you use markdown syntax in the commit message. It is not needed, the commit message is a plain text that should describe why the change was done. It is useful when using `git blame` or an IDE to help understand the code.

config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -22,21 +22,27 @@ spec:
         description: SSP is the Schema for the ssps API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
```
akrejcir, line 25: All changes in this file are just adding newlines to `description` fields. I'm not sure why, but for some developers the `operator-sdk` adds newlines and for some it does not. It probably depends on the environment. We will need to investigate later.

Anyway, this change is not part of this PR, so you don't need to commit this file.

data/crd/ssp.kubevirt.io_ssps.yaml
```diff
@@ -20,21 +20,27 @@ spec:
         description: SSP is the Schema for the ssps API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
```
akrejcir, line 23: See my comment on `config/crd/bases/ssp.kubevirt.io_ssps.yaml`.

akrejcir (changes requested):
Looks good, just 2 files don't need to be committed.

borod108:
@sradco please take a look :)

PR #185, opened 2021-06-21
fixed deprecated label which is set during upgrade to up-to-date template

ksimon1:
fixed deprecated label which is set during upgrade to up-to-date template
When deprecated label is set to old templates and new templates are deployed with the same name, during update the deprecated label stays on updated template. 
This patch updates the behaviour, so at first it loads all templates which will be deployed and then it applies deprecated label only to templates which will not be updated.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1964583

**Release note**:
```
NONE
```

### DIFF
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -22,6 +22,7 @@ import (
 var (
 	loadTemplatesOnce sync.Once
 	templatesBundle   []templatev1.Template
+	deployedTemplates = make(map[string]bool)
 )
 
 // Define RBAC rules needed by this operand:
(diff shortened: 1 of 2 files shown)


omeryahud:
Previously, we applied the `deprecated` label explicitly on deprecated templates, this change means that a template that is removed from a future release (does not exist in the released bundle) will be dynamically deprecated, is that the intended flow?
@kwiesmueller @rnetser @ksimon1 

Personally, I think that's fine.

kwiesmueller:
@omeryahud I would say this is intended. Tbh. I'm not sure how that problem surfaced just now as it must have been broken for a while now.
As long as we test that we don't deprecate templates not managed by SSP, this should be fine.

omeryahud:
> @omeryahud I would say this is intended. Tbh. I'm not sure how that problem surfaced just now as it must have been broken for a while now.
> As long as we test that we don't deprecate templates not managed by SSP, this should be fine.

It probably came up now because this is the first time the new operator upgrades itself

PR #189, opened 2021-06-29
add eventually command to some tests

ksimon1:
add eventually command to some tests.
These tests failed on BM due to bad timing of components creation

**Release note**:
```
NONE
```

### DIFF
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -89,23 +89,26 @@ var _ = Describe("Common templates", func() {
 	})
 
 	Context("resource creation", func() {
-		table.DescribeTable("created cluster resource", func(res *testResource) {
-			resource := res.NewResource()
-			err := apiClient.Get(ctx, res.GetKey(), resource)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(hasOwnerAnnotations(resource.GetAnnotations())).To(BeTrue())
+		table.DescribeTable("created cluster resource", func(res *testResource, name string) {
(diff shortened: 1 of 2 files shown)


ksimon1:
@akrejcir can you please check this PR?

akrejcir:
I'm surprised that the `Eventually` helps. It should not be needed, because in `BeforeEach` on line 88, there is `waitUntilDeployed()` call, that should wait until SSP operator deploys everything.

Do you know why `waitUntilDeployed()` is not enough?

ksimon1:
there is not problem, that something is not deployed. There is a problem, that sometimes, during update / patch, the api server throws errors e.g. 
```Message: "Operation cannot be fulfilled on clusterroles.rbac.authorization.k8s.io \"template:view\": the object has been modified; please apply your changes to the latest version and try again",
```

akrejcir:
> ```
> Message: "Operation cannot be fulfilled on clusterroles.rbac.authorization.k8s.io \"template:view\": the object has been modified; please apply your changes to the latest version and try again",
> ```

That error is usual for update or patch operation, but in this PR, the `Eventually` blocks are around `Get()` methods.
Does the get method return an error too? Maybe some timeout?

ksimon1:
> Does the get method return an error too? Maybe some timeout?
yes, there were problems with timing of tests which `eventually` command solved.

akrejcir:
Ok, Looks good.

/lgtm

ksimon1:
closing in favor of #198

PR #191, opened 2021-07-05
[WIP] implement dataimportcron api

kwiesmueller:
First implementation of https://github.com/kubevirt/community/blob/master/design-proposals/golden-image-delivery-and-update-pipeline.md

Depends on https://github.com/kubevirt/containerized-data-importer/pull/1808 which currently provides the new CDI API since 1.36.
The functional implementation is incomplete both on CDI and SSP as of this PR.

**Release note**:
```release-note
- Add the DataImportCron API to the SSP CustomResource for specifying importers managed by the SSP Operator.
```

/cc @nunnatsa @davidvossel @rnetser


internal/operands/data-import-crons/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (c *dataImportCrons) Reconcile(request *common.Request) ([]common.ResourceStatus, error) {
+	return common.CollectResourceStatus(request, reconcileDataImportCronsFuncs(request)...)
+}
+
+func (c *dataImportCrons) Cleanup(request *common.Request) error {
+	var existingDataImportCrons cdiv1beta1.DataImportCronList
+	requirement, err := labels.NewRequirement(common.AppKubernetesManagedByLabel, selection.Equals, []string{common.OperatorName})
+	if err != nil {
+		panic(err)
```
nunnatsa, line 58: panic? do we want to kill the operator for this error?

kwiesmueller: yes, this is an assertion. If we get an error here, there is a programming error that the user can not fix.

internal/operands/data-import-crons/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+		cron := &cronTemplates[i]
+		funcs = append(funcs, func(request *common.Request) (common.ResourceStatus, error) {
+			return common.CreateOrUpdate(request).
+				NamespacedResource(cron).
+				WithAppLabels(operandName, operandComponent).
+				UpdateFunc(func(newRes, foundRes client.Object) {
+					newDataImportCron := newRes.(*cdiv1beta1.DataImportCron)
+					foundDataImportCron := foundRes.(*cdiv1beta1.DataImportCron)
+					foundDataImportCron.Labels = newDataImportCron.Labels
+					foundDataImportCron.Annotations = newDataImportCron.Annotations
```
akrejcir, line 89: Labels and annotations do not need to be updated here in `UpdateFunc`, unless we specifically want to prevent users from adding custom labels and annotations to the resource.

Labels and annotations are reconciled here: https://github.com/kubevirt/ssp-operator/blob/master/internal/common/resource.go#L137

internal/operands/data-import-crons/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+		err := request.Client.Delete(request.Context, &obj)
+		if err != nil && !errors.IsNotFound(err) {
+			request.Logger.Error(err, fmt.Sprintf("Error deleting \"%s\": %s", obj.GetName(), err))
+			return err
+		}
+	}
+	return nil
+}
+
+func reconcileDataImportCronsFuncs(request *common.Request) []common.ReconcileFunc {
```
akrejcir, line 76: Have you thought about how to handle the case, when user removes some `DataImportCronTemplates` from the `SSP.spec`? Do we want them to be removed from the cluster?

internal/operands/data-import-crons/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+	funcs := make([]common.ReconcileFunc, 0, len(request.Instance.Spec.DataImportCronTemplates))
+	cronTemplates := request.Instance.Spec.DataImportCronTemplates
+	for i := range cronTemplates {
+		cron := &cronTemplates[i]
+		funcs = append(funcs, func(request *common.Request) (common.ResourceStatus, error) {
+			return common.CreateOrUpdate(request).
+				NamespacedResource(cron).
```
akrejcir, line 83: In what namespace do we want to create the `DataImportCron`s? In the design proposal, I only found the namespace mentioned in an example: https://github.com/kubevirt/community/blob/master/design-proposals/golden-image-delivery-and-update-pipeline.md#dataimportcron-example

If any namespace can be specified by user, then in the code the `DataImportCron` should be a cluster resource, even if in reality it is not.

In the SSP code the difference between cluster and namespaced types is that, namespaced types use owner references, and cluster types use owner annotations.
See: https://github.com/kubevirt/ssp-operator/blob/master/internal/common/resource.go#L158

The owner references only work for objects in the same namespace as the SSP CR.

I know it is confusing, and we should change the names, or unify the owner method in a future PR.

akrejcir (changes requested):

config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -22,60 +22,231 @@ spec:
         description: SSP is the Schema for the ssps API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
```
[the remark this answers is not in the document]
kwiesmueller: Will do.

internal/common/labels.go
```diff
@@ -23,6 +23,8 @@ const (
 	AppComponentMonitoring AppComponent = "monitoring"
 	AppComponentSchedule   AppComponent = "schedule"
 	AppComponentTemplating AppComponent = "templating"
+
+	OperatorName = "ssp-operator"
```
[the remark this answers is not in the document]
kwiesmueller: I'm using it in the dataImportCron to check the label.

internal/operands/data-import-crons/reconcile.go
kwiesmueller: Yes, I got that change locally already.
Mostly the same logic as above, I'm getting all crons with the applabel and the ones not part of the request get removed.

internal/operands/data-import-crons/reconcile.go
kwiesmueller: Not sure about the namespace question right now, but thanks for the info on the method to use.

internal/operands/data-import-crons/reconcile.go
kwiesmueller: We want to prevent them from updating labels and annotations as they can and should do that on the SSP CR instead.

internal/operands/data-import-crons/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (c *dataImportCrons) Reconcile(request *common.Request) ([]common.ResourceStatus, error) {
+	return common.CollectResourceStatus(request, reconcileDataImportCronsFuncs(request)...)
+}
+
+func (c *dataImportCrons) Cleanup(request *common.Request) error {
+	var existingDataImportCrons cdiv1beta1.DataImportCronList
+	requirement, err := labels.NewRequirement(common.AppKubernetesManagedByLabel, selection.Equals, []string{common.OperatorName})
```
ksimon1, line 56: Can you please move this to something like `func getLabelRequirement`? Because I think sonarcloud will complain, that this is too big function.

internal/operands/data-import-crons/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+		panic(err)
+	}
+	if err := request.Client.List(request.Context, &existingDataImportCrons, &client.ListOptions{
+		LabelSelector: labels.NewSelector().Add(*requirement),
+	}); err != nil {
+		request.Logger.Error(err, fmt.Sprintf("Error listing resources for deletion: %s", err))
+		return err
+	}
+	for _, obj := range existingDataImportCrons.Items {
+		err := request.Client.Delete(request.Context, &obj)
```
davidvossel, line 67: is there a reason why the cron's need to be deleted here? I thought SSP would put an owner reference on the cron's causing them to automatically get cleaned up when SSP is deleted.

kwiesmueller: As far as I understand it's because the ownerRefs don't work across namespaces so if the SSP CR gets removed it would only be able to remove stuff in its own namespace.

kwiesmueller:
Adding a
/hold
to make sure all the TODOs get resolved that make this a DO NOT MERGE right now.
Some of them being:
- [ ] update with actual common-templates using datasources
- [ ] remove dependency pin onto WIP kubevirt PR for CDI update

kwiesmueller:
I ran a manual test on a working OpenShift with this current state and SSP works!
The modified templates I use cause DataSources to be created accordingly and importcrons get properly reconciled as well.
Time to get some actual functionality.

internal/operands/data-import-crons/reconcile.go
davidvossel: ah, intresting. yup makes sense then.

PR #192, opened 2021-07-08
refactor reconcileOlderTemplates function

ksimon1:
refactor reconcileOlderTemplates function to reduce its cognitive cmplexity 

**Release note**:
```
NONE
```

### DIFF
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -174,27 +174,25 @@ func reconcileEditRole(request *common.Request) (common.ResourceStatus, error) {
 		}).
 		Reconcile()
 }
+func getOldTemplatesLabelSelector() labels.Selector {
+	baseRequirement, err := labels.NewRequirement(TemplateTypeLabel, selection.Equals, []string{"base"})
+	if err != nil {
+		panic(fmt.Sprintf("Failed creating label selector for '%s=%s'", TemplateTypeLabel, "base"))
+	}
+
+	// Only fetching older templates  to prevent duplication of API calls
+	versionRequirement, err := labels.NewRequirement(TemplateVersionLabel, selection.NotEquals, []string{Version})
+	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/operands/common-templates/reconcile.go
```diff
@@ -211,16 +209,13 @@ func reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, e
...
-			template.Annotations = make(map[string]string)
-		}
-		template.Annotations[TemplateDeprecatedAnnotation] = "true"
 		funcs = append(funcs, func(*common.Request) (common.ResourceStatus, error) {
 			return common.CreateOrUpdate(request).
 				ClusterResource(template).
 				WithAppLabels(operandName, operandComponent).
 				UpdateFunc(func(_, foundRes client.Object) {
 					foundTemplate := foundRes.(*templatev1.Template)
+					foundTemplate.Annotations[TemplateDeprecatedAnnotation] = "true"
```
kwiesmueller, line 218: Will this be safe enough? Are we sure that our older templates always have annotations already?

ksimon1: This function will tag as deprecated all templates which have different version, then is current version. The old templates don't have to have this label.

akrejcir: I think Kevin means, that this could panic, if `foundTemplate.Annotations == nil`. In normal situation this should not happen, but user could remove them.

ksimon1: if user removes annotations, it will not panic, because before operator calls that `UpdateFunc` (which adds the deprecated annotation), operator framework creates ownerAnnotations (https://github.com/kubevirt/ssp-operator/blob/master/internal/common/resource.go#L123), which creates new `map[string]string{}` (https://github.com/kubevirt/ssp-operator/blob/master/vendor/github.com/operator-framework/operator-lib/handler/enqueue_annotation.go#L170)

akrejcir: Ok, I missed that.

PR #193, opened 2021-07-08
switch base image from golang-1.15 to ubi8/ubi-minimal

ksimon1:
switch base image from golang-1.15 to ubi8/ubi-minimal
Due to pull rate limit from docker hub our tests constantly fails.
This PR switches main docker image to ubi8 image.

@akrejcir @kwiesmueller @omeryahud @vatsalparekh @dominikholler WDYT about this change?

**Release note**:

```
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 # Build the manager binary
-FROM golang:1.15 as builder
+FROM registry.access.redhat.com/ubi8/ubi-minimal as builder
+
+RUN microdnf install -y make golang-1.15.* which && microdnf clean all
 
 # Consume required variables so we can work with make
 ARG IMG_REPOSITORY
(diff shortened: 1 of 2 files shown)


akrejcir:
A good idea.

/lgtm

kwiesmueller:
Can this image be accessed by everyone?
We should not have anything upstream that is not fully open and accessible.
If we want to use it for CI under the hood, we should replace it in OpenShift CI.

ksimon1:
Ubi images are meant to be "public" https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image. And we already using them in ssp operator https://github.com/kubevirt/ssp-operator/blob/master/Dockerfile#L33

kwiesmueller:
/lgtm
/approve
Thanks!

PR #194, opened 2021-07-09
Fix VERSION variable in Makefile, fix Dockerfile, update CSV

akrejcir:
The `VERSION` variable in `Makefile` should be equal to the next release version in the sequence.

A file was not copied to the builder image in `Dockerfile`. It produces an error message, that is currently ignored, but when the operator-sdk will be updated, it will cause the build to fail.

Regenerated CSV in data/olm-catalog.

**Release note**:
```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 # Current Operator version
-VERSION ?= 0.1.3
+# The value here should be equal to the next version:
+# - next minor version on main branch
+# - next patch version on release branches
+VERSION ?= 0.12.0
+
 # Default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
 #operator-sdk version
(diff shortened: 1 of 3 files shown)


data/olm-catalog/ssp-operator.clusterserviceversion.yaml
```diff
@@ -355,11 +355,12 @@ spec:
                 env:
                 - name: KVM_INFO_IMAGE
                 - name: VALIDATOR_IMAGE
+                  value: quay.io/kubevirt/kubevirt-template-validator:latest
```
ksimon1, line 358: Is this value required here?

akrejcir: I'm not sure if it needs to be stored, this file is generated by the build process. We can probably add it to `.gitignore`.
I can investigate it in another PR.

Makefile
```diff
@@ -1,5 +1,9 @@
 # Current Operator version
-VERSION ?= 0.1.3
+# The value here should be equal to the next version:
+# - next minor version on main branch
+# - next patch version on release branches
+VERSION ?= 0.12.0
```
ksimon1, line 5: I would vote for setting this variable from git tag, but it can be done in next PR.
/lgtm

akrejcir: Good idea.

PR #197, opened 2021-07-13
Fix flaky tests

ksimon1:
Add eventually command to edit permission tests
Add check which controls if at least one validator pod is running

edit permissions tests are failing on baremetal. This PR
add eventually command, which will try to run test multiple time

Some validator tests are failing on baremetal due to not validating
rules. One of the possible root cause can be, that the validator is not
running. This PR adds check which controls if at least one validator pod
is running before test.

**Release note**:
```
NONE
```

### DIFF
--- a/tests/tests_common_test.go
+++ b/tests/tests_common_test.go
@@ -214,15 +214,15 @@ func getResourceKey(obj client.Object) client.ObjectKey {
 }
 
 func expectUserCan(sars *authv1.SubjectAccessReviewSpec) {
-	sar, err := coreClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, &authv1.SubjectAccessReview{
-		Spec: *sars,
-	}, metav1.CreateOptions{})
-	ExpectWithOffset(1, err).ToNot(HaveOccurred())
-	ExpectWithOffset(1, sar.Status.Allowed).To(BeTrue(),
-		fmt.Sprintf("user [%s] with groups %v cannot [%s] resource: [%s], subresource: [%s], name: [%s] in group [%s/%s] in namespace [%s]",
-			sars.User, sars.Groups, sars.ResourceAttributes.Verb, sars.ResourceAttributes.Resource,
-			sars.ResourceAttributes.Subresource, sars.ResourceAttributes.Name, sars.ResourceAttributes.Group,
-			sars.ResourceAttributes.Version, sars.ResourceAttributes.Namespace))
+	Eventually(func() bool {
+		sar, err := coreClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, &authv1.SubjectAccessReview{
+			Spec: *sars,
+		}, metav1.CreateOptions{})
+		return err == nil && sar.Status.Allowed
+	}, 15*time.Second, 2*time.Second).Should(BeTrue(), fmt.Sprintf("user [%s] with groups %v cannot [%s] resource: [%s], subresource: [%s], name: [%s] in group [%s/%s] in namespace [%s]",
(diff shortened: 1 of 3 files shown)


tests/tests_common_test.go
```diff
@@ -214,15 +214,15 @@ func getResourceKey(obj client.Object) client.ObjectKey {
...
-		fmt.Sprintf("user [%s] with groups %v cannot [%s] resource: [%s], subresource: [%s], name: [%s] in group [%s/%s] in namespace [%s]",
-			sars.User, sars.Groups, sars.ResourceAttributes.Verb, sars.ResourceAttributes.Resource,
-			sars.ResourceAttributes.Subresource, sars.ResourceAttributes.Name, sars.ResourceAttributes.Group,
-			sars.ResourceAttributes.Version, sars.ResourceAttributes.Namespace))
+	Eventually(func() bool {
+		sar, err := coreClient.AuthorizationV1().SubjectAccessReviews().Create(ctx, &authv1.SubjectAccessReview{
+			Spec: *sars,
+		}, metav1.CreateOptions{})
+		return err == nil && sar.Status.Allowed
+	}, 15*time.Second, 2*time.Second).Should(BeTrue(), fmt.Sprintf("user [%s] with groups %v cannot [%s] resource: [%s], subresource: [%s], name: [%s] in group [%s/%s] in namespace [%s]",
```
akrejcir, line 222: If the 15 seconds timeout is randomly chosen, can you use `shortTimout` instead?

tests/tests_suite_test.go
```diff
@@ -364,6 +366,18 @@ func waitUntilDeployed() {
 	deploymentTimedOut = false
 }
 
+func waitUntilTemplateValidatorIsRunning() {
```
akrejcir, line 369: Can you move this function to `validator_test.go`, because it is only used there.

tests/tests_suite_test.go
```diff
@@ -364,6 +366,18 @@ func waitUntilDeployed() {
...
+	key := client.ObjectKey{Name: validator.VirtTemplateValidator, Namespace: strategy.GetNamespace()}
```
akrejcir, line 370: I think it would be clearer if the `Name` is `validator.DeploymentName`. Then, if we choose to change the name in the future, the test would be correct.

tests/tests_suite_test.go
```diff
@@ -364,6 +366,18 @@ func waitUntilDeployed() {
...
+	foundValidator := new(apps.Deployment)
+	Expect(apiClient.Get(ctx, key, foundValidator)).ToNot(HaveOccurred())
+
+	EventuallyWithOffset(1, func() bool {
+		Expect(apiClient.Get(ctx, key, foundValidator)).ToNot(HaveOccurred())
+
+		return foundValidator.Status.AvailableReplicas > 1
+	}, timeout, time.Second).Should(BeTrue())
```
akrejcir, line 378: This function is called in `BeforeEach` for a lot of tests. In the worst case, this `EventuallyWithOffset()` will wait for 10 minutes for every test, which would take more than 9 hours.

Can you add a short-circuit logic similar to the one in `waitUntilDeployed()` function?

akrejcir (changes requested):

akrejcir:
Strange, that this PR is needed. The SSP operator should not be in `PhaseDeployed` if the template validator deployment has 0 `AvailableReplicas`. See: https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/template-validator/reconcile.go#L155

I think this needs more investigation.

akrejcir:
I have created another PR that should stabilize the tests, when timing is not good:
https://github.com/kubevirt/ssp-operator/pull/198

ksimon1:
closing in favor of https://github.com/kubevirt/ssp-operator/pull/198

PR #198, opened 2021-07-16
Stabilize functional tests

akrejcir:

- Use `GenerateName` to create unique temporary resources for tests. This helps with timing issues, when a temporary resource from one test was not yet deleted and is used by a different test.

- Template validator tests wait until new templates propagate to template validator.

- Make test failures more verbose

**Release note**:
```release-note
None
```


ksimon1 (commented):
/lgtm

PR #208, opened 2021-08-02
Protect old style templates from deletion

akrejcir:
Adds a webhook to protect old style templates from deletion. If a VM uses the template and it is deleted, the VM would not be able to start.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1929165

**Notes**
This PR adds a `cache.Reflector` to the validator that watches all VMs in the cluster. It uses an efficient implementation of the `cache.Store` interface to only store relevant VMs, and only useful data from a VM.

**Release note**:
```release-note
None
```


internal/operands/common-templates/constants.go
```diff
@@ -11,6 +11,8 @@ const (
 	TemplateWorkloadLabelPrefix  = "workload.template.kubevirt.io/"
 	TemplateDeprecatedAnnotation = "template.kubevirt.io/deprecated"
 
+	TemplateTypeLabelBaseValue = "base"
```
ksimon1, line 14: can you please update these places https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/common-templates/reconcile.go#L178 and all occurances here https://github.com/kubevirt/ssp-operator/blob/157b424d10a5c81813ce1bbcb66e326c42b0715b/internal/operands/common-templates/reconcile_test.go with this variable?

internal/template-validator/webhooks/hook.go
```diff
@@ -19,53 +19,147 @@
...
+	}
+
+	// Old versions of common templates had annotation validation
+	_, ok = template.Annotations[annotationValidationKey]
+	if !ok {
+		return ToAdmissionResponseOK()
+	}
+
+	// Old template cannot be removed if a VM uses it.
+	// We cannot filter by labels, because a VM may refer to the template by annotation.
```
ksimon1, line 121: Shouldn't have all vms created from common template special labels, which points them to template? From CT release v0.6.0 should have all templates `vm.kubevirt.io/template` label.

internal/template-validator/webhooks/utils.go
```diff
@@ -69,29 +70,35 @@ func ToAdmissionResponse(causes []metav1.StatusCause) *admissionv1.AdmissionResp
...
-	var err error
-	raw := ar.Request.Object.Raw
-	newVM := kubevirt.VirtualMachine{}
-
-	err = json.Unmarshal(raw, &newVM)
+	newVM := &kubevirt.VirtualMachine{}
+	err := json.Unmarshal(ar.Request.Object.Raw, newVM)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
```
ksimon1, line 81: you can use directly
`return vm, err`

internal/template-validator/webhooks/utils.go
```diff
@@ -69,29 +70,35 @@ func ToAdmissionResponse(causes []metav1.StatusCause) *admissionv1.AdmissionResp
...
-		err = json.Unmarshal(raw, &oldVM)
-		if err != nil {
-			return nil, nil, err
-		}
-		return &newVM, &oldVM, nil
+	return newVM, nil
+}
+
+func GetAdmissionReviewTemplate(ar *admissionv1.AdmissionReview) (*templatev1.Template, error) {
+	const resourceName = "templates"
```
ksimon1, line 88: can you please export this variable as global constant? You are using it 4 times in this PR.

internal/template-validator/webhooks/utils.go
```diff
@@ -69,29 +70,35 @@ func ToAdmissionResponse(causes []metav1.StatusCause) *admissionv1.AdmissionResp
...
+	if ar.Request.Operation == admissionv1.Delete {
+		obj = &ar.Request.OldObject
+	}
+
+	template := &templatev1.Template{}
+	err := json.Unmarshal(obj.Raw, template)
+	if err != nil {
+		return nil, err
+	}
+	return template, nil
```
ksimon1, line 103: you can use directly
`return template, err`

validator.Dockerfile
```diff
@@ -15,6 +15,9 @@ COPY go.sum go.sum
 COPY vendor/ vendor/
 
 # Copy the go source
+COPY main.go main.go
```
ksimon1, line 18: Do you need these files in validator image?

internal/operands/common-templates/constants.go
akrejcir: Ok

internal/template-validator/webhooks/hook.go
akrejcir: Yes, but I was not sure how to handle VMs from older templates.
I will change this part of the code. Calling List API in a webhook is probably too slow and may timeout if there are a lot of VMs.

internal/template-validator/webhooks/utils.go
akrejcir: `newVM` is not `nil` in case of error. But maybe it does not matter.

internal/template-validator/webhooks/utils.go
akrejcir: Ok

internal/template-validator/webhooks/utils.go
akrejcir: Again, `template` is not `nil` in case of error.

validator.Dockerfile
akrejcir: I copied the whole source code to the builder container, because the validator code imports some parts of it. `main.go` is not needed, I included it for completeness.

These files will not be in the validator image, they are only used in the builder image.

internal/template-validator/webhooks/utils.go
akrejcir: On second thought, I would like to leave it as is. I'm not sure in which file to put the global constant, and I don't want to unnecessarily mix validator and operator code.

internal/template-validator/webhooks/hook.go
ksimon1: The older vms have `vm.cnv.io/template: <template name>` label, but you are right, that currently you can't put `OR` between two labels.
To not fetch too many results, you can use Limit variable in ListOptions and then update VMs in batches.

internal/template-validator/webhooks/utils.go
ksimon1: I think it doesn't matter, because if err != nil then it will fail in fucntion which called the GetAdmissionReviewVM function

internal/template-validator/webhooks/utils.go
ksimon1: Fair point

internal/template-validator/webhooks/hook.go
akrejcir: I will post a different solution soon. We can use a lightweight controller, to watch all VMs, but only cache VMs that are relevant, so it will be memory efficient. Then this validating webhook can use the cache.

internal/template-validator/webhooks/utils.go
kwiesmueller: Yeah, I'd only return a variable if it's content is supposed to be of value in the return.
So do things like `return template, err` if the template value still might be used afterwards.
Only return one of them if only one value is supposed to be used.
Doesn't matter if the value should be nil anyways, it's just a nice convention to make sure the consumer never gets a value they should not use and the next developer knows that it is intentional.

internal/template-validator/virtinformers/vmcache.go
```diff
@@ -0,0 +1,255 @@
...
+		v.vmsForTemplate.Add(newVal.Template, newVal.Vm)
+		return nil
+	}
+
+	if reflect.DeepEqual(oldVal, newVal) {
+		return nil
+	}
+
+	v.store[key] = *newVal
+	v.vmsForTemplate.Delete(oldVal.Template, oldVal.Vm)
```
kwiesmueller, line 152: Couldn't you add an update mechanism to `templateMap` instead of deleting and recreating?

kwiesmueller (commented):
Looks good, just some small comment.
Could you also make sure to stick with the naming scheme and make shorts like `VM` and `VMI` consistently uppercase? I think we do that everywhere so far and I'd prefer to keep it consistent.

internal/template-validator/virtinformers/vmcache.go
akrejcir: I considered it, but I couldn't think of a more efficient way how to implement update, than simply delete followed by add.
Can you think of something?

akrejcir:
> Looks good, just some small comment.
> Could you also make sure to stick with the naming scheme and make shorts like `VM` and `VMI` consistently uppercase? I think we do that everywhere so far and I'd prefer to keep it consistent.

I prefer to use camelCase even for abbreviations like VM and CPU. There are not that many places that we use uppercase `VM` and `VMI`, only in a few functions in functional test code to create the VMs.

akrejcir:
Missing test IDs.

/hold

kwiesmueller:
The uppercase abbreviations is more a convention in Go and I think across both our project, kubevirt and kubernetes.
But I'm not gonna block on it, don't know what the others think.

internal/template-validator/virtinformers/vmcache.go
kwiesmueller: Wouldn't `Add` already act as an update? Or does it also have to remove?
If so then yes, I think delete and recreate is probably best.

akrejcir: The `templateMap` is a mapping from template name to a set of VM names. The update that I need is to change the template that is referenced by a VM, so update from `(template1, vm)` to `(template2, vm)`. But the template name is the key of the map, so the old key needs to be removed and the new key added.

I'm not sure I can explain it better that the code.

kwiesmueller: Right, that's how I understood it. Then disregard my question. Good this way.

akrejcir:
CI is not starting.
/hold

PR #214, opened 2021-08-17
Add total_rejected_vms matrics

borod108:
Add a metrics to template_validator to monitor the number of rejected Vms

```release-note
The metrics endpoint at the template_validtor pods will now have a total_rejected_vms metrics that counts the number of Vms it rejected.
```

### DIFF
--- a/tests/metrics_test_utils.go
+++ b/tests/metrics_test_utils.go
@@ -1,0 +1,45 @@
+package tests
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"regexp"
+	"strconv"
+
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+)
+
+var regexpForMetrics = map[string]*regexp.Regexp{
+	"total_rejected_vms": regexp.MustCompile(`total_rejected_vms ([0-9]+)`),
+}
+
+func intMetricValue(metricName string, metricsPort uint16, pod *v1.Pod) int {
+	conn, err := portForwarder.Connect(pod, metricsPort)
+	Expect(err).ToNot(HaveOccurred())
+	defer conn.Close()
+	client := &http.Client{
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+				return conn, nil
+			},
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+	resp, err := client.Get(fmt.Sprintf("https://localhost:%d/metrics", metricsPort))
+	Expect(err).ToNot(HaveOccurred(), "Can't get metrics from %s", pod.Name)
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
+	regex, ok := regexpForMetrics[metricName]
+	if !ok {
+		panic(fmt.Sprintf("metricName %s does not have a defined regexp string, please add one to the regexpForMetrics map", metricName))
+	}
+	valueOfMetric := regex.FindSubmatch(body)
+	intValue, err := strconv.Atoi(string(valueOfMetric[1]))
+	Expect(err).ToNot(HaveOccurred())
+	return intValue
+}
--- a/tests/validator_test.go
+++ b/tests/validator_test.go
@@ -4,13 +4,14 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"reflect"
+	"strings"
+	"time"
+
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
-	"reflect"
-	"strings"
-	"time"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
@@ -408,7 +409,6 @@ var _ = Describe("Template validator", func() {
 		It("[test_id:2960] Negative test - Create a VM with machine type violation", func() {
 			template = TemplateWithRules()
 			Expect(apiClient.Create(ctx, template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
-
 			// set value unfulfilling validation
 			vmi = addDomainResourcesToVMI(vmi, 2, "test", "128M")
 			vm = NewVirtualMachine(vmi)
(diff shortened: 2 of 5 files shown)


tests/metrics_test_utils.go
```diff
@@ -0,0 +1,59 @@
...
+	client := &http.Client{
+		Transport: &http.Transport{
+			DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+				return conn, nil
+			},
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+	resp, err := client.Get(fmt.Sprintf("https://localhost:%d/metrics", metricsPort))
+	Expect(err).To(BeNil(), "Can't get metrics from %s", pod.Name)
```
akrejcir, line 40: Instead of `.To(BeNil())`, you can use `.To(Succeed())` or `.ToNot(HaveOccured())`. These matchers expect an error type, and their message in case of failure is more clear.

I didn't know the second argument to `.To()` could be a format string, nice.

tests/metrics_test_utils.go
```diff
@@ -0,0 +1,59 @@
...
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
+	regexString, ok := regexpForMetricByName(metricName)
+	if !ok {
+		panic(fmt.Sprintf("metricName %s does not have a defined regexp string, please add one to the regexpForMetrics map", metricName))
+	}
+	regex := *regexp.MustCompile(regexString)
+	valueOfMetric := regex.FindSubmatch(body)
+	intValue, _ := strconv.Atoi(string(valueOfMetric[1]))
```
akrejcir, line 49: Please, don't ignore error here either.

tests/metrics_test_utils.go
```diff
@@ -0,0 +1,59 @@
...
+	return intValue
+}
+
+func regexpForMetricByName(metricName string) (string, bool) {
+	regexp, ok := regexpForMetrics[metricName]
+	if !ok {
```
akrejcir, line 55: This `if` is not needed. If an element is not found in `regexpForMetrics`, it returns `"", false`.

tests/validator_test.go
```diff
@@ -408,7 +409,21 @@ var _ = Describe("Template validator", func() {
...
 			Expect(apiClient.Create(ctx, template)).ToNot(HaveOccurred(), "Failed to create template: %s", template.Name)
+			// set value unfulfilling validation
+			vmi = addDomainResourcesToVMI(vmi, 2, "test", "128M")
+			vm = NewVirtualMachine(vmi)
+			vm.ObjectMeta.Annotations = map[string]string{
+				TemplateNameAnnotation:      template.Name,
+				TemplateNamespaceAnnotation: template.Namespace,
+			}
+			eventuallyFailToCreateVm(vm)
+			Expect(errors.IsInvalid(apiClient.Create(ctx, vm))).To(BeTrue(), "Should match error type because of unfulfilled validations")
```
akrejcir, line 420: This `Expect()` is not needed. The same check is already done in `eventuallyFailToCreateVm(vm)`

tests/validator_test.go
```diff
@@ -408,7 +409,21 @@ var _ = Describe("Template validator", func() {
...
+		})
 
+		It("[test_id:7060]vm rejected metrics increses by one when vm is rejected", func() {
+			defer expectTotalRejectedCountToIncBy(1)()
```
akrejcir, line 424: I think it is problematic to call `Expect()` in a function that is executed using `defer`.
If this test case fails, for example on line 426, the `defer` function will still be executed.

Can you inline this function like:
```go
It("[test_id:7060]vm rejected metrics increses by one when vm is rejected", func() {
    rejectedBefore := totalRejectedVmsMetricsValue()

    // ... the rest of test case

    Expect(totalRejectedVmsMetricsValue() - rejectedBefore).To(Equal(1))
}
```

tests/validator_test.go
```diff
@@ -775,6 +791,21 @@ func eventuallyFailToCreateVm(vm *kubevirtv1.VirtualMachine) bool {
...
 
+func expectTotalRejectedCountToIncBy(diff int) func() {
+	rejectedBefore := totalRejectedVmsMetricsValue()
+	return func() { Expect(totalRejectedVmsMetricsValue() - rejectedBefore).To(Equal(diff)) }
+}
+
+func totalRejectedVmsMetricsValue() (sum int) {
+	pods, err := GetRunningPodsByLabel(validator.VirtTemplateValidator, validator.KubevirtIo, strategy.GetNamespace())
+	Expect(err).ToNot(HaveOccurred(), "Could not find the validator pods")
+	Expect(len(pods.Items)).To(BeNumerically(">", 0))
```
akrejcir, line 802: You can use `Expect(pods.Items).ToNot(BeEmpty())`.

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/davecgh/go-spew v1.1.1
+	github.com/emicklei/go-restful v2.10.0+incompatible // indirect
```
akrejcir, line 8: This module doesn't need to be written here.
Can you run:
```
go mod vendor
go mod tidy
```

It should remove it.

akrejcir (changes requested):

go.mod
borod108: I run both before committing and run them again now, still there :/

tests/metrics_test_utils.go
```diff
@@ -0,0 +1,59 @@
...
+
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+var regexpForMetrics = map[string]string{
+	"total_rejected_vms": `total_rejected_vms ([0-9]+)`,
+}
```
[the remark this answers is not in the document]
borod108: good point

tests/metrics_test_utils.go
```diff
@@ -0,0 +1,59 @@
...
+
+func intMetricValue(metricName string, metricsPort uint16, pod *v1.Pod) int {
+	cfg, err := config.GetConfig()
+	Expect(err).ToNot(HaveOccurred())
+	coreClient, err := kubernetes.NewForConfig(cfg)
+	Expect(err).ToNot(HaveOccurred())
+
+	portForwarder := NewPortForwarder(cfg, coreClient.CoreV1().RESTClient())
+	conn, _ := portForwarder.Connect(pod, metricsPort)
```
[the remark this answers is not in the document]
borod108: yes

go.mod
akrejcir: Maybe you have different go version?
I have go version `go1.16.6`. And on my machine `go mod tidy` removes the line. 

In any case, the line is not needed and you can remove it. Also the change it `vendor/modules.txt` is not needed.

tests/metrics_test_utils.go
borod108: sure

tests/metrics_test_utils.go
borod108: yep! removed this useless function.

tests/tests_common_test.go
```diff
@@ -2,11 +2,12 @@ package tests
...
 import (
 	"fmt"
-	"github.com/onsi/gomega/format"
-	"github.com/onsi/gomega/types"
 	"reflect"
 	"time"
 
+	"github.com/onsi/gomega/format"
+	"github.com/onsi/gomega/types"
+
```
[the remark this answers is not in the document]
borod108: yep

tests/metrics_test_utils.go
borod108: done

tests/validator_test.go
borod108: hmm, that is what I had in the beginning and I found the deffer thing nicer - one line in the beginning that explains the essence of the test right away. is it really problematic if this deffer runs in any case? We will just get another failure message right? And it will be true...

tests/validator_test.go
borod108: this is git :/ I did not really change this code at all, it just got confused because of the similarity with the function I did add below... I think changing this is for a separate PR.

tests/validator_test.go
borod108: I can, and I will :)

go.mod
borod108: done

tests/validator_test.go
akrejcir: So the line was added by mistake during rebase.

Can you rebase this PR on the latest master and remove this line?

tests/validator_test.go
akrejcir: I'm not sure how golang handles `panic` inside `defer` function that is called while panicking.
And the function `totalRejectedVmsMetricsValue()` is not fast, it calls API. So it would be better to not call it if not necessary.

borod108: ok

tests/validator_test.go
borod108: you are right! done.

borod108:
@akrejcir thank you for the review, applied your feedback, please take a look :)

PR #225, opened 2021-08-26
Recording rule and alert for SSPHighRateRejectedVms

borod108:
Adding kubevirt_ssp_rejected_vms_total recording rule and
SSPHighRateRejectedVms alert to warn user in case many VMs
are being rejected.

```release-note
Added SSPHighRateRejectedVms Alert which will activate as a warning in case of high rate of rejected VMs
```

### DIFF
--- a/internal/operands/metrics/resources.go
+++ b/internal/operands/metrics/resources.go
@@ -42,7 +42,7 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 							For:   "5m",
 							Annotations: map[string]string{
 								"summary":     "All SSP operator pods are down.",
-								"runbook_url": "http://kubevirt.io/monitoring/runbooks/SSPOperatorDown",
+								"runbook_url": "https://kubevirt.io/monitoring/runbooks/SSPOperatorDown",
 							},
 							Labels: map[string]string{
 								"severity": "Critical",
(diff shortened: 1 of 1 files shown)


internal/operands/metrics/resources.go
```diff
@@ -54,12 +54,28 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
...
 								"severity": "Critical",
 							},
 						},
+						{
+							Record: "kubevirt_ssp_rejected_vms_total",
+							Expr:   intstr.FromString("sum(increase(total_rejected_vms{pod=~'virt-template-validator.*'}[1h]))"),
+						},
+						{
+							Alert: "SSPHighRateRejectedVms",
+							Expr:  intstr.FromString("kubevirt_ssp_rejected_vms_total > 5"),
```
akrejcir, line 69: Can you add an explanation why did you choose 6 rejections per hour to be "high rate"?

borod108: Nothing beyond - we talked about it with other people and it kinda seemed to make sense.
It is not important enough to be configurable - this is basically to alert the admin to someone who does not have a grasp of how template validation works.
If we find that this Alert is less useful to users in the future we can change it, but without actual user research we won't know what number makes sense.
I think this is ok to leave without an explanation for now, since the explanation "best guess" does not explain much...

borod108: It is also not really 6 per hour, since Prometheus works withe extrapolated rates, but it is not that important and explaining that will be much more confusing.
The runbook itself, when ready, will have a more detailed explanation on what it means and what to do about it.

akrejcir: Ok, thanks for clarification.

PR #226, opened 2021-08-26
Code change: Improve validator code

akrejcir:
This PR contains these code changes:
- Created `RuleType` type to emulate enum.
- Split `validation.Path` struct into 2 structs `Path` and `Results`
- `Rule` struct uses specific types instead of generic `string` and `interface{}`

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/template-validator/validation/eval_test.go
+++ b/internal/template-validator/validation/eval_test.go
@@ -15,13 +15,13 @@ var _ = Describe("Eval", func() {
 			rules := []Rule{
 				{
 					Name: "rule-1",
-					Rule: "integer",
+					Rule: IntegerRule,
 					// any legal path is fine
 					Path:    "jsonpath::.spec.domain.cpu.cores",
 					Message: "testing",
 				}, {
 					Name: "rule-1",
-					Rule: "string",
+					Rule: StringRule,
 					// any legal path is fine
 					Path:    "jsonpath::.spec.domain.cpu.cores",
 					Message: "testing",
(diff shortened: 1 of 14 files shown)


internal/template-validator/validation/eval_test.go
```diff
@@ -226,18 +222,18 @@ var _ = Describe("Eval", func() {
...
 					Name: "rule-2",
-					Rule: "integer",
-					Min:  6,
-					Max:  8,
+					Rule: IntegerRule,
+					Min:  &path.IntOrPath{Int: 6},
+					Max:  &path.IntOrPath{Int: 8},
 					// any legal path is fine
-					Path:    "jsonpath::.spec.domain.cpu.cores",
+					Path:    *path.NewOrPanic("jsonpath::.spec.domain.cpu.cores"),
```
ksimon1, line 236: Just asking, when the Path is not correct, should't `res.Status[x].Error` be not nil?

akrejcir: Yes, but in this test case the Path is correct. It parses correctly, and points to an existing field in the VM structure.
The `vmCirros` VM used in this test has `.spec.domain.cpu = nil`, but that is not an error, because the validator uses a "reference" VM, that contains default values for nil fields. The `cores` value is taken from the reference VM, and the rule is evaluated.

ksimon1: ah yes, sorry. I missed there are no " around `*path.NewOrPanic("jsonpath::.spec.domain.cpu.cores")` :)

PR #228, opened 2021-08-27
implement dataimportcron api

kwiesmueller:
This PR contains the API changes needed for https://github.com/kubevirt/ssp-operator/pull/191.
It is meant to merge before the actual implementation to allow early consumption by the HCO team.

The API changes in this PR are not meant to be released before v0.13 as they are not functional yet.

**Release note**:

```release-note
Implement basic non-functional dataimportcron API on the SSP CR
```

/cc @nunnatsa


api/v1beta1/constants.go
```diff
@@ -0,0 +1,5 @@
+package v1beta1
+
+const (
+	GoldenImagesNSname = "kubevirt-os-images"
```
akrejcir, line 4: Is it a good idea to export a hardcoded namespaces as part of API?
Is there a reason why golden images cannot be in a different namespace?

api/v1beta1/ssp_webhook.go
```diff
@@ -159,3 +167,16 @@ func validateOperandPlacement(namespace string, placement *api.NodePlacement) er
 
 	return clt.Create(context.TODO(), deployment, &client.CreateOptions{DryRun: []string{metav1.DryRunAll}})
 }
+
+// TODO: also validate DataImportCronTemplates in general once CDI exposes its own validation
```
akrejcir, line 171: Maybe the validation could do a dry-run create API call, to test if `DataImportCron` is valid. But maybe it would time out if there are a lot of `DataImportCronTemplates` specified.

api/v1beta1/ssp_webhook.go
```diff
@@ -159,3 +167,16 @@ func validateOperandPlacement(namespace string, placement *api.NodePlacement) er
...
+func validateDataImportCronTemplates(r *SSP) error {
+	for _, cron := range r.Spec.DataImportCronTemplates {
+		if len(cron.Name) < 1 {
```
akrejcir, line 174: Please use `if cron.Name == ""`, because we don't care about the length.

api/v1beta1/ssp_webhook.go
```diff
@@ -159,3 +167,16 @@ func validateOperandPlacement(namespace string, placement *api.NodePlacement) er
...
+			return fmt.Errorf("missing name in DataImportCronTemplate")
+		}
+		if len(cron.Namespace) > 0 && cron.Namespace != GoldenImagesNSname {
```
akrejcir, line 177: This may be a good time to allow deploying golden images to any namespace, and not hard-code a constant namespace.
Like we do for VM templates.

api/v1beta1/ssp_types.go
```diff
@@ -57,6 +58,26 @@ type SSPSpec struct {
 
 	// NodeLabeller is configuration of the node-labeller operand
 	NodeLabeller NodeLabeller `json:"nodeLabeller,omitempty"`
+
+	// DataImportCronTemplates defines a list of DataImportCrons managed by the SSP
+	// Operator. This is intended for images used by CommonTemplates.
+	DataImportCronTemplates []DataImportCronTemplate `json:"dataImportCronTemplates,omitempty"`
```
akrejcir, line 64: Since this is related to common templates, can you move this field inside `CommonTemplates` struct?

api/v1beta1/ssp_webhook_test.go
```diff
@@ -138,8 +139,74 @@ var _ = Describe("SSP Validation", func() {
...
+			table.Entry("golden image namespace provided", GoldenImagesNSname, "test-name", false),
+			table.Entry("invalid namespace provided", "invalid-namespace", "test-name", true),
+		)
+
+	})
 })
 
+func checkExpectedError(err error, shouldFail bool) {
+	if shouldFail {
+		Expect(err).To(HaveOccurred())
```
akrejcir, line 204: You can use `ExpectWithOffset(1, err)` to make the error message point the calling function.

akrejcir:
Can you also rename this PR or #191, so they have different names?

api/v1beta1/constants.go
kwiesmueller: I'm not sure why that has been hardcoded in the past, but it is.
If we want to change that, I'd prefer to make that decision at another place and change it then.
Regarding the export, I probably wouldn't hardcode it in the first place, but that's what it is. Having it here in the API it's at least consumable by anything that needs to know the namespace.

api/v1beta1/ssp_webhook.go
kwiesmueller: CDI plans on exporting the actual validation code and as we already consume their API we can just use that once they do.

api/v1beta1/ssp_webhook.go
kwiesmueller: Right, old habit. Both are very common ways to check an empty string in Go. Will change.

api/v1beta1/ssp_webhook.go
kwiesmueller: I agree, but would prefer to have that as a separate decision and not as part of this API change.

kwiesmueller:
No CI again :-/

PR #238, opened 2021-09-14
Load common templates before creating the operand

akrejcir:
This patch loads common templates before starting the reconciler. This allows early exit, if the template bundle is incorrect.

Duplicate common templates files are removed. They were used only for unit tests and are not needed.

**Release note**:
```release-note
None
```


controllers/ssp_controller.go
```diff
@@ -596,31 +608,33 @@ func watchSspResource(bldr *ctrl.Builder) {
 	bldr.For(&ssp.SSP{}, builder.WithPredicates(pred))
 }
 
-func watchNamespacedResources(builder *ctrl.Builder) {
+func watchNamespacedResources(builder *ctrl.Builder, opers []operands.Operand) {
```
ksimon1, line 611: wouldn't be better to use better name than `opers`? e.g. sspOperands

internal/operands/common-templates/reconcile_test.go
```diff
@@ -75,6 +86,20 @@ var _ = Describe("Common-Templates operand", func() {
 		}
 	})
 
+	It("should correctly read templates", func() {
+		Expect(testTemplates).To(HaveLen(2))
+
+		templ1 := testTemplates[0]
```
ksimon1, line 92: are you sure, that templates will be allways loaded in the same order?

controllers/ssp_controller.go
akrejcir: yes, I will change it.

internal/operands/common-templates/reconcile_test.go
akrejcir: Yes, the templates are loaded one by one sequentially from the file and added to a slice.
See the `ReadTemplates()` function in `common-templates/reconcile.go`:
https://github.com/kubevirt/ssp-operator/pull/238/files#diff-f6d56152bde2c23e8384a420483683cbfc8b9d3152bd9c104914c737b8a99cf3R239

ksimon1: I meant that if they ever changed the functionality of how data is loaded in some future version of "k8s.io/apimachinery/pkg/util/yaml", the test would not work. Maybe it would be better to write a test
```
for _, template := range testTemplates {
  if templ1.Name == "centos8-server-medium" {
    Expect(templ1.Annotations).To(HaveKey("name.os.template.kubevirt.io/centos8"))
    Expect(templ1.Objects).To(HaveLen(1))
  }
  ....
}
```
It is just a small thing, so feel free to keep it as it is.

akrejcir: The order is defined in our code, not in the `k8s.io/apimachinery/pkg/util/yaml` library (see the link).
We use the `yaml` library to create a decoder and then load each template in a loop until an EOF is reached.
The decoder has to read the file sequentially, otherwise I think it would not make sense.

But this is just a unit tests, so if it fails, we will notice.

ksimon1:
@akrejcir can you please rebase this PR?

PR #239, opened 2021-09-16
Enable changing of namespace during run of ssp operator

ksimon1:
Enable changing of namespace during run of ssp operator
/cc @akrejcir @rnetser 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1958085

**Release note**:
```
Enable changing of namespace during run of ssp operator
```

### DIFF
--- a/api/v1beta1/ssp_webhook_test.go
+++ b/api/v1beta1/ssp_webhook_test.go
@@ -119,7 +119,7 @@ var _ = Describe("SSP Validation", func() {
 		})
 	})
 
-	It("should not allow update of commonTemplates.namespace", func() {
+	It("should allow update of commonTemplates.namespace", func() {
 		oldSsp := &SSP{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      "test-ssp",
@@ -136,8 +136,7 @@ var _ = Describe("SSP Validation", func() {
 		newSsp.Spec.CommonTemplates.Namespace = "new-ns"
 
 		err := newSsp.ValidateUpdate(oldSsp)
-		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(ContainSubstring("commonTemplates.namespace cannot be changed."))
+		Expect(err).ToNot(HaveOccurred())
 	})
 
 	Context("DataImportCronTemplates", func() {
--- a/tests/commonTemplates_test.go
+++ b/tests/commonTemplates_test.go
@@ -11,9 +11,11 @@ import (
 	templatev1 "github.com/openshift/api/template/v1"
 	authv1 "k8s.io/api/authorization/v1"
 	core "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ssp "kubevirt.io/ssp-operator/api/v1beta1"
+	sspv1beta1 "kubevirt.io/ssp-operator/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	cdiv1beta1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
@@ -208,6 +210,43 @@ var _ = Describe("Common templates", func() {
 	})
 
 	Context("resource change", func() {
+		Context("namespace change", func() {
+			var newTemplateNamespace string
+			var namespaceObj *v1.Namespace
+			BeforeEach(func() {
+				strategy.SkipSspUpdateTestsIfNeeded()
+
+				namespaceObj = &v1.Namespace{ObjectMeta: metav1.ObjectMeta{GenerateName: "ssp-operator-functests-new-templates"}}
+				err := apiClient.Create(ctx, namespaceObj)
+				Expect(err).To(BeNil(), "err should be nil")
+
+				newTemplateNamespace = namespaceObj.GetName()
+			})
+
+			AfterEach(func() {
+				strategy.RevertToOriginalSspCr()
+				err := apiClient.Delete(ctx, namespaceObj)
+				Expect(err).To(BeNil(), "err should be nil")
+			})
+
+			It("[test_id:6057] should update commonTemplates.namespace", func() {
+				foundSsp := getSsp()
+				Expect(foundSsp.Spec.CommonTemplates.Namespace).ToNot(Equal(newTemplateNamespace), "namespaces should not equal")
+
+				updateSsp(func(foundSsp *sspv1beta1.SSP) {
+					foundSsp.Spec.CommonTemplates.Namespace = newTemplateNamespace
+				})
+				waitUntilDeployed()
+
+				templates := &templatev1.TemplateList{}
+				Eventually(func() bool {
+					err := apiClient.List(ctx, templates, &client.ListOptions{
+						Namespace: newTemplateNamespace,
+					})
+					return len(templates.Items) > 0 && err == nil
+				}, shortTimeout).Should(BeTrue(), "templates should be in new namespace")
+			})
+		})
 		table.DescribeTable("should restore modified resource", expectRestoreAfterUpdate,
 			table.Entry("[test_id:5315]edit cluster role", &editClusterRole),
 			table.Entry("[test_id:5316]view role", &viewRole),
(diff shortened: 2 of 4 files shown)


tests/webhook_test.go
```diff
@@ -157,12 +158,21 @@ var _ = Describe("Validation webhook", func() {
 			strategy.RevertToOriginalSspCr()
 		})
 
-		It("[test_id:6057] should fail to update commonTemplates.namespace", func() {
-			originalNs := foundSsp.Spec.CommonTemplates.Namespace
-			foundSsp.Spec.CommonTemplates.Namespace = originalNs + "-updated"
+		It("[test_id:6057] should not fail to update commonTemplates.namespace", func() {
```
akrejcir, line 161: You could remove this test, because the webhook does not restrict the namespace anymore. 
Or if you want, you can add a new test to `tests/commonTemplates_test.go` to check if templates are deployed to different namespace when the `.spec.commonTemplates.namespace` is changed.

api/v1beta1/ssp_webhook_test.go
```diff
@@ -119,7 +119,7 @@ var _ = Describe("SSP Validation", func() {
 		})
 	})
 
-	It("should not allow update of commonTemplates.namespace", func() {
+	It("should allow update of commonTemplates.namespace", func() {
```
akrejcir, line 122: You can remove this test. The webhook does not restrict the namespace anymore.

akrejcir (changes requested):

api/v1beta1/ssp_webhook_test.go
ksimon1: I would prefer to keep it there, because it should check that webhook doesn't prevent to change namespace

tests/commonTemplates_test.go
```diff
@@ -208,6 +208,30 @@ var _ = Describe("Common templates", func() {
...
 	Context("resource change", func() {
+
+		Context("namespace change", func() {
+			AfterEach(func() {
+				strategy.RevertToOriginalSspCr()
+			})
+
+			It("[test_id:6057] should not fail to update commonTemplates.namespace", func() {
+				foundSsp := getSsp()
+				const testNamespace = "ssp-operator-functests"
```
akrejcir, line 219: Can you create a new namespace for this test? Would be better than polluting the testing namespace.
Also the tests can be run in a different namespace, so instead of a constant, you can use `strategy.GetNamespace()`.

tests/commonTemplates_test.go
```diff
@@ -208,6 +208,30 @@ var _ = Describe("Common templates", func() {
...
+				Expect(foundSsp.Spec.CommonTemplates.Namespace).ToNot(Equal(testNamespace), "namespaces should not equal")
+				foundSsp.Spec.CommonTemplates.Namespace = testNamespace
+
+				err := apiClient.Update(ctx, foundSsp)
```
akrejcir, line 223: This update may fail. It is more stable to call `updateSsp()` that retries updates:
```go
updateSsp(func(foundSsp *sspv1beta1.SSP) {
	foundSsp.Spec.CommonTemplates.Namespace = testNamespace
})
```

tests/commonTemplates_test.go
```diff
@@ -208,6 +208,30 @@ var _ = Describe("Common templates", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+
```
akrejcir, line 225: The test needs to wait until SSP operator is finished with reconciliation. You can call: `waitUntilDeployed()`

tests/commonTemplates_test.go
```diff
@@ -208,6 +208,30 @@ var _ = Describe("Common templates", func() {
 	})
 
 	Context("resource change", func() {
+
+		Context("namespace change", func() {
+			AfterEach(func() {
```
akrejcir, line 213: Please add:
```go
BeforeEach(func() {
	strategy.SkipSspUpdateTestsIfNeeded()
}
```

The test suite can be configured to skip any tests that change the SSP resource. This is useful if some other operator watches the SSP and it could revert any changes in the middle of the test.

akrejcir (changes requested):

ksimon1:
@akrejcir can you please review?

tests/commonTemplates_test.go
```diff
@@ -208,6 +209,34 @@ var _ = Describe("Common templates", func() {
...
+		Context("namespace change", func() {
+			BeforeEach(func() {
+				strategy.SkipSspUpdateTestsIfNeeded()
+			})
+
+			AfterEach(func() {
+				strategy.RevertToOriginalSspCr()
+			})
+
+			It("[test_id:6057] should not fail to update commonTemplates.namespace", func() {
```
akrejcir, line 221: You can just call the test: `should update commonTemplates.namespace`.

tests/tests_suite_test.go
```diff
@@ -84,6 +85,11 @@ func (s *newSspStrategy) Init() {
 		return apiClient.Create(ctx, namespaceObj)
 	}, timeout, time.Second).ShouldNot(HaveOccurred())
 
+	Eventually(func() error {
+		namespaceObj := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: s.GetNewTemplatesNamespace()}}
+		return apiClient.Create(ctx, namespaceObj)
+	}, timeout, time.Second).ShouldNot(HaveOccurred())
```
akrejcir, line 91: This namespace is only used in 1 test case. Can you move the code, so that this namespace is created and removed inside the test?

akrejcir (changes requested):
Just small comments.

tests/commonTemplates_test.go
```diff
@@ -208,6 +211,42 @@ var _ = Describe("Common templates", func() {
 	})
 
 	Context("resource change", func() {
+		Context("namespace change", func() {
+			BeforeEach(func() {
+				Eventually(func() error {
+					namespaceObj := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: strategy.GetNewTemplatesNamespace()}}
+					return apiClient.Create(ctx, namespaceObj)
```
akrejcir, line 218: Instead of  setting `Name: strategy.GetNewTemplateNamespace()`, can you set `GenerateName` to some constant? It will create a namespaces with a random suffix.

Then you can remove `GetNewTemplateNamespace()` from `strategy`. The methods in `strategy` are used to get values that are configurable from outside of the test script, using environment variables, and the new template namespace is not configurable.

And if you want, you can remove the `Eventually` block and create the namespace directly.

tests/commonTemplates_test.go
```diff
@@ -208,6 +211,42 @@ var _ = Describe("Common templates", func() {
...
+				}, timeout, 10*time.Second).ShouldNot(HaveOccurred())
+				strategy.SkipSspUpdateTestsIfNeeded()
```
akrejcir, line 220: This call should be at the top of `BeforeEach()`. When the test is skipped, the namespace does not need to be created.

tests/commonTemplates_test.go
```diff
@@ -208,6 +211,42 @@ var _ = Describe("Common templates", func() {
...
+			})
+
+			AfterEach(func() {
+				Eventually(func() error {
+					namespaceObj := &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: strategy.GetNewTemplatesNamespace()}}
+					return apiClient.Delete(ctx, namespaceObj)
+				}, timeout, time.Second).ShouldNot(HaveOccurred())
```
akrejcir, line 227: This is probably an unstable method to wait for deletion, because when the namespace does not exist, the `Delete()` method will return an error. But it will retry the deletion for 10 minutes, before timing out.

You can use something like:
```go
AfterEach(func() {
	err := apiClient.Delete(ctx, namespace)
	expectSuccessOrNotFound(err)
	waitForDeletion(client.ObjectKey{Name: namespace.Name}, &v1.Namespace{})

	strategy.RevertToOriginalSspCr()
}
```

akrejcir (changes requested):

tests/commonTemplates_test.go
akrejcir: Also, it may be better to call `strategy.RevertToOriginalSspCr()` before deleting the namespace, so the operator will not be confused.

akrejcir:
This looks good, thank you.
/lgtm

PR #240, opened 2021-09-16
Add total_restored_common_templates metrics

borod108:
Add a metric that increases when a common template is updated during
the reconcile step of the common_templates operand.

```release-note
Added metric named total_restored_common_templates that counts the number of common templates updated during reconcile.
```

### DIFF
--- a/internal/common/resource.go
+++ b/internal/common/resource.go
@@ -13,16 +13,21 @@ import (
 type StatusMessage = *string
 
 type ResourceStatus struct {
-	Resource     client.Object
 	Progressing  StatusMessage
 	NotAvailable StatusMessage
 	Degraded     StatusMessage
 }
 
-type ReconcileFunc = func(*Request) (ResourceStatus, error)
+type ReconcileResult struct {
+	Status          ResourceStatus
+	Resource        client.Object
+	OperationResult controllerutil.OperationResult
+}
 
-func CollectResourceStatus(request *Request, funcs ...ReconcileFunc) ([]ResourceStatus, error) {
-	res := make([]ResourceStatus, 0, len(funcs))
+type ReconcileFunc = func(*Request) (ReconcileResult, error)
+
+func CollectResourceStatus(request *Request, funcs ...ReconcileFunc) ([]ReconcileResult, error) {
+	res := make([]ReconcileResult, 0, len(funcs))
 	for _, f := range funcs {
 		status, err := f(request)
 		if err != nil {
@@ -43,7 +48,7 @@ type ReconcileBuilder interface {
 	UpdateFunc(ResourceUpdateFunc) ReconcileBuilder
 	StatusFunc(ResourceStatusFunc) ReconcileBuilder
 
-	Reconcile() (ResourceStatus, error)
+	Reconcile() (ReconcileResult, error)
 }
 
 type reconcileBuilder struct {
@@ -90,7 +95,7 @@ func (r *reconcileBuilder) WithAppLabels(name string, component AppComponent) Re
 	return r
 }
 
-func (r *reconcileBuilder) Reconcile() (ResourceStatus, error) {
+func (r *reconcileBuilder) Reconcile() (ReconcileResult, error) {
 	if r.addLabels {
 		AddAppLabels(r.request.Instance, r.operandName, r.operandComponent, r.resource)
 	}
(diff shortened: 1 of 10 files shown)


internal/common/resource.go
```diff
@@ -19,10 +19,15 @@ type ResourceStatus struct {
 	Degraded     StatusMessage
 }
 
-type ReconcileFunc = func(*Request) (ResourceStatus, error)
+type ReconcileResult struct {
+	ResourceStatus
+	OperationResult controllerutil.OperationResult
+}
```
akrejcir, line 25: Can you also move `Resource` field from `ResourceStatus` struct to this struct?
I think it would make more sense. Then the status `ResourceStatus` would contain only status messages and all the other information will be in `ReconcileResult`.

And can you give name to the inlined `ResourceStatus` struct? For example: `status`.
Inlining a struct would be useful if we wanted to call methods defined on the inlined struct. But the `ResourceStatus` does not have any methods.

internal/common/resource.go
```diff
@@ -34,7 +39,7 @@ func CollectResourceStatus(request *Request, funcs ...ReconcileFunc) ([]Resource
 }
 
 type ResourceUpdateFunc = func(expected, found client.Object)
-type ResourceStatusFunc = func(resource client.Object) ResourceStatus
+type ResourceStatusFunc = func(resource client.Object) ReconcileResult
```
akrejcir, line 42: This function should still return `ResourceStatus`. It is used to create resource specific status, for example, on deployment waiting until its pods are ready.

internal/operands/common-templates/reconcile.go
```diff
@@ -108,7 +117,13 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ResourceS
 	funcs = append(funcs, oldTemplateFuncs...)
 	funcs = append(funcs, reconcileTemplatesFuncs(request)...)
 
-	return common.CollectResourceStatus(request, funcs...)
+	results, err := common.CollectResourceStatus(request, funcs...)
+	for _, r := range results {
```
akrejcir, line 121: The `results` contains results not only from created templates, but also from old templates (leftovers from old ssp operator version), and non-template resources like roles and role bindings. See the definition of `funcs` on line 77.

You could split the `funcs` list and call `CollectResourceStatus()` on parts of it, and then increase the metric only for template resources:

```go
funcs = append(funcs, oldTemplateFuncs...)
results, err := common.CollectResourceStatus(request, funcs...)
if err != nil {
	return nil, err
}

resultsTemplates, err := common.CollectResourceStatus(request, reconcileTemplatesFuncs(request)...)
if err != nil {
	return nil, err
}
for _, r := range resultsTemplates {
	if r.OperationResult == controllerutil.OperationResultUpdated {
		CommonTemplatesRestored.Inc()
	}
}
return append(results, resultsTemplates...), nil
```

akrejcir (changes requested):

internal/operands/common-templates/reconcile.go
borod108: yep!

internal/common/resource.go
borod108: yep combined with your previous comment this works great.

internal/common/resource.go
borod108: Right this is far better.

internal/operands/common-templates/reconcile.go
```diff
@@ -106,9 +115,21 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ResourceS
...
+	reconcileTemplatesResults, err := common.CollectResourceStatus(request, reconcileTemplatesFuncs(request)...)
+	if err != nil {
+		return nil, err
+	}
+	for _, r := range reconcileTemplatesResults {
+		if r.OperationResult == controllerutil.OperationResultUpdated {
+			CommonTemplatesRestored.Inc()
+		}
+	}
+	return append(results, reconcileTemplatesResults...), err
```
akrejcir, line 132: Can you change the second argument to `nil`? It will be more clear that this `return` statement does not return any error.

akrejcir (changes requested):
Looks good. Just one minor comment.

internal/operands/common-templates/reconcile.go
borod108: sure

PR #241, opened 2021-09-23
Recording rule and alert for SSPCommonTemplatesModificationReverted

borod108:
Adding kubevirt_ssp_total_restored_common_templates rule and
SSPCommonTemplatesModificationReverted alert to warn the user in case the
operator reverted changes made manually to a common template.

```release-note
Added SSPCommonTemplatesModificationReverted Alert which will activate as a warning in case of manual changes to a common are reverted.
```


akrejcir:
I think the alert will be triggered during update of SSP operator to new version, because it can update existing templates to new versions. Is it acceptable?

borod108:
@akrejcir hmm that is a good point, is there a way to know if that is the reason for the update? I will look into this.
Anyway I can easily add this as a note in the runbook, how often do we expect this to happen?
@sradco do you have a thought on this?

akrejcir:
The current operator version is accessible through an env variable, see `getOperatorVersion()` method in `controllers/ssp_controller.go`. You could move this method to some common package and make it accessible from other parts of the codebase.
The SSP CR has a status filed `.status.observedVersion`, which contains the last operator version that finished reconciliation.

During an update, the operator version will be increased, but the `ObservedVersion` will increase after all resources are updated. You can use this to detect if an update is in progress.

borod108:
@akrejcir great idea.
I think I should have put it in a different commit and perhaps PR though :/
If you find it important I will move it.
But in any case let me know if this changes are what you had in mind.

sradco:
> @akrejcir hmm that is a good point, is there a way to know if that is the reason for the update? I will look into this. Anyway I can easily add this as a note in the runbook, how often do we expect this to happen? @sradco do you have any thoughts on this?

@borod108  If the alert will fire frequently and many time due to upgrades the user will eventually ignore this alert..

borod108:
@sradco yes I agree, I have added a fix based on @akrejcir suggestion.

internal/operands/metrics/resources.go
```diff
@@ -76,6 +76,22 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 								"severity": "Warn",
```
sradco, line 76: @borod108  Its a bug here  "severity": "warning".
Above severity for "SSPDown" and "SSPTemplateValidatorDown" alerts should be "critical" with lower case.

internal/operands/metrics/resources.go
```diff
@@ -76,6 +76,22 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
...
 							},
 						},
+						{
+							Record: "kubevirt_ssp_total_restored_common_templates",
+							Expr:   intstr.FromString("sum(increase(total_restored_common_templates{pod=~'ssp-operator.*'}[1h]))"),
+						},
+						{
+							Alert: "SSPCommonTemplatesModificationReverted",
+							Expr:  intstr.FromString("kubevirt_ssp_total_restored_common_templates > 0"),
```
sradco, line 85: @borod108 Is there a way that "kubevirt_ssp_total_restored_common_templates" will return no values?

borod108: I do not think so. If it is registered it will return a value.

internal/operands/metrics/resources.go
```diff
@@ -76,6 +76,22 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
...
+							For:   "5m",
+							Annotations: map[string]string{
+								"summary":     "Common Templates manual modifications were reverted by the operator",
+								"runbook_url": "https://kubevirt.io/monitoring/runbooks/SSPCommonTemplatesModificationReverted",
+							},
+							Labels: map[string]string{
+								"severity": "Warn",
```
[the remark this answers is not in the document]
borod108: done

internal/operands/metrics/resources.go
borod108: OK, will fix in separate PRs for others - is this a convention? because I did see the Alerts firing properly on my env.

internal/operands/metrics/resources.go
sradco: @borod108 Where did you add the OR vector(0) ?

PR #246, opened 2021-10-15
Create DataImportCrons specified in SSP CR

akrejcir:
**What this PR does / why we need it**:
- Added reconciliation for `DataImportCrons` defined in SSP resource.
- Added permissions to view and edit `DataSources` and `DataImportCrons` to roles `os-images.kubevirt.io:view` and `os-images.kubevirt.io:edit`.

The `DataImportCrons` will be used by CDI to create `DataSources`, that will be referenced by common templates.

**Release note**:
```release-note
- Create DataImportCrons defined in SSP resource.
- Added permissions to view and edit DataSources and DataImportCrons to roles "os-images.kubevirt.io:view" and "os images.kubevirt.io:edit"
```

PR #247, opened 2021-10-19
Detect if a resource is being deleted and wait for cleanup

akrejcir:
The reconciliation function did not handle the case when a resource is being deleted, but is not gone yet. This is noticeable for namespace filled with other resources.

After this change, resources that are being deleted are treated as not ready, and the SSP resource will not be in Deployed state.

This PR is needed by CI in https://github.com/kubevirt/ssp-operator/pull/246

**Release note**:
```release-note
When removing SSP CR, the operator will wait for all deployed resources to be removed before removing the finalizer.
```


internal/common/resource.go
```diff
@@ -132,6 +138,80 @@ func CreateOrUpdate(request *Request) ReconcileBuilder {
...
+		return CleanupResult{}, err
+	}
+
+	isOwned := false
+	for _, reference := range found.GetOwnerReferences() {
+		if reflect.DeepEqual(reference, resource.GetOwnerReferences()[0]) {
+			isOwned = true
+		}
+	}
+	if found.GetAnnotations() != nil {
```
ksimon1, line 173: is possible to add here another check `&& !isOwned` ?

akrejcir: Sure. This is one of the cases, when I cannot decide between readability and performance. Both improvements are very small.

PR #258, opened 2021-11-24
Add common labels into alert definitions 

assafad:
We want to be able to list all kubevirt alerts so we added labels to differentiate them.

Signed-off-by: assafad aadmi@redhat.com

**Release note**:

```release-note
Added common labels into alert definitions and constant to use as runbook url prefix
```

### DIFF
--- a/internal/operands/metrics/resources.go
+++ b/internal/operands/metrics/resources.go
@@ -6,7 +6,15 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
-const PrometheusRuleName = "prometheus-k8s-rules-cnv"
+const (
+	PrometheusRuleName       = "prometheus-k8s-rules-cnv"
+	runbookURLBasePath       = "http://kubevirt.io/monitoring/runbooks/"
+	severityAlertLabelKey    = "severity"
+	partOfAlertLabelKey      = "kubernetes_operator_part_of"
+	partOfAlertLabelValue    = "kubevirt"
+	componentAlertLabelKey   = "kubernetes_operator_component"
+	componentAlertLabelValue = "ssp-operator"
+)
 
 func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 	return &promv1.PrometheusRule{
@@ -42,10 +50,12 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 							For:   "5m",
 							Annotations: map[string]string{
 								"summary":     "All SSP operator pods are down.",
-								"runbook_url": "https://kubevirt.io/monitoring/runbooks/SSPOperatorDown",
+								"runbook_url": runbookURLBasePath + "SSPOperatorDown",
 							},
 							Labels: map[string]string{
-								"severity": "critical",
+								severityAlertLabelKey:  "critical",
+								partOfAlertLabelKey:    partOfAlertLabelValue,
+								componentAlertLabelKey: componentAlertLabelValue,
 							},
 						},
 						{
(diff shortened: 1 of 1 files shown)


ksimon1:
Hi,
thank you for your PR. Can you please sign your commit (`git commit -s`)?
/ok-to-test

akrejcir:
Can you please create a regular commit? This one is a merge commit.
And also, you don't need to create a new PR if you want to change something, you can force push new changes:
```bash
git push -f ${YOUR_FORK_REPO} HEAD:add-labels
```

akrejcir:
Sorry, it is not a merge commit, but it has wrong title and author.

simonpasquier (commented):
I might miss the bigger context for this change but assuming that when the alerts fire, they carry a `namespace` and `service` labels, why do we want additional static labels?

sradco:
> I might miss the bigger context for this change but assuming that when the alerts fire, they carry a `namespace` and `service` labels, why do we want additional static labels?

Hi @simonpasquier , The namespace and service are insufficient. We need to be able to differentiate which operator sent the alert. This will allow us to better analyse the data about fired alerts and also we will be able to visualize for CNV the alerts that are specific to it.

internal/operands/metrics/resources.go
```diff
@@ -45,7 +45,9 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 								"runbook_url": "https://kubevirt.io/monitoring/runbooks/SSPOperatorDown",
 							},
 							Labels: map[string]string{
-								"severity": "critical",
+								"severity":                      "critical",
+								"kubernetes_operator_part_of":   "kubevirt",
+								"kubernetes_operator_component": "ssp-operator",
```
akrejcir, line 50: Can you extract the repeated strings into constants?

internal/operands/metrics/resources.go
```diff
@@ -45,7 +47,9 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 								"runbook_url": "https://kubevirt.io/monitoring/runbooks/SSPOperatorDown",
 							},
 							Labels: map[string]string{
-								"severity": "critical",
+								"severity":                      "critical",
+								"kubernetes_operator_part_of":   SSPOperatorPartOf,
+								"kubernetes_operator_component": SSPOperatorComponent,
```
akrejcir, line 52: Can you also extract the keys to constants (`severity`, `kubernetes_operator_part_of`, `kubernetes_operator_component`) ? So we can be sure there is not a typo in the strings used in multiple places.

akrejcir (changes requested):
Can you squash the 2 commits together?

internal/operands/metrics/resources.go
assaf-admi: Sure, I'll do it in a separate PR.

akrejcir: Why in a separate PR? They are added in this PR, so you can add them as constants instead of as literals?

internal/operands/metrics/resources.go
```diff
@@ -45,7 +47,9 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
 								"runbook_url": "https://kubevirt.io/monitoring/runbooks/SSPOperatorDown",
 							},
 							Labels: map[string]string{
-								"severity": "critical",
+								"severity":                      "critical",
```
ksimon1, line 50: Can you please extract severity as separate constant?

internal/operands/metrics/resources.go
```diff
@@ -7,6 +7,8 @@ import (
 )
 
 const PrometheusRuleName = "prometheus-k8s-rules-cnv"
+const SSPOperatorPartOf = "kubevirt"
```
ksimon1, line 10: Please rewrite this to:
```
const (
  PrometheusRuleName = "prometheus-k8s-rules-cnv"
  SSPOperatorPartOf = "kubevirt"
  SSPOperatorComponent = "ssp-operator"
)
```
Are these variable  `SSPOperatorPartOf` and `SSPOperatorComponent` will be used in different package(e.g. tests)?
If no, can you please make them non accessible from other packages?

internal/operands/metrics/resources.go
assafad: Hi @erkanerol, do you think extracting the keys into constants makes sense in the case?

internal/operands/metrics/resources.go
assafad: Hi, @ksimon1, severity changes between alerts so we rather not using constant in the case.

sradco:
@assafad please update the runbook_url value as well to have a prefix + the runbook name

PR #261, opened 2021-11-30
Create DataSource for each common template that uses it

akrejcir:
New version of common templates will reference `DataSources` instead of an existing `PVCs`.
This PR adds code to create these `DataSources`.

This PR depends on: https://github.com/kubevirt/common-templates/pull/385

**Release note**:
```release-note
DataSource is created for each common template that uses it.
```


internal/template-bundle/bundle.go
```diff
@@ -0,0 +1,101 @@
...
+		template := templatev1.Template{}
+		err = decoder.Decode(&template)
+		if err == io.EOF {
+			return bundle, nil
+		}
+		if err != nil {
+			return nil, err
+		}
+		if template.Name != "" {
+			bundle = append(bundle, template)
```
ksimon1, line 49: Would be possible to combine `readTemplates` and `extractDataSources` into one function (to not iterate through the template slice twice)?

akrejcir: Yes, would be possible, but I think it would be less clear code.
The way it is now, the two functions do only one thing. Combining them would be a little faster, but more complex code. And more performance is not important, because the templates are loaded only once when the operator starts.

akrejcir:
I'm verifying this.

/hold

akrejcir:
Verified 
/hold cancel

akrejcir:
This PR depends on: https://github.com/kubevirt/common-templates/pull/385

PR #262, opened 2021-11-30
Limit TemplateValidator Replicas in SNO mode

borod108:
When in SNO mode there can be only one TemplateValidator. Or less.

```release-note
When running in Single Replica Topology Mode there will be no more than one TemplateValidator Replica running.
```


internal/common/environment.go
```diff
@@ -23,3 +28,16 @@ func EnvOrDefault(envName string, defVal string) string {
 func GetOperatorVersion() string {
 	return EnvOrDefault(OperatorVersionKey, defaultOperatorVersion)
 }
+
+func IsSingleNodeMode(topologyMode osconfv1.TopologyMode) bool {
```
ksimon1, line 32: Is this fn used somewhere?

internal/operands/template-validator/reconcile.go
```diff
@@ -136,7 +136,11 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
 	if image == "" {
 		panic("Cannot reconcile without valid image name")
 	}
-	deployment := newDeployment(request.Namespace, *validatorSpec.Replicas, image)
+	number_of_replicas := *validatorSpec.Replicas
```
ksimon1, line 139: can you please use camelCase format

internal/common/environment.go
```diff
@@ -23,3 +28,16 @@ func EnvOrDefault(envName string, defVal string) string {
...
+	return topologyMode == osconfv1.SingleReplicaTopologyMode
+}
+
+func GetInfrastructureTopology(c client.Client) (osconfv1.TopologyMode, error) {
```
akrejcir, line 36: This function is only called from `ssp_operator.go`, can you move it there?
I know it is related to environment, but I think moving a function to a different package is useless, if it is not going to be used in multiple places.

internal/common/environment.go
```diff
@@ -23,3 +28,16 @@ func EnvOrDefault(envName string, defVal string) string {
...
+	infraConfig := &osconfv1.Infrastructure{}
```
akrejcir, line 37: Openshift API needs to be added to scheme, otherwise the GET call will fail. You can add it in `internal.common.scheme` package.

internal/common/environment.go
```diff
@@ -23,3 +28,16 @@ func EnvOrDefault(envName string, defVal string) string {
...
+	if err := c.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
+		return osconfv1.SingleReplicaTopologyMode, err
```
akrejcir, line 39: Is it OK to return `SingleReplicaTopologyMode` in case of any error? It seems suboptimal, if for example there is a network error, we assume single node topology.

Maybe on an error we should restart the SSP operator to try again?

internal/common/request.go
```diff
@@ -17,4 +18,9 @@ type Request struct {
 	Instance     *ssp.SSP
 	Logger       logr.Logger
 	VersionCache VersionCache
+	TopologyMode osconfv1.TopologyMode
+}
+
+func (r *Request) IsSingleModeTopology() bool {
```
akrejcir, line 24: A typo. Should be `IsSingleNodeTopology`.

tests/single_node_test.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var _ = Describe("Single Node Topology", func() {
+	BeforeEach(func() {
+		strategy.SkipUnlessSingleReplicaTopologyMode()
+	})
+
+	AfterEach(func() {
+		strategy.RevertToOriginalSspCr()
+		waitUntilDeployed()
```
akrejcir, line 17: Please move this call to `BeforeEach()`

tests/single_node_test.go
```diff
@@ -0,0 +1,61 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		defer watch.Stop()
+
+		var newValidatorReplicas int32 = 3
+		updateSsp(func(foundSsp *sspv1beta1.SSP) {
+			foundSsp.Spec.TemplateValidator.Replicas = &newValidatorReplicas
+		})
+
+		// Watch changes until above change
+		err = WatchChangesUntil(watch, func(updatedSsp *sspv1beta1.SSP) bool {
```
akrejcir, line 31: You are missing the second `WatchChangesUntil` call. The reason for 2 calls is to ensure that the operator has seen the above change done by `updateSsp`.

First `WatchChangesUntil` will trigger before the operator sees the change, because `updatedSsp.Generation > updatedSsp.Status.ObservedGeneration` .
And the second `WatchChangesUntil` should watch until `updatedSsp.Generation == updatedSsp.Status.ObservedGeneration`. To be sure the the operator has seen the change.

tests/single_node_test.go
```diff
@@ -0,0 +1,61 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		defer watch.Stop()
+
+		var newValidatorReplicas int32 = 0
+		updateSsp(func(foundSsp *sspv1beta1.SSP) {
+			foundSsp.Spec.TemplateValidator.Replicas = &newValidatorReplicas
+		})
+
+		// Watch changes until above change
+		err = WatchChangesUntil(watch, func(updatedSsp *sspv1beta1.SSP) bool {
```
akrejcir, line 52: The same here.

tests/tests_suite_test.go
```diff
@@ -173,6 +177,18 @@ func (s *newSspStrategy) SkipSspUpdateTestsIfNeeded() {
 	// Do not skip SSP update tests in this strategy
 }
 
+func (s *newSspStrategy) SkipUnlessSingleReplicaTopologyMode() {
+	if topologyMode != osconfv1.SingleReplicaTopologyMode {
+		Skip("Tests that are specific for HighlyAvailableTopologyMode are disabled", 1)
+	}
+}
+
+func (s *newSspStrategy) SkipUnlessHighlyAvailableTopologyMode() {
```
akrejcir, line 186: This method is not used anywhere. If you don't plan to use it in a future PR, can you remove it?

akrejcir (changes requested):

internal/common/environment.go
borod108: fixed

internal/operands/template-validator/reconcile.go
borod108: thank you, old habbits die hard :)

tests/tests_suite_test.go
borod108: I am actually planning to merge this PR only after the SNO lane will be added and then I will need this method. If I will not use it I will remove it. Thank you for noticing.

internal/common/environment.go
borod108: I was naively taught that files should be small and things should sit in seperate places depending on semantics. I see we do not do that much, and perhaps this is the Go way.
The logic behind moving stuff into seperate places early is that the chances someone else will move it later if it grows are smaller, since moving code is harder than just writing in the same place (for lazy people like me).
However I do not have strong feelings about this - if you think I should move it, I will, let me know.

internal/common/environment.go
borod108: oh I did not know! Thank you!

internal/common/environment.go
borod108: I was checking the error value and restarting, but you are right, this is confusing. Changed to ""

internal/common/request.go
borod108: LOL, I was thinking of SingleReplica and TopologyMode, and combined them in the most confusing way :)

tests/single_node_test.go
borod108: yes, done

tests/single_node_test.go
borod108: oh and there was another problem there. Is this better now?
by the way can we have a timing issue here? by going to wait for ">" after it is already "=="?

tests/single_node_test.go
borod108: yes

internal/common/environment.go
akrejcir: You are right, I missed it.

tests/single_node_test.go
akrejcir: This is good.

There should not be a timing issue. The watch is started at the beginning of the test, before any changes to the SSP CR. A watch is basically a channel that contains all changes to a resource in chronological order, as they happened on the server. So an event with ">" will be before "==", because SSP operator must first see ">" and then change it to "==".

tests/tests_suite_test.go
akrejcir: Ok

internal/common/environment.go
akrejcir: I don't know if it is a "Go way", it is probably just my code style, and I did not realize it.
This file does not have a clear purpose, I made it to contain environment variables, but it can have other code.

I think file size alone does not matter with current IDEs, I don't usually read the full file, I only jump to relevant functions. 
Separating based on semantics is a good idea. Also moving functions is easy with current IDEs. For example there is a functionality in GoLand that can move a function or structure to a different file and fix all dependencies in "one click".

You can keep this here.

controllers/ssp_controller.go
```diff
@@ -84,6 +86,11 @@ func CreateAndSetupSspReconciler(mgr ctrl.Manager) error {
 		return err
 	}
 
+	infrastructureTopology, err := common.GetInfrastructureTopology(mgr.GetClient())
```
akrejcir, line 89: The CI failure is related to this line. When you look into the logs of ssp-operator pod on CI, there is this error:
```
2021-12-06T14:53:57.729Z	ERROR	setup	unable to create controller	{"controller": "SSP", "error": "the cache is not started, can not read objects"}
```

Searching the source code for `"the cache is not started, can not read objects"`, I see that this error is returned by the client cache, when informers were not started. Controller-runtime uses cached client, and the cachce internally uses informers to watch resources and update itself. The cache will be started when the manager is started.

This line of code is executed before the manager is started, so the cache is not running. Instead of `mgr.GetClient()` you should be able to use the uncached API reader `mgr.GetAPIReader()`.

borod108: Oh, thank you for the explanation!

borod108:
@akrejcir @ksimon1 can you take a look?

tests/single_node_test.go
```diff
@@ -0,0 +1,74 @@
...
+				updatedSsp.Generation == updatedSsp.Status.ObservedGeneration && updatedSsp.Status.Phase == api.PhaseDeployed
+		}, shortTimeout)
+		Expect(err).ToNot(HaveOccurred())
+
+		deployment := getTemplateValidatorDeployment()
+		Expect(int(deployment.Status.Replicas)).Should(Equal(1), "In Single Mode Topology the number of replicas is at most 1")
+	})
+
+	It("[test_id:7793] Number of Template Validator replicas can be set to 0", func() {
+		watch, err := StartWatch(sspListerWatcher)
```
akrejcir, line 50: This test case does not need to use watch. It is enough to call `waitUntilDeployed()` before checking the template validator deployment.

Watch on SSP is useful when the test needs to be sure that the operator reacted to a change, but there is no external way to find out. For example in the previous test, you set the number of replicas to 3, but because this is single node, the deployment is not actually changed. In that case the test needs to watch the `ObservedGeneration` to be sure that the operator has seen the new SSP CR.

tests/tests_suite_test.go
```diff
@@ -352,6 +388,13 @@ func getSsp() *sspv1beta1.SSP {
 	return foundSsp
 }
 
+func getTemplateValidatorDeployment() *apps.Deployment {
```
akrejcir, line 391: This function is only used from a single file, it could be moved there. But you don't have to.

tests/validator_test.go
```diff
@@ -196,6 +196,7 @@ var _ = Describe("Template validator operand", func() {
 	})
 
 	It("[test_id:4913] should successfully start template-validator pod", func() {
+		strategy.SkipUnlessHighlyAvailableTopologyMode()
```
akrejcir, line 199: This line is a little confusing to me. Would it be better to call the function `SkipForSingleReplicaTopologyMode()` ?

Reading this, I need to know that HighlyAvailableTopology == not SingleReplicaTopology, that there are no other choices.

akrejcir (commented):
Looks good. I only have a few optional comments. You can implement them in a future PR if you want.

PR #263, opened 2021-11-30
Add tool to generate metrics documentation

machadovilaca:

This PR adds a tool to automatically generate documentation with information about SSP exposed metrics.

Fixes #

**Release note**:

```release-note
NONE
```


docs/metrics.md
```diff
@@ -0,0 +1,19 @@
...
+### kubevirt_ssp_rejected_vms_total
+The total number of vms rejected by virt-template-validator
+### kubevirt_ssp_template_validator_up_total
+The total number of running virt-template-validator pods
+### kubevirt_ssp_total_restored_common_templates
+The total number of common templates restored by the operator back to their original state
+## Developing new metrics
+After developing new metrics or changing old ones, please run `make generate-doc` to regenerate this document.
+
+If you feel that the new metric doesn't follow these rules, please change `tools/metricsdocs` with your needs.
```
borod108, line 19: If at this point we do not have any builtin mechanism to consistently generate a different doc for a specific metric, I think we should drop this line.
If in the future we add such a feature (which I doubt we will need) we can add it back. Else we are just saying - if you do not like this, change the code. It is very nice to note which code to change, but you already kindly added that information in the beginning of this doc.

internal/operands/metrics/resources.go
```diff
@@ -16,6 +16,133 @@ const (
...
+		{
+			Name:    "kubevirt_ssp_total_restored_common_templates",
+			Expr:    intstr.FromString("sum(increase(total_restored_common_templates{pod=~'ssp-operator.*'}[1h])) OR on() vector(0)"),
+			Summary: "The total number of common templates restored by the operator back to their original state",
+		},
+	}
+}()
+
+func getPrometheusRules() []promv1.Rule {
+	rules := []promv1.Rule{
```
borod108, line 56: I think it would be nice to have consistent depth of reference, if we separated SspMetrics into a seperate var, lets separate Alerts rules as well.

internal/operands/metrics/resources.go
```diff
@@ -16,6 +16,133 @@ const (
 	componentAlertLabelValue = "ssp-operator"
 )
 
+type MetricDesc struct {
```
borod108, line 19: So if I understand correctly we define MetricDesc then we cast it to metric in one place, and build a list of those.
In another place we go over MetricDesc and create promv1.Rules out of them.
I feel this is overly complicated for what we are doing here. We basically never use MetricDesc for anything directly.
Perhaps your idea is to make it easier to add metric rules as in the Open Close principle? If so, I like the idea, however it will be a bit confusing because the summary is actually part of the annotations. So either lets make the MetricDesc structure have "annotations" and in them the "summary", or maybe we can just promv1.Rule?
I would go with just having promv1.Rule

tools/metricsdocs/metricsdocs.go
```diff
@@ -0,0 +1,88 @@
...
+
+func writeToFile(metricsList metricList) {
+	fmt.Print(opening)
+	metricsList.writeOut()
+	fmt.Print(footer)
+}
+
+type metric struct {
+	name        string
+	description string
```
borod108, line 43: Again I think we can just use promv1.Rule and define the list type over it.

borod108 (commented):
Thank you, this is very nice, I think we can make it even a bit simpler (added some comments to that end).

machadovilaca:
@borod108

Updated the code to match some of the requested changes. 

Some simplifications, namely related to the usage of `promv1.Rule` are not possible because it can't have the `annotations` field has stated in kubevirt/hostpath-provisioner-operator#195.

But on the other hand, in my perspective, this also means that, through the use of `RecordRulesDesc`, the fetching and parsing of record rules and their descriptions to generate the document is now more clear.

tools/metricsdocs/metricsdocs.go
```diff
@@ -0,0 +1,96 @@
...
+	background = "This document aims to help users that are not familiar with metrics exposed by the SSP Operator.\n" +
+		"All metrics documented here are auto-generated by the utility tool `tools/metricsdocs` and reflects exactly what is being exposed.\n\n"
+
+	KVSpecificMetrics = "## SSP Operator Metrics List\n"
+
+	opening = title +
+		background +
+		KVSpecificMetrics
+
+	// footer
```
borod108, line 22: you can safely remove this comment :)

tools/metricsdocs/metricsdocs.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (m metric) writeOut() {
+	fmt.Println("###", m.name)
+	fmt.Println(m.description)
+}
+
+type metricList []metric
+
+// Len implements sort.Interface.Len
```
borod108, line 69: I think you can remove this and the next similar comments.

internal/operands/metrics/resources.go
```diff
@@ -16,6 +16,129 @@ const (
...
 )
 
+// RecordRulesDesc represent SSP Operator Prometheus Record Rules
+type RecordRulesDesc struct {
+	Name        string
+	Expr        intstr.IntOrString
+	Description string
+}
+
+// RecordRulesDescList returns SSP Operator Prometheus Record Rules
```
borod108, line 26: As far as I understand - this comment does not give an accurate description, the object below does not return things. It is a list that holds objects which are partial representations of the recording rules. I would just remove this comment, the name is good enough.

borod108 (commented):
Thank you for the improvements, added very small comments about the comments, please address them, and we are good to go as far as I can see :)

borod108:
@akrejcir @ksimon1 please take a look so we can merge. It looks good to me once the extra comments are removed.

machadovilaca:
@borod108 removed the `footer` comment and tried to improve the one for `RecordRulesDescList`. Did not remove the latter and the ones from the methods that implement `sort.Interface` because Go convections states that "exported names should have doc comments"[1] and most linters will throw an error on those.

[1] https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

borod108:
@machadovilaca awesome!
/lgtm

tools/metricsdocs/metricsdocs.go
akrejcir: Or if you want to check at compile time if `metricList` implements an interface you can add this line:
```go
var _ sort.Interface = metricList{}
```

akrejcir (changes requested):
Can you cleanup the commits? There are some redundant changes. One commit creates a file and a later commit removes this file, or one commit creates a file and later one updates it.
Can you change the commits so files are created how you want them the first time?

tools/metricsdocs/metricsdocs.go
```diff
@@ -0,0 +1,94 @@
...
+	footer = footerHeading + footerContent
+)
+
+func main() {
+	metricsList := recordRulesDescToMetricList(metrics.RecordRulesDescList)
+	sort.Sort(metricsList)
+	writeToFile(metricsList)
+}
+
+func writeToFile(metricsList metricList) {
```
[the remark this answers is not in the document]
machadovilaca: updated

machadovilaca:
@akrejcir @borod108 updated

PR #269, opened 2021-12-03
Add initial controller to watch missing CRDs

akrejcir:
If not all needed CRDs are installed, a controller that watches and waits for CRDs will be started.
When CRDs are installed, the initial controller will exit and the SSP controller will start.

These CRDs are needed:
- `datavolumes.cdi.kubevirt.io`
- `datasources.cdi.kubevirt.io`
- `dataimportcrons.cdi.kubevirt.io`
- `prometheusrules.monitoring.coreos.com` (is installed with OCP)

Without this PR, the SSP controller would fail to watch the resources if their CRDs do not exist, and the application would exit with an error. Which would cause the operator deployment to enter a crash loop.

**Release note**:
```release-note
The operator waits until all needed CRDs are installed.
```

### DIFF
--- a/controllers/setup.go
+++ b/controllers/setup.go
@@ -1,0 +1,89 @@
+package controllers
+
+import (
+	"context"
+	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"path/filepath"
+
+	"kubevirt.io/ssp-operator/internal/operands"
+	"kubevirt.io/ssp-operator/internal/operands/common-templates"
+	"kubevirt.io/ssp-operator/internal/operands/data-sources"
+	"kubevirt.io/ssp-operator/internal/operands/metrics"
+	"kubevirt.io/ssp-operator/internal/operands/node-labeller"
+	"kubevirt.io/ssp-operator/internal/operands/template-validator"
+	"sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+func CreateAndSetupReconciler(mgr controllerruntime.Manager) error {
+	templatesFile := filepath.Join(templateBundleDir, "common-templates-"+common_templates.Version+".yaml")
+	templatesBundle, err := common_templates.ReadTemplates(templatesFile)
+	if err != nil {
+		return err
+	}
+
+	sspOperands := []operands.Operand{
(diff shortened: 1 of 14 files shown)


akrejcir:
I need to verify one more thing.
/hold

There are no tests, because our current functional testing setup does not know how to restart the operator.

akrejcir:
Verified

/hold cancel

dominikholler:
@akrejcir Can you please explicitly list in the commit message, that the controller is waiting for "datavolumes.cdi.kubevirt.io",	"datasources.cdi.kubevirt.io" and "dataimportcrons.cdi.kubevirt.io" and explain why the controller is waiting for them?

controllers/setup.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	var requiredCrds []string
+	for i := range sspOperands {
+		requiredCrds = append(requiredCrds, sspOperands[i].RequiredCrds()...)
+	}
+
+	// Check if all needed CRDs exist
+	crdList := &extv1.CustomResourceDefinitionList{}
+	err = mgr.GetAPIReader().List(context.TODO(), crdList)
```
ksimon1, line 40: wouldn't be possible to use `ListOptions` with `FieldSelector` (use field selector for CRD names)? And list only CRDS which are interesting for us? And with that result it should be possible to easily check is all CRDs are ready `if len(crdList.Items) == len(requiredCrds)`

akrejcir: No, it is not possible. 
The field selector does not support set operations like label selector, so we cannot use: 
```
metadata.name IN {"datasources.cdi.kubevirt.io", "dataimportcrons.cdi.kubevirt.io"}
```
nor does it support logical `OR`, like:
```
(metadata.name == "datasources.cdi.kubevirt.io") OR (metadata.name == "dataimportcrons.cdi.kubevirt.io")
```

akrejcir:
> @akrejcir Can you please explicitly list in the commit message, that the controller is waiting for "datavolumes.cdi.kubevirt.io", "datasources.cdi.kubevirt.io" and "dataimportcrons.cdi.kubevirt.io" and explain why the controller is waiting for them?

Ok, I have updated the commit message and PR description.

akrejcir:
@borod108 I have applied your suggestions.

PR #278, opened 2021-12-29
Add test cases for runbook URLs

assafad:
This PR adds new tests cases for alert definitions. We expect to see "runbook_url" annotations in alerts and those urls must be accessible.

**Release note**:

```release-note
None
```


assafad:
Hi @sradco , @erkanerol, can you please review this PR?

tests/metrics_test.go
```diff
@@ -67,4 +67,10 @@ var _ = Describe("Metrics", func() {
 			expectAppLabelsRestoreAfterUpdate(&prometheusRuleRes)
 		})
 	})
+
+	Context("alerts", func() {
+		It("[test_id:CNV-7851]should have available runbook URLs", func() {
```
sradco, line 72: @erkanerol should the test number have the "CNV-" prefix?

erkanerol: I missed that :( That id should be polarion id of that test case. If there is no such a polarion test case, you should create it or delete that test_id label from there.

assafad: Here is the polarion test case: https://polarion.engineering.redhat.com/polarion/#/project/CNV/workitem?id=CNV-7851
Should I change the id to 7851?

erkanerol: Yes. Please drop `CNV-` prefix.

assafad: Thanks, done

tests/tests_common_test.go
```diff
@@ -412,3 +414,23 @@ func addDomainResourcesToVMI(vmi *kubevirtv1.VirtualMachineInstance, cpuCores ui
 	}
 	return vmi
 }
+
+func expectRunbookURLs() {
```
akrejcir, line 418: This function is only called from one test case. Can you inline it there?

tests/tests_common_test.go
```diff
@@ -412,3 +414,23 @@ func addDomainResourcesToVMI(vmi *kubevirtv1.VirtualMachineInstance, cpuCores ui
...
+	promRuleObj := metrics.NewPrometheusRule(strategy.GetNamespace())
```
akrejcir, line 419: This is a functional test, so I think it would be better to check the runbook URLs on actual PrometheusRule resource that was created by the operator.

Can you change this test, so that it fetches the PrometheusRule from the cluster?

tests/tests_common_test.go
```diff
@@ -412,3 +414,23 @@ func addDomainResourcesToVMI(vmi *kubevirtv1.VirtualMachineInstance, cpuCores ui
...
+				ExpectWithOffset(1, rule.Annotations).ToNot(BeNil())
+				url, ok := rule.Annotations["runbook_url"]
+				ExpectWithOffset(1, ok).To(BeTrue())
+				checkAvailabilityOfUrl(url)
+			}
+		}
+	}
+}
+
+func checkAvailabilityOfUrl(url string) {
```
akrejcir, line 432: Can you move this function to `metrics_test.go`? Or maybe you can inline it.

tests/tests_common_test.go
```diff
@@ -412,3 +414,23 @@ func addDomainResourcesToVMI(vmi *kubevirtv1.VirtualMachineInstance, cpuCores ui
...
+	resp, err := http.Get(url)
+	ExpectWithOffset(2, err).ToNot(HaveOccurred())
+	ExpectWithOffset(2, resp.StatusCode).Should(Equal(200))
```
akrejcir, line 435: Can you use `http.StatusOK` constant, instead of the literal `200`?

akrejcir (changes requested):
It think it would make sense if the functional test checks the actual PrometheusRule resource that is created in the cluster.

tests/tests_common_test.go
akrejcir: Once you inline this function into `metrics_test.go`, you can get the rule from cluster like this:
```go
promRule := &promv1.PrometheusRule{}
Expect(apiClient.Get(ctx, prometheusRuleRes.GetKey(), promRule)).To(Succeed())
```

Similarly how it is done in test case: `[test_id:4665] should create prometheus rule`.

akrejcir:
The test looks good.
Can you cleanup the commits? Remove unneeded code, squash everything into 1 commit and then force push it into this PR?

akrejcir:
Thanks!

/lgtm

ksimon1:
Tests are failing due to issue in CDI. New version of CDI should be released tomorrow.

PR #287, opened 2022-01-12
Change default log level to be less verbose

borod108:
Fix https://bugzilla.redhat.com/show_bug.cgi?id=2037421

```release-note
Changed log level
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -185,7 +185,7 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 	}
 	sspRequest.Logger.V(1).Info("CR status updated")
 
-	sspRequest.Logger.V(1).Info("Reconciling operands...")
+	sspRequest.Logger.Info("Reconciling operands...")
 	reconcileResults, err := r.reconcileOperands(sspRequest)
 	if err != nil {
 		return handleError(sspRequest, err)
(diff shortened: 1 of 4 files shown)


ksimon1:
/lgtm
I would appreciate the ability to set this from e.g. CR, but it can be done in some different PR.

ksimon1:
@borod108 can you please remove [wip] from subject?

borod108:
@ksimon1 I want to get the full version of the fix, this way it is very annoying, i will only ask to merge this if I will not make a full fix in time...

controllers/ssp_controller.go
```diff
@@ -197,7 +197,7 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	sspRequest.Logger.V(1).Info("CR status updated")
+	sspRequest.Logger.Info("CR status updated")
```
akrejcir, line 200: Can you keep these 2 log lines in debug, and add a third log line with info level at the end of this method, that will say something like: Reconciliation finished.

akrejcir (changes requested):
Looks good, just one minor comment.

controllers/ssp_controller.go
borod108: actually this is done on purpose to give more information (if the log line it is there we have more info on what failed)

akrejcir: ok

PR #288, opened 2022-01-12
Use CamelCased reason values in SSP status Condition

0xFelix:
According to kubernetes API conventions the reason field of a Condition shall contain a CamelCased value.

This fixes the reason field of SSP status Conditions not following kubernetes API conventions

See https://bugzilla.redhat.com/show_bug.cgi?id=2032845

Fixes #

**Release note**:

```release-note
The reason field of SSP status Conditions now follows kubernetes API conventions by using CamelCased values.
```


borod108:
/lgtm.
I think we need to mention the change in the relase-note in this case although it is very minor, since it is visible to a user.

akrejcir:
Yes, there is a release note section at the end. Currently it is `None`.

Right now the release notes are probably not used anywhere, but we want to create a script that would collect them and add them to a release.

PR #289, opened 2022-01-13
Delete immutable resources only if immutable part was changed

akrejcir:
Usually only the `.spec` field of resource is immutable, `.metadata` and `.status` can be updated.
With this PR, the resource is only deleted, if `.spec` is changed, otherwise it is updated.

Partially fixes bug: https://bugzilla.redhat.com/show_bug.cgi?id=2037290

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/common/resource.go
+++ b/internal/common/resource.go
@@ -1,7 +1,6 @@
 package common
 
 import (
-	"context"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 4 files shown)


akrejcir:
I need to change one more thing.
/hold

akrejcir:
Overriding the test of upgrade from old operator. The old operator does not work on new cluster version that is used in CI.
See: https://github.com/openshift/release/pull/25287

/override ci/prow/e2e-upgrade-from-old-operator-functests

PR #292, opened 2022-01-20
fix applying deprecated label

ksimon1:
the application of deprecated label was not working correctly,
because it doesn't skip old templates with the same name as newly
deployed templates.
Added a condition, which check if template has newer version than is
the version in ssp operator. If yes, template is not set as deprecated
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2042799

**Release note**:
```
NONE
```

### DIFF
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -12,12 +12,9 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	semver "github.com/blang/semver/v4"
 	"kubevirt.io/ssp-operator/internal/common"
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	kubevirt.io/controller-lifecycle-operator-sdk v0.2.1
 	kubevirt.io/qe-tools v0.1.7
 	sigs.k8s.io/controller-runtime v0.9.3
+	golang.org/x/mod v0.5.0
```
akrejcir, line 34: We already import semver package: `github.com/blang/semver/v4`.
But this is ok, you don't need to change this PR.

akrejcir (commented):
/lgtm

go.mod
ksimon1: I will update it, because its not worth to introduce new dependency

PR #306, opened 2022-02-07
Move webhook code to its own package

akrejcir:
Having webhook logic in API package is not a good design. When another application wants to import API module, it usually does not need the webhook implementation.

**Release note**:
```release-note
None
```


ksimon1:
/lgtm
does HCO has to do any changes?

akrejcir:
No, they don't use the webhook code.

PR #309, opened 2022-02-15
Fix flaky tests

borod108:
Remove a line that causes unnecessary flakiness

```release-note
None
```

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -123,8 +123,6 @@ var _ = Describe("SSPOperatorReconcilingProperly metric", func() {
 	})
 
 	It("[test_id:7369] should set SSPOperatorReconcilingProperly metrics to 0 on failing to reconcile", func() {
-		foundSsp := getSsp()
-		Expect(foundSsp.Status.Phase).To(Equal(lifecycleapi.PhaseDeployed), "SSP should be in phase Deployed")
 		// add a finalizer to the validator deployment, do that it can't be deleted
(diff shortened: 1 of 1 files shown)


tests/misc_test.go
```diff
@@ -123,8 +123,6 @@ var _ = Describe("SSPOperatorReconcilingProperly metric", func() {
 	})
 
 	It("[test_id:7369] should set SSPOperatorReconcilingProperly metrics to 0 on failing to reconcile", func() {
-		foundSsp := getSsp()
-		Expect(foundSsp.Status.Phase).To(Equal(lifecycleapi.PhaseDeployed), "SSP should be in phase Deployed")
 		// add a finalizer to the validator deployment, do that it can't be deleted
 		Expect(apiClient.Get(ctx, deploymentRes.GetKey(), deployment)).ToNot(HaveOccurred())
 		controllerutil.AddFinalizer(deployment, finalizerName)
```
akrejcir, line 128: Can you also wrap the `Update` below in an `Eventually`? It may fail because of concurrent update.

akrejcir (changes requested):

tests/misc_test.go
akrejcir: I will post another PR with this fix.

akrejcir: This PR: https://github.com/kubevirt/ssp-operator/pull/313

PR #310, opened 2022-02-15
Test running VMs with RHEL images

akrejcir:
Added tests that create and start VMs with RHEL 8 and RHEL 9 images.

**Release note**:
```release-note
None
```


automation/common/deploy-kubevirt-and-cdi.sh
```diff
@@ -28,13 +28,11 @@ spec:
   configuration:
     selinuxLauncherType: "virt_launcher.process"
     developerConfiguration:
+      useEmulation: true
```
ksimon1, line 31: you don't need this, when ssp tests runs on azure

akrejcir: Ok, removed this change.

tests/misc_test.go
```diff
@@ -190,3 +199,154 @@ var _ = Describe("SCC annotation", func() {
...
+
+		// Wait for VM to start
+		Eventually(func() (bool, error) {
+			foundVm := &kubevirtv1.VirtualMachine{}
+			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), foundVm)
+			if err != nil {
+				return false, err
+			}
+
+			return foundVm.Status.Ready, nil
```
dominikholler, line 327: @akrejcir What does the Status.Ready means here? Does it mean the VM is booting, or that the guest is booted?

akrejcir: The `status.Ready` is `true` when the pod has ready condition `true`. In this case, there is no readiness probe, so the pod is ready as soon as it's running. It does not mean anything about the VM.

I can add a readiness probe, that will check if the guest os has successfully booted.

dominikholler: @akrejcir thanks for adding the ReadinessProbe!

PR #314, opened 2022-02-23
code change: Fix issues found by inspection tool

akrejcir:
Fixed some issues found by "Inspect code" tool in Goland.

**Release note**:
```release-note
None
```


internal/common/scheme.go
```diff
@@ -19,5 +19,5 @@ func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(Scheme))
 	utilruntime.Must(extv1.AddToScheme(Scheme))
 	utilruntime.Must(sspv1beta1.AddToScheme(Scheme))
-	utilruntime.Must(osconfv1.AddToScheme(Scheme))
+	utilruntime.Must(osconfv1.Install(Scheme))
```
ksimon1, line 22: What does this change do?

akrejcir: `AddToScheme` is deprecated, `Install` is the same function, with different name:
https://github.com/openshift/api/blob/master/config/v1/register.go#L21

ksimon1: So can you please change it for all schemas?

akrejcir: Not all modules rename it. Some keep the `AddToScheme` name. I guess they will rename it in later versions.

akrejcir: You are right, it is only deprecated in OpenShift API.

PR #319, opened 2022-03-02
Add default podAntiAffinity to PodTemplateSpec

fossedihelm:
Default PodTemplateSpec now has a podAntiAffinity that makes pod replicas to be scheduled on different nodes.
This affinity will be merged with the requested one.
The merge affect nodeAffinity, podAffinity, podAntiaffinity, tolerations, and nodeSelector

Fixes #

**Release note**:

```release-note
TemplateValidator has a podAntiAffinity that makes pod replicas to be scheduled on different nodes.
```

### DIFF
--- a/internal/operands/template-validator/reconcile.go
+++ b/internal/operands/template-validator/reconcile.go
@@ -2,12 +2,11 @@ package template_validator
 
 import (
 	"fmt"
-
 	admission "k8s.io/api/admissionregistration/v1"
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
-	lifecycleapi "kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
+	"kubevirt.io/ssp-operator/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"kubevirt.io/ssp-operator/internal/common"
@@ -145,7 +144,7 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
 		numberOfReplicas = 1
 	}
 	deployment := newDeployment(request.Namespace, numberOfReplicas, image)
-	addPlacementFields(deployment, validatorSpec.Placement)
+	injectPlacementMetadata(&deployment.Spec.Template.Spec, &validatorSpec)
 	return common.CreateOrUpdate(request).
 		NamespacedResource(deployment).
 		WithAppLabels(operandName, operandComponent).
@@ -173,15 +172,75 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
 		Reconcile()
 }
 
-func addPlacementFields(deployment *apps.Deployment, nodePlacement *lifecycleapi.NodePlacement) {
-	if nodePlacement == nil {
+// Merge all Tolerations, Affinity and NodeSelectors from NodePlacement into pod spec
+func injectPlacementMetadata(podSpec *v1.PodSpec, componentConfig *v1beta1.TemplateValidator) {
+	if componentConfig == nil || componentConfig.Placement == nil {
 		return
 	}
+	if podSpec == nil {
+		podSpec = &v1.PodSpec{}
+	}
+	nodePlacement := componentConfig.Placement
+	if podSpec.NodeSelector == nil {
+		podSpec.NodeSelector = make(map[string]string, len(nodePlacement.NodeSelector))
+	}
+	for nsKey, nsVal := range nodePlacement.NodeSelector {
+		// Favor podSpec over NodePlacement. This prevents cluster admin from clobbering
+		// node selectors that operator intentionally set.
+		if _, ok := podSpec.NodeSelector[nsKey]; !ok {
+			podSpec.NodeSelector[nsKey] = nsVal
+		}
+	}
 
-	podSpec := &deployment.Spec.Template.Spec
-	podSpec.Affinity = nodePlacement.Affinity
-	podSpec.NodeSelector = nodePlacement.NodeSelector
-	podSpec.Tolerations = nodePlacement.Tolerations
+	if nodePlacement.Affinity != nil {
+		if podSpec.Affinity == nil {
+			podSpec.Affinity = nodePlacement.Affinity.DeepCopy()
+		} else {
+			mergeNodeAffinity(podSpec.Affinity, nodePlacement.Affinity.NodeAffinity)
+			mergePodAffinity(podSpec.Affinity, nodePlacement.Affinity.PodAffinity)
+			mergePodAntiAffinity(podSpec.Affinity, nodePlacement.Affinity.PodAntiAffinity)
+		}
+	}
+	podSpec.Tolerations = append(podSpec.Tolerations, nodePlacement.Tolerations...)
+}
+
+func mergeNodeAffinity(currentAffinity *v1.Affinity, configNodeAffinity *v1.NodeAffinity) {
+	if configNodeAffinity != nil {
+		if currentAffinity.NodeAffinity == nil {
+			currentAffinity.NodeAffinity = configNodeAffinity.DeepCopy()
+			return
+		}
+		if configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+			if currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+				currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.DeepCopy()
+			} else {
+				currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = append(currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms, configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms...)
+			}
+		}
+		currentAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, configNodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution...)
+	}
+}
+
+func mergePodAffinity(currentAffinity *v1.Affinity, configPodAffinity *v1.PodAffinity) {
+	if configPodAffinity != nil {
+		if currentAffinity.PodAffinity == nil {
+			currentAffinity.PodAffinity = configPodAffinity.DeepCopy()
+			return
+		}
+		currentAffinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution, configPodAffinity.RequiredDuringSchedulingIgnoredDuringExecution...)
+		currentAffinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution, configPodAffinity.PreferredDuringSchedulingIgnoredDuringExecution...)
+	}
+}
+
+func mergePodAntiAffinity(currentAffinity *v1.Affinity, configPodAntiAffinity *v1.PodAntiAffinity) {
+	if configPodAntiAffinity != nil {
+		if currentAffinity.PodAntiAffinity == nil {
+			currentAffinity.PodAntiAffinity = configPodAntiAffinity.DeepCopy()
+			return
+		}
+		currentAffinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution, configPodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution...)
+		currentAffinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution, configPodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution...)
+	}
 }
 
 func reconcileValidatingWebhook(request *common.Request) (common.ReconcileResult, error) {
--- a/internal/operands/template-validator/reconcile_test.go
+++ b/internal/operands/template-validator/reconcile_test.go
@@ -2,6 +2,8 @@ package template_validator
 
 import (
 	"context"
+	"github.com/onsi/ginkgo/extensions/table"
+	lifecycleapi "kubevirt.io/controller-lifecycle-operator-sdk/pkg/sdk/api"
 	"testing"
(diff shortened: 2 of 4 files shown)


internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,113 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+// Merge all Tolerations, Affinity and NodeSelectors from NodePlacement into pod spec
+func injectPlacementMetadata(podSpec *v1.PodSpec, componentConfig *v1beta1.TemplateValidator) {
+	if componentConfig == nil || componentConfig.Placement == nil {
 		return
 	}
+	if podSpec == nil {
+		podSpec = &v1.PodSpec{}
+	}
+	nodePlacement := componentConfig.Placement
+	if len(nodePlacement.NodeSelector) == 0 {
```
iholder-redhat, line 184: Any reason why this is being done instead of `if nodePlacement.NodeSelector == nil`? If `nodePlacement.NodeSelector` is an empty map, can't we just continue and avoid re-allocating a new map?

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,113 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+		nodePlacement.NodeSelector = make(map[string]string)
+	}
+	if len(podSpec.NodeSelector) == 0 {
+		podSpec.NodeSelector = make(map[string]string, len(nodePlacement.NodeSelector))
+	}
+	// podSpec.NodeSelector
```
iholder-redhat, line 190: nit: I think this is unnecessary

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,113 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+	for nsKey, nsVal := range nodePlacement.NodeSelector {
+		// Favor podSpec over NodePlacement. This prevents cluster admin from clobbering
+		// node selectors that operator intentionally set.
+		if _, ok := podSpec.NodeSelector[nsKey]; !ok {
+			podSpec.NodeSelector[nsKey] = nsVal
+		}
+	}
+
+	// podSpec.Affinity
```
iholder-redhat, line 199: nit: same as above, I'd remove this

iholder-redhat (commented):
Thanks @fossedihelm! Great job! :+1: 

Overall looks great.

-----

Small note about my perspective on comments:
IMO comments should be used to explain complicated code that cannot be simplified easily. In such cases comments need to expose as less implementation details as possible, i.e. explain what the codes does and not how it does it. Also, comments should never point out the obvious (explain something that is immediately obvious from looking at the code).

The reason I'm even mentioning this is because comments can easily do more harm than good. The danger is that the code changes but the comments stay outdated. This can lead to a very confusing code.

Example for what I'm referring to:
```go
// Iterating over Pod's node selector
for i := Pod.NodeSelector { ... }
```

What do you think? :)

fossedihelm:
@iholder-redhat I pushed the requested changes. I also push a change on `validator_test` file: it is the fix of the failing tests. Thanks

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
 }
 
-func addPlacementFields(deployment *apps.Deployment, nodePlacement *lifecycleapi.NodePlacement) {
-	if nodePlacement == nil {
+// Merge all Tolerations, Affinity and NodeSelectors from NodePlacement into pod spec
+func injectPlacementMetadata(podSpec *v1.PodSpec, componentConfig *v1beta1.TemplateValidator) {
+	if componentConfig == nil || componentConfig.Placement == nil {
 		return
 	}
+	if podSpec == nil {
```
akrejcir, line 180: This check can be removed, because the calling function never passes `nil` as `podSpec`. But if you want to keep this check, it can return on `nil`, because the function will be noop.

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
 
-	podSpec := &deployment.Spec.Template.Spec
-	podSpec.Affinity = nodePlacement.Affinity
-	podSpec.NodeSelector = nodePlacement.NodeSelector
-	podSpec.Tolerations = nodePlacement.Tolerations
+func mergeNodeAffinity(currentAffinity *v1.Affinity, configNodeAffinity *v1.NodeAffinity) {
+	if currentAffinity.NodeAffinity == nil {
+		currentAffinity.NodeAffinity = configNodeAffinity.DeepCopy()
+	} else {
+		// need to copy all affinity terms one by one
```
akrejcir, line 229: Can you remove the comment? It is clear from the code.

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+		if configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+			if currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+				currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.DeepCopy()
+			} else {
+				// merge the list of terms from NodePlacement into podSpec
+				for _, term := range configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
```
akrejcir, line 235: This loop can be replaced by this line:
```go
currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = append(currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms, configNodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms...)
```

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+					currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = append(currentAffinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms, term)
+				}
+			}
+		}
+
+		//PreferredDuringSchedulingIgnoredDuringExecution
```
akrejcir, line 241: Can you remove this comment too?

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+		for _, term := range configNodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution {
```
akrejcir, line 242: This loop can also be replaced by:
```go
currentAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, configNodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution...)
```

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+			currentAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, term)
+		}
+
+	}
+}
+
+func mergePodAffinity(currentAffinity *v1.Affinity, configPodAffinity *v1.PodAffinity) {
+	if currentAffinity.PodAffinity == nil {
+		currentAffinity.PodAffinity = configPodAffinity.DeepCopy()
+	} else {
```
akrejcir, line 252: Can you add early `return` ?

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+		//RequiredDuringSchedulingIgnoredDuringExecution
```
akrejcir, line 253: Also this comment

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+		for _, term := range configPodAffinity.RequiredDuringSchedulingIgnoredDuringExecution {
```
akrejcir, line 254: Loop can be:
```go
currentAffinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution, configPodAffinity.RequiredDuringSchedulingIgnoredDuringExecution...)
```

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+			currentAffinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAffinity.PreferredDuringSchedulingIgnoredDuringExecution, term)
+		}
+	}
+}
+
+func mergePodAntiAffinity(currentAffinity *v1.Affinity, configPodAntiAffinity *v1.PodAntiAffinity) {
+	if currentAffinity.PodAntiAffinity == nil {
+		currentAffinity.PodAntiAffinity = configPodAntiAffinity.DeepCopy()
+	} else {
+		//RequiredDuringSchedulingIgnoredDuringExecution
```
akrejcir, line 268: Comment

internal/operands/template-validator/reconcile.go
```diff
@@ -173,15 +172,108 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
...
+		for _, term := range configPodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution {
+			currentAffinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution, term)
+		}
+		//PreferredDuringSchedulingIgnoredDuringExecution
+		for _, term := range configPodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution {
```
akrejcir, line 273: Loop:
```go
currentAffinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(currentAffinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution, configPodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution...)
```

internal/operands/template-validator/reconcile_test.go
```diff
@@ -83,6 +88,207 @@ var _ = Describe("Template validator operand", func() {
 		ExpectResourceExists(newValidatingWebhook(namespace), request)
 	})
 
+	Context("with different requests", func() {
+
+		nodeAffinity := &core.NodeAffinity{
```
akrejcir, line 93: Can you assign to these variables in `BeforeEach()`, so the test are independent?

internal/operands/template-validator/reconcile_test.go
```diff
@@ -83,6 +88,207 @@ var _ = Describe("Template validator operand", func() {
...
+						LabelSelector: &meta.LabelSelector{
+							MatchExpressions: []meta.LabelSelectorRequirement{
+								{
+									Key:      "prefPodAffinityKey",
+									Operator: meta.LabelSelectorOpIn,
+									Values:   []string{"true"},
+								},
+							},
+						},
+						TopologyKey: "kubernetes.io/hostname",
```
akrejcir, line 137: This string is used in many places, can you extract it to a constant?

internal/operands/template-validator/reconcile_test.go
```diff
@@ -83,6 +88,207 @@ var _ = Describe("Template validator operand", func() {
...
+						},
+					},
+					TopologyKey: "kubernetes.io/hostname",
+				},
+			},
+		}
+
+		var setPodAntiAffinity = func(request *common.Request) {
+			request.Instance.Spec.TemplateValidator.Placement.Affinity.PodAntiAffinity = antiAffinity
+			return
```
akrejcir, line 259: This `return` can be removed.

internal/operands/template-validator/reconcile_test.go
```diff
@@ -83,6 +88,207 @@ var _ = Describe("Template validator operand", func() {
...
+		}
+		var setPodAffinity = func(request *common.Request) {
+			request.Instance.Spec.TemplateValidator.Placement.Affinity.PodAffinity = podAffinity
+			return
```
akrejcir, line 263: This `return` can be removed.

internal/operands/template-validator/reconcile_test.go
```diff
@@ -83,6 +88,207 @@ var _ = Describe("Template validator operand", func() {
...
+		}
+		var setNodeAffinity = func(request *common.Request) {
+			request.Instance.Spec.TemplateValidator.Placement.Affinity.NodeAffinity = nodeAffinity
+			return
```
akrejcir, line 267: This `return` can be removed.

internal/operands/template-validator/reconcile_test.go
```diff
@@ -83,6 +88,207 @@ var _ = Describe("Template validator operand", func() {
...
+		}
+
+		table.DescribeTable("should create correct deployment", func(requestAdjustFunctions []func(*common.Request), expectedNodeAffinity *core.NodeAffinity, expectedPodAffinity *core.PodAffinity, expectedPodAntiAffinity *core.PodAntiAffinity) {
+			for _, f := range requestAdjustFunctions {
+				f(&request)
+			}
+			res, err := operand.Reconcile(&request)
+			Expect(err).ToNot(HaveOccurred())
+			deploymentResult := res[4]
```
akrejcir, line 276: Can you use the client to get the deployment instead of accessing the result of the call? The resource in result is only used for getting names and logging.

You can use:
```go
deployment := &apps.Deployment{}
key := client.ObjectKeyFromObject(newDeployment(namespace, replicas, "test-img"))
Expect(request.Client.Get(request.Context, key, deployment)).To(Succeed())
```

akrejcir (changes requested):
Thanks for the PR.
My comments are mostly about code style. You can ignore them, if you want to. Except the last one about using client in `reconcile_test.go`, that would be better.

fossedihelm:
@akrejcir Thanks for your review. It was constructive and instructive. I'm going to address your suggestions. Thanks

internal/operands/template-validator/reconcile_test.go
akrejcir: Right, I didn't notice that. You can leave it as is.

akrejcir:
I've noticed that the commit message is not split into a title and body. When I run `git branch -vv` it shows the whole commit message as title.
But you don't need to change it now.

fossedihelm:
> I've noticed that the commit message is not split into a title and body. When I run `git branch -vv` it shows the whole commit message as title. But you don't need to change it now.

Thanks. I think I forgot a `\n`. :slightly_frowning_face:

Issue #323, opened 2022-03-09
Don't log the error when an object is modified outside of the operator


akrejcir:
The operator creates many resources and some of them can be updated by kubernetes, or other pods in the cluster. When an outside component updates a resource while the operator is updating it, kubernetes API returns an error. These errors cause the operator to retry the update again, but they pollute the logs. 

The error should not be logged, or logged with a debug level.

This is an example of the log line:
```json
 {
  "level":"error",
  "ts":1646818725.6815221,
  "logger":"controller-runtime.manager.controller.ssp",
  "msg":"Reconciler error",
  "reconciler group":"ssp.kubevirt.io",
  "reconciler kind":"SSP",
  "name":"ssp-kubevirt-hyperconverged",
  "namespace":"kubevirt-hyperconverged",
  "error":"Operation cannot be fulfilled on ssps.ssp.kubevirt.io \"ssp-kubevirt-hyperconverged\": the object has been modified; please apply your changes to the latest version and try again",
  "stacktrace": "..."
}
```

/kind enhancement

PR #327, opened 2022-03-23
Handle error when CR is in conflict

opokornyy:

When action on CR is rejected by API server,
dont show the error in log and reenque the request.

Fixes https://github.com/kubevirt/ssp-operator/issues/323

**Release note**:
```
NONE
```


akrejcir:
Have you tested that the error is no longer in logs after this change?

Issue #322, opened 2022-03-07
Add a new column to the 'oc get' for SSP resource.


akrejcir:
This task is to add a new column to the `oc get ssp`, to show `.status.phase` field.

/kind enhancement

ksimon1:
@akrejcir can we close this issue, since https://github.com/kubevirt/ssp-operator/pull/328 is merged?

PR #328, opened 2022-03-24
Add new column

opokornyy:

When 'oc get ssp' command is executed,
that the '.status.phase' field will be displayed.

Fixes # https://github.com/kubevirt/ssp-operator/issues/322

**Release note**:
```
Print column Phase when getting SSP objects by 'oc get ssp' command.
```

### DIFF
--- a/api/v1beta1/ssp_types.go
+++ b/api/v1beta1/ssp_types.go
@@ -95,6 +95,7 @@ type SSPStatus struct {
 // +kubebuilder:subresource:status
 
 // SSP is the Schema for the ssps API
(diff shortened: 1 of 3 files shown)


akrejcir:
@opokornyy One small thing: Can you run `go mod vendor` and `go mod tidy` ? It will update the copy of `api` package that is in the `vendor` directory.

0xFelix:
Looks good to me, but a release note would be helpful in my opinion.

lyarwood:
I agree with @0xFelix , some brief releasenote text would be nice here if you have time @opokornyy, otherwise this looks ready.

lyarwood:
Thanks :)

/approve

PR #330, opened 2022-03-25
[release-v0.13] `e2e-upgrade-functests` lane tests release from a specified branch

akrejcir:
This patch adds a new parameter for `e2e-upgrade-functests lane`, which specifies the release branch with the latest release.

This is needed when a PR is targeted for a release branch that is not latest. Without this change, the CI would test
downgrade, instead of upgrade.

**Release note**:
```release-note
None
```


ksimon1:
/hold
@dominikholler can we merge this? (it will be merged to 4.10)

PR #331, opened 2022-03-28
Remove selected template labels on reconcile

akrejcir:
The common reconcile method does not remove labels and annotations from resources, which is useful to not remove user-defined labels and annotations.
    
This PR adds a special logic to remove labels and annotations from templates, if they are not user-defined. For example, if the default template for an OS changes, then the label needs to be removed from the old template.

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-logr/logr"
 	templatev1 "github.com/openshift/api/template/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


lyarwood:
> **What this PR does / why we need it**: The common reconcile method does not remove labels and annotations from resources, which is useful to not remove user-defined labels and annotations.
>
> This PR adds a special logic to remove labels and annotations from templates, if they are not user-defined. For example, if the default template for an OS changes, then the label needs to be removed from the old template.

Should this also be written up as an issue/bug somewhere?

lyarwood:
Regardless of the missing issue/bug this looks fine to me FWIW.

/lgtm

akrejcir:
> 
> Should this also be written up as an issue/bug somewhere?

I think it's not needed. This problem was only in master branch, which was not released.

lyarwood:
> > Should this also be written up as an issue/bug somewhere?
> 
> I think it's not needed. This problem was only in master branch, which was not released.

ACK that's a fair point.

PR #332, opened 2022-03-29
Add metrics types to SSP metrics description

assafad:
When querying Prometheus it is extremely important to know the metric type: gauge/counter/histogram/summary.
This PR Adds metrics types to SSP metrics documentation.

```release-note
None
```


tools/metricsdocs/metricsdocs.go
```diff
@@ -54,12 +55,13 @@ func metricDescriptionToMetric(rrd metrics.RecordRulesDesc) metric {
...
 		name:        rrd.Name,
 		description: rrd.Description,
+		mtype:       rrd.Type,
 	}
 }
 
 func (m metric) writeOut() {
 	fmt.Println("###", m.name)
-	fmt.Println(m.description)
+	fmt.Println(m.description + ". Type: " + m.mtype)
```
sradco, line 64: @assafad Please add a dot at the end of the line.

internal/operands/metrics/resources.go
```diff
@@ -42,26 +43,31 @@ var RecordRulesDescList = []RecordRulesDesc{
...
+		Type:        "Gauge",
 	},
 	{
 		Name:        "kubevirt_ssp_num_of_operator_reconciling_properly",
 		Expr:        intstr.FromString("sum(ssp_operator_reconciling_properly)"),
 		Description: "The total number of ssp-operator pods reconciling with no errors",
+		Type:        "Gauge",
 	},
 	{
 		Name:        "kubevirt_ssp_rejected_vms_total",
```
sradco, line 61: Hi, I believe we should remove the kubevirt_ssp_rejected_vms_total  and also kubevirt_ssp_total_restored_common_templates recording rules, since they are only used in 1 alert and the expression is simple and there is no need to waist resources on saving them as recording rules.

assafad:
Hi @borod108, can you please have a look?

PR #333, opened 2022-04-04
Update apiextensions version used in csv-generator to v1

akrejcir:
The `csv-generator` deserializes CRD into the correct struct.

**Release note**:
```release-note
None
```


data/crd/ssp.kubevirt.io_ssps.yaml
```diff
@@ -14,7 +14,11 @@ spec:
     singular: ssp
   scope: Namespaced
   versions:
-  - name: v1beta1
+  - additionalPrinterColumns:
```
0xFelix, line 17: Shouldn't that be part of https://github.com/kubevirt/ssp-operator/pull/328?

akrejcir: yes, but we forgot about it.

PR #343, opened 2022-04-21
Remove evictionStrategy from VM in SNO

fossedihelm:
In single replica infrastructure having "LiveMigrate" in
`template.spec.evictionStrategy` is wrong, since live migration requires
at least one node in which migrate.
Remove the field only in single node.

Fixes #

**Release note**:

```release-note
Remove evictionStrategy `LiveMigrate` from template in SNO cluster
```

### DIFF
--- a/internal/template-bundle/bundle.go
+++ b/internal/template-bundle/bundle.go
@@ -34,7 +34,11 @@ func ReadBundle(filename string, topologyMode osconfv1.TopologyMode, scheme *run
 	}
 
 	codec := serializer.NewCodecFactory(scheme).LegacyCodec(kubevirtv1.GroupVersion)
-	manageSingleReplicaInfrastructure(templates, &topologyMode, codec)
+	err = manageSingleReplicaInfrastructure(templates, &topologyMode, codec)
+	if err != nil {
+		return Bundle{}, err
+	}
+
 	return Bundle{
 		Templates:   templates,
 		DataSources: sources,
@@ -197,9 +201,11 @@ func manageSingleReplicaInfrastructure(templates []templatev1.Template, topology
 		return nil
 	}
 
-	for templateIdx, template := range templates {
-		for objIdx, rawObj := range template.Objects {
-			vmUnstructured, isVm, err := decodeToVMUnstructured(&rawObj)
+	for templateIdx := range templates {
+		objects := templates[templateIdx].Objects
+		for objIdx := range objects {
+			rawObj := &objects[objIdx]
+			vmUnstructured, isVm, err := decodeToVMUnstructured(rawObj)
 			if err != nil {
 				return err
 			}
@@ -218,7 +224,11 @@ func manageSingleReplicaInfrastructure(templates []templatev1.Template, topology
 			}
 
 			unstructured.RemoveNestedField(vmUnstructured.Object, "spec", "template", "spec", "evictionStrategy")
-			templates[templateIdx].Objects[objIdx].Raw = []byte(runtime.EncodeOrDie(codec, vmUnstructured))
+			encoded, err := runtime.Encode(codec, vmUnstructured)
+			if err != nil {
+				return err
+			}
+			templates[templateIdx].Objects[objIdx].Raw = encoded
 		}
 	}
 	return nil
--- a/internal/template-bundle/bundle.go
+++ b/internal/template-bundle/bundle.go
@@ -3,8 +3,12 @@ package template_bundle
 import (
 	"bytes"
 	"fmt"
+	osconfv1 "github.com/openshift/api/config/v1"
 	"io"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	templatev1 "github.com/openshift/api/template/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -18,7 +22,7 @@ type Bundle struct {
 	DataSources []cdiv1beta1.DataSource
 }
 
-func ReadBundle(filename string) (Bundle, error) {
+func ReadBundle(filename string, topologyMode osconfv1.TopologyMode, scheme *runtime.Scheme) (Bundle, error) {
 	templates, err := readTemplates(filename)
 	if err != nil {
 		return Bundle{}, err
@@ -29,6 +33,8 @@ func ReadBundle(filename string) (Bundle, error) {
 		return Bundle{}, err
 	}
 
+	codec := serializer.NewCodecFactory(scheme).LegacyCodec(kubevirtv1.GroupVersion)
+	manageSingleReplicaInfrastructure(templates, &topologyMode, codec)
 	return Bundle{
 		Templates:   templates,
 		DataSources: sources,
@@ -93,14 +99,11 @@ func extractDataSources(templates []templatev1.Template) ([]cdiv1beta1.DataSourc
 }
 
 func vmTemplateUsesSourceRef(template *templatev1.Template) (bool, error) {
-	vmUnstructured := &unstructured.Unstructured{}
-	err := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(template.Objects[0].Raw), 1024).Decode(vmUnstructured)
+	vmUnstructured, isVm, err := decodeToVMUnstructured(&template.Objects[0])
 	if err != nil {
 		return false, err
 	}
 
-	isVm := vmUnstructured.GetAPIVersion() == "kubevirt.io/v1" &&
-		vmUnstructured.GetKind() == "VirtualMachine"
 	if !isVm {
 		return false, fmt.Errorf("template %s contains unexpected object: %s, %s", template.Name, vmUnstructured.GetAPIVersion(), vmUnstructured.GetKind())
 	}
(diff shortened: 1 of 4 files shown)


internal/operands/common-templates/reconcile.go
```diff
@@ -1,7 +1,13 @@
 package common_templates
 
 import (
+	"bytes"
 	"fmt"
+	v1 "github.com/openshift/api/config/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	v12 "kubevirt.io/api/core/v1"
```
ksimon1, line 10: can you rename this to e.g. kubevirtv1?

internal/operands/common-templates/reconcile.go
```diff
@@ -78,6 +84,7 @@ func (c *commonTemplates) RequiredCrds() []string {
 }
 
 func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ReconcileResult, error) {
+	c.manageSingleReplicaInfrastructure(request)
```
ksimon1, line 87: can you move this fn to here https://github.com/kubevirt/ssp-operator/blob/master/internal/operands/common-templates/reconcile.go#L51, because it doesn't have to run every reconcile iteration, does it?

internal/operands/common-templates/reconcile.go
```diff
@@ -113,6 +120,30 @@ func incrementTemplatesRestoredMetric(reconcileResults []common.ReconcileResult,
...
+			vm := &v12.VirtualMachine{}
+			err := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(rawObj.Raw), 1024).Decode(vm)
+			if err != nil {
+				continue
+			}
+
+			if vm.Spec.Template.Spec.EvictionStrategy != nil && *vm.Spec.Template.Spec.EvictionStrategy == v12.EvictionStrategyLiveMigrate {
+				request.Logger.Info(fmt.Sprintf("Deleting EvictionStrategy from Spec"))
+				vm.Spec.Template.Spec.EvictionStrategy = nil
+				codec := serializer.NewCodecFactory(request.Client.Scheme()).LegacyCodec(v12.GroupVersion)
```
ksimon1, line 139: Do you have to create NewCodecFactory for each VM?

internal/operands/common-templates/reconcile.go
```diff
@@ -113,6 +120,30 @@ func incrementTemplatesRestoredMetric(reconcileResults []common.ReconcileResult,
...
+	if request.TopologyMode != v1.SingleReplicaTopologyMode {
+		return
+	}
+
+	for templateIdx, template := range c.templatesBundle {
+		for objIdx, rawObj := range template.Objects {
+			vm := &v12.VirtualMachine{}
+			err := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(rawObj.Raw), 1024).Decode(vm)
+			if err != nil {
+				continue
```
ksimon1, line 133: Do we want to continue if there is some error?

internal/operands/common-templates/reconcile.go
fossedihelm: Thanks, my bad. We definitely don't want it.

internal/operands/common-templates/reconcile.go
fossedihelm: Thanks, :+1:

internal/operands/common-templates/reconcile.go
fossedihelm: Since `Objects` are `runtime.RawExtension` I think that it is not guaranteed that they are all `VirtualMachine`. It should be, but in case they are not we can skip it IMO. If you think that it is better to block it I can first Decode to `Unstructured`, check if it is a VM, then Decode to VM and check if an error occurs. Or maybe the assumption that Objects can contain something different from VM is wrong and so I can simply raise an error here. WDYT? Thanks

internal/operands/common-templates/reconcile.go
fossedihelm: I think that you are right, we can move this in creation. In this case `setup.go` should pass `infrastructureTopology` and `scheme|codec` information. Thanks for your review

akrejcir: Or you can move it to this file: https://github.com/kubevirt/ssp-operator/blob/master/internal/template-bundle/bundle.go
It contains code that loads the templates from yaml.

internal/operands/common-templates/reconcile.go
akrejcir: If the object is not a VM, then I think it is ok to skip it.
But it has to be decoded into `Unstructured` in any case, because the VM object in template may not be correct according to its schema. Templates can have variables that are substituted. The variable names are strings and they may be in places where `VirtualMachine` struct contains a different type, so the decode would fail.

akrejcir (changes requested):

internal/operands/common-templates/reconcile.go
fossedihelm: Nice point! You're right. I'm going to adjust the decode and work with `Unstructured`, thanks

internal/operands/common-templates/reconcile.go
fossedihelm: Agreed, thanks :+1:

fossedihelm:
@ksimon1 @akrejcir  I pushed the requested changes:

- Moved the manageSNO function during the reading templates https://github.com/kubevirt/ssp-operator/blob/master/internal/template-bundle/bundle.go
- Use `Unstructured` instead of `VirtualMachine` to remove evictionStrategy if necessary

Thanks!

internal/template-bundle/bundle.go
```diff
@@ -29,6 +33,8 @@ func ReadBundle(filename string) (Bundle, error) {
 		return Bundle{}, err
 	}
 
+	codec := serializer.NewCodecFactory(scheme).LegacyCodec(kubevirtv1.GroupVersion)
+	manageSingleReplicaInfrastructure(templates, &topologyMode, codec)
```
akrejcir, line 37: Can you check the error returned from this function?

internal/template-bundle/bundle.go
```diff
@@ -176,3 +191,35 @@ func createDataSource(name, namespace string) cdiv1beta1.DataSource {
...
 	}
 }
+
+func manageSingleReplicaInfrastructure(templates []templatev1.Template, topologyMode *osconfv1.TopologyMode, codec runtime.Encoder) error {
+	if *topologyMode != osconfv1.SingleReplicaTopologyMode {
+		return nil
+	}
+
+	for templateIdx, template := range templates {
+		for objIdx, rawObj := range template.Objects {
```
akrejcir, line 201: These loops unnecessarily copy the objects inside the slices. Can you change them to only use indexes?
```go
for templateIdx := range templates {
	objects := templates[templateIdx].Objects
	for objIdx := range objects {
		rawObj := &objects[objIdx]
```

internal/template-bundle/bundle.go
```diff
@@ -176,3 +191,35 @@ func createDataSource(name, namespace string) cdiv1beta1.DataSource {
...
+			if err != nil {
+				return err
+			}
+
+			if !found || val != string(kubevirtv1.EvictionStrategyLiveMigrate) {
+				continue
+			}
+
+			unstructured.RemoveNestedField(vmUnstructured.Object, "spec", "template", "spec", "evictionStrategy")
+			templates[templateIdx].Objects[objIdx].Raw = []byte(runtime.EncodeOrDie(codec, vmUnstructured))
```
akrejcir, line 221: Can you use `runtime.Encode`  and return the error, instead of `EncodeOrDie` and panic?

internal/template-bundle/bundle_test.go
```diff
@@ -66,6 +85,28 @@ var _ = Describe("Template bundle", func() {
...
+
+func ExpectEvictionStrategyLiveMigrate(obj *runtime.RawExtension) {
+	vmUnstructured := expectEvictionStrategy(obj)
+	val, found, err := unstructured.NestedFieldNoCopy(vmUnstructured.Object, "spec", "template", "spec", "evictionStrategy")
+	Expect(err).ToNot(HaveOccurred())
+	Expect(found).To(BeTrue())
+	Expect(val).To(Equal(string(kubevirtv1.EvictionStrategyLiveMigrate)))
+}
+
+func expectEvictionStrategy(obj *runtime.RawExtension) *unstructured.Unstructured {
```
akrejcir, line 103: Can you rename this function? It does not do what it's name says.

akrejcir (changes requested):
Looks good. I have only small comments.

@ksimon1, I've noticed that the `evictionStrategy` is only added to templates in downstream. So we cannot write a functional test for upstream CI. It that ok?

ksimon1:
> Looks good. I have only small comments.
> 
> @ksimon1, I've noticed that the `evictionStrategy` is only added to templates in downstream. So we cannot write a functional test for upstream CI. It that ok?

We can do a test for it. In u/s it will always pass, but in d/s it will test this feature

jean-edouard:
@akrejcir thank you for the thorough review, I pushed a commit to address your comments, PTAL!

jean-edouard:
/retest
Looks like my latest commit breaks the functional tests, not sure what's going on, will investigate tomorrow if the test fail again.

fossedihelm:
@akrejcir Thanks for your review and thanks @jean-edouard for addressing the changes. I will try to figure out what is happening in e2e tests

akrejcir:
The test failures are flakes, not related to this PR.

/retest

akrejcir:
After investigating the tests a bit, I think the failures are probably related to the newest CDI release. So they are not flakes.
I will post a PR to fix it soon.

fossedihelm:
@akrejcir Thanks for the update :+1:

akrejcir:
The CI failures are not related to this PR.

PR #349, opened 2022-05-09
update openshift.io/scc label test

ksimon1:
update openshift.io/scc label test to check if scc label has value restricted-v2

**Release note**:
```
NONE
```

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -168,7 +168,7 @@ func validateSspIsFailingToReconcileMetric() {
 var _ = Describe("SCC annotation", func() {
 	const (
 		sccAnnotation = "openshift.io/scc"
-		sccRestricted = "restricted"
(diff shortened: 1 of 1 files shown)


akrejcir:
Does the annotation value depend on the version of openshift? If yes, can you change the test to allow both values?

tests/misc_test.go
```diff
@@ -168,7 +168,7 @@ func validateSspIsFailingToReconcileMetric() {
 var _ = Describe("SCC annotation", func() {
 	const (
 		sccAnnotation = "openshift.io/scc"
-		sccRestricted = "restricted"
+		sccRestricted = "restricted*"
```
akrejcir, line 171: This is probably a typo. It would correctly match `restricted-v2`, but also other strings like: `prefix-resticte-suffix`.
Can you change it to `^restricted` ?

PR #351, opened 2022-05-12
Log the reason why reconciliation is triggered

akrejcir:
Logs the type and namespace/name of the object that triggered the reconciliation.

This change will be helpful for: https://bugzilla.redhat.com/2082912

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -45,6 +45,7 @@ import (
 	osconfv1 "github.com/openshift/api/config/v1"
 	ssp "kubevirt.io/ssp-operator/api/v1beta1"
 	"kubevirt.io/ssp-operator/internal/common"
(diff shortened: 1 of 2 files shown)


0xFelix:
I think it would be better to keep it as simple as possible and log all events in the `handler_hook`. The logs could be turned on/off with a debug switch. Transferring some of the events into the `Reconcile` function feels hacky and goes against the controller-runtime philosophy. It would be also more correct like this because in a series of events it is currently unclear which event is logged as reason for the reconciliation. I fear this could lead to false assumptions.

akrejcir:
This PR probably introduces a bug. I'm looking into it.

/hold

akrejcir:
Issue fixed.

/unhold

PR #358, opened 2022-05-19
fix security constraint spp and validator containers

ksimon1:
fix security constraint spp and validator containers
This PR add adds multiple security context attributes for ssp operator and
template validator containers.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2088129

**Release note**:
```
NONE
```


dominikholler:
@ksimon1 @sarahbx  Is this a new bug, or is this something we did not notice before?

config/manager/manager.template.yaml
```diff
@@ -20,8 +20,6 @@ spec:
     spec:
       serviceAccountName: ssp-operator
       priorityClassName: system-cluster-critical
-      securityContext:
```
sarahbx, line 23: Is this not required for both pod and container spec?

sarahbx (commented):
lgtm

PR #361, opened 2022-05-24
Add predicates to only watch relevant resource changes

akrejcir:
Some changes in a resource are not relevant for the reoconcile function. Mainly changes in `.status` field.
This change adds a predicate to filter out these changes.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2082912

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -46,6 +46,7 @@ import (
 	ssp "kubevirt.io/ssp-operator/api/v1beta1"
 	"kubevirt.io/ssp-operator/internal/common"
 	handler_hook "kubevirt.io/ssp-operator/internal/controller/handler-hook"
+	"kubevirt.io/ssp-operator/internal/controller/predicates"
 	"kubevirt.io/ssp-operator/internal/operands"
 )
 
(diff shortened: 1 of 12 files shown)


controllers/ssp_controller.go
```diff
@@ -680,19 +676,43 @@ func watchClusterResources(builder *ctrl.Builder, sspOperands []operands.Operand
...
+	// Deduplicate watches
+	watchedTypes := make(map[reflect.Type]operands.WatchType)
 	for _, operand := range sspOperands {
-		for _, t := range watchTypesFunc(operand) {
-			if _, ok := watchedTypes[reflect.TypeOf(t)]; ok {
-				continue
+		for _, watchType := range watchTypesFunc(operand) {
+			key := reflect.TypeOf(watchType.Object)
+			if wt, ok := watchedTypes[key]; ok {
+				watchType.WatchFullObject = watchType.WatchFullObject || wt.WatchFullObject
```
0xFelix, line 686: Can you place a comment here, what this is for? I assume it should watch the full object as soon as more than one operand defines a Watch for it?

controllers/ssp_controller.go
```diff
@@ -680,19 +676,43 @@ func watchClusterResources(builder *ctrl.Builder, sspOperands []operands.Operand
...
 
-			builder.Watches(
-				&source.Kind{Type: t},
-				handler_hook.New(handler, hookFunc),
-			)
-			watchedTypes[reflect.TypeOf(t)] = struct{}{}
+	for _, watchType := range watchedTypes {
+		var predicates []predicate.Predicate
+		if !watchType.WatchFullObject {
+			predicates = []predicate.Predicate{relevantChangesPredicate()}
```
0xFelix, line 695: Maybe the Predicate could be part of WatchType so each operand could precisely define the predicate for its types? No/nil predicate would indicate `WatchFullObject`.

internal/operands/common-templates/reconcile.go
```diff
@@ -35,10 +35,11 @@ func init() {
...
 }
 
-func WatchClusterTypes() []client.Object {
-	return []client.Object{
-		&templatev1.Template{},
-	}
+func WatchClusterTypes() []operands.WatchType {
+	return []operands.WatchType{{
+		Object:          &templatev1.Template{},
+		WatchFullObject: true,
```
0xFelix, line 41: `WatchFullObject = true` is the default?

internal/operands/data-sources/reconcile.go
```diff
@@ -39,14 +39,14 @@ func init() {
...
-	return []client.Object{
-		&rbac.ClusterRole{},
-		&rbac.Role{},
-		&rbac.RoleBinding{},
-		&core.Namespace{},
-		&cdiv1beta1.DataSource{},
-		&cdiv1beta1.DataImportCron{},
+func WatchClusterTypes() []operands.WatchType {
+	return []operands.WatchType{
+		{Object: &rbac.ClusterRole{}, WatchFullObject: true},
```
0xFelix, line 44: `WatchFullObject = true` is the default?

internal/operands/data-sources/reconcile.go
```diff
@@ -39,14 +39,14 @@ func init() {
...
+		{Object: &rbac.Role{}, WatchFullObject: true},
+		{Object: &rbac.RoleBinding{}, WatchFullObject: true},
+		{Object: &core.Namespace{}},
+		{Object: &cdiv1beta1.DataSource{}},
+		{Object: &cdiv1beta1.DataImportCron{}},
```
0xFelix, line 49: `WatchFullObject = true` is the default?

ksimon1: when bool is not defined, it has always value false

controllers/ssp_controller.go
akrejcir: Not "more then one operand". If at least one operand wants to watch the full object, then the watch should allow it.

controllers/ssp_controller.go
akrejcir: I would rather keep it simple. If sometime we will need different predicates for various operands, we can add the code.

internal/operands/common-templates/reconcile.go
akrejcir: No, `false` is default.

controllers/ssp_controller.go
0xFelix: Ok, a comment would still be helpful.

akrejcir: Done

akrejcir:
I need to manually verify that the loop does not happen.
/hold

akrejcir:
This PR alone will not fix the loop, because CDI resources have their `generation` updated also on `status` changes.

/hold cancel

akrejcir:
I will change this, so that it does not assume that `generation` does not update when `status` updates.
/hold

akrejcir:
Holding until I manually verify the loop.
/hold

akrejcir:
The failing tests in CI are related to prometheus configuration in the testing cluster, they are not caused by this PR.

akrejcir:
Karel gave `/lgtm` on the commit with new functionality. The other commit only fixes CI issues unrelated to this PR.
I'm merging this PR.

/override tide

PR #382, opened 2022-06-09
add eventually in AfterEach in validator webhook tests which tries to delete template

ksimon1:
add eventually in AfterEach in validator webhook tests which tries to delete template

**Release note**:
```
NONE

```

### DIFF
--- a/tests/validator_test.go
+++ b/tests/validator_test.go
@@ -426,10 +426,11 @@ var _ = Describe("Template validator webhooks", func() {
 			waitForDeletion(client.ObjectKeyFromObject(vm), vm)
 		}
 		if template != nil {
-			err := apiClient.Delete(ctx, template)
-			if !errors.IsNotFound(err) {
-				Expect(err).ToNot(HaveOccurred(), "Failed to delete Template")
-			}
+			Eventually(func(g Gomega) {
+				if err := apiClient.Delete(ctx, template); err != nil {
+					g.Expect(errors.ReasonForError(err)).To(Equal(metav1.StatusReasonNotFound))
+				}
+			}, shortTimeout, time.Second).Should(Succeed(), "Template should be deleted")
 		}
 	})
 


akrejcir:
Why does Delete need to be called in a loop? If you want to wait until the template is deleted, you can call `Delete()` once and then call `waitForDeletion()` function. For example:
https://github.com/kubevirt/ssp-operator/blob/c6b9e70fa7f74d22da13205dec251bf0d2ef1c21/tests/misc_test.go#L217-L219

tests/validator_test.go
```diff
@@ -426,10 +426,13 @@ var _ = Describe("Template validator webhooks", func() {
...
-			if !errors.IsNotFound(err) {
-				Expect(err).ToNot(HaveOccurred(), "Failed to delete Template")
-			}
+			Eventually(func() bool {
+				err := apiClient.Delete(ctx, template)
+				if err != nil && !errors.IsNotFound(err) {
+					fmt.Println("template wasn't deleted: " + err.Error())
+				}
+				return errors.IsNotFound(err)
+			}, shortTimeout).Should(BeTrue(), "Template should be deleted")
```
akrejcir, line 435: This will print the log line on every iteration. Can you use this instead:
```go
Eventually(func(g Gomega) {
   if err := apiClient.Delete(ctx, template); err != nil {
      g.Expect(errors.ReasonForError(err)).To(Equal(metav1.StatusReasonNotFound))
   }
}, shortTimeout, time.Second).Should(Succeed(), "Template should be deleted")
```

It is a special form of the `Eventually`, but it is useful in this case. Hopefully the error message will be meaningful.
https://pkg.go.dev/github.com/onsi/gomega#Eventually

fossedihelm (commented):
Thanks @ksimon1. One question below.

tests/validator_test.go
```diff
@@ -426,10 +426,11 @@ var _ = Describe("Template validator webhooks", func() {
...
 		if template != nil {
-			err := apiClient.Delete(ctx, template)
-			if !errors.IsNotFound(err) {
-				Expect(err).ToNot(HaveOccurred(), "Failed to delete Template")
-			}
+			Eventually(func(g Gomega) {
+				if err := apiClient.Delete(ctx, template); err != nil {
+					g.Expect(errors.ReasonForError(err)).To(Equal(metav1.StatusReasonNotFound))
+				}
+			}, shortTimeout, time.Second).Should(Succeed(), "Template should be deleted")
```
fossedihelm, line 433: Can I ask why we need to try to delete the resource repeatedly? I would expect something similar to what happens for the vm just above this. Thanks

ksimon1:
I am overriding tests, as the errors are not related to this PR
/approve

tests/validator_test.go
fossedihelm: Thanks for the explanation

fossedihelm (approved):

PR #389, opened 2022-06-28
Revert infra_label for SSP alerts

assafad:
Reverts [#380](https://github.com/kubevirt/ssp-operator/pull/380). The design is not yet agreed as a standard for all operators alerts, so this label won't be used for now.

**Release note**:

```release-note
None
```


assafad:
Hi @sradco, can you please have a look?

akrejcir:
Can you write the reason for revert into the commit message, and mention the hash of the reverted commit?

ksimon1:
Is this PR going to be in 4.11 or only in 4.12?

sradco:
Currently the design for this label is still not ongoing and we may implement the solution in a different way. This is why we don;t want to add labels that we will not use.

/lgtm

akrejcir:
The failing CI tests are not related to this PR. They fail when importing DataVolumes.

akrejcir:
@assafad Do you want this PR to be backported to 4.11?

akrejcir:
@assafad told me that this needs to be backported.

/cherry-pick release-v0.15

akrejcir:
The original PR that added infra labels was not backported to 4.11, so this revert does not need to be.

PR #396, opened 2022-06-30
[backport release-v0.12] change a way ho golang is installed in dockerfile

ksimon1:
change a way ho golang is installed in dockerfile
Because golang 1.16 is no longer available in ubi
image, all tests are failing. This commit downloads golang from official go site.

**Release note**:
```
NONE
```


akrejcir:
Can you move the new common templates file to a different commit, and mention it in the PR's description?

PR #398, opened 2022-07-11
Add PR check for updated metrics docs

machadovilaca:
Signed-off-by: João Vilaça <jvilaca@redhat.com>

Validate in each PR that the metrics docs file is up to date

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -5,6 +5,19 @@ on:
   pull_request: {}
 
 jobs:
+  validate_metrics_docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Golang 1.18
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.18'
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Generate metrics docs
+        run: make generate-doc
+      - name: Validate no changes
+        run: git diff --exit-code
 
   build:
     runs-on: ubuntu-latest


.github/workflows/pull-request.yaml
```diff
@@ -5,6 +5,19 @@ on:
...
+      - name: Install Golang 1.18
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.18'
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Generate metrics docs
+        run: make generate-doc
+      - name: Validate no changes
+        run: git difftool -y --trust-exit-code --extcmd=./hack/diff-csv.sh
```
akrejcir, line 20: Why is the default `git diff` not enough?

machadovilaca: using extensions, in this case the ./hack/diff-csv.sh, allows us to ignore certain patterns that might have been changed on some compilation steps, such as the creation date on manifests

akrejcir: Ok, is it currently useful? I have tried searching for string `createdAt` in the repo, but it is not anywhere. What manifests do you mean?

machadovilaca: Ah, you are right, I tested `make bundle` and checked the `ClusterServiceVersion` generated in the `out/` directory, and unlike, for example, HCO, SSP does not create a `createdAt` annotation. Thank you for the warning, removing

akrejcir:
No need to run e2e tests, because this change is not related to them.

PR #399, opened 2022-07-13
Adding TLSSecurityProfile to CRD

borod108:
As part of supporting Openshift cluster-wide crypto policy,
adding required fields to SSP CRD.

```release-note
None
```


akrejcir:
Are we going to use this new field in a future PR? If yes, it would make sense to do the API change in the same PR as usage.

borod108:
@tiraboschi as Anrdrej requested, we will post this PR again with all the code for supporting the crypto policy, but can you just take a look and confirm the API is right?

borod108:
@akrejcir can we merge this, so that https://github.com/kubevirt/ssp-operator/pull/408 is cleaner?

akrejcir:
Why not close this PR and only keep https://github.com/kubevirt/ssp-operator/pull/408 ?
Anyway I'm taking a look...

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/go-logr/logr v1.2.3
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.19.0
-	github.com/openshift/api v0.0.0
+	github.com/openshift/api v3.9.0+incompatible
```
akrejcir, line 13: This change is not needed.

api/go.mod
```diff
@@ -3,6 +3,7 @@ module kubevirt.io/ssp-operator/api
 go 1.17
 
 require (
+	github.com/openshift/api v3.9.0+incompatible
```
akrejcir, line 6: Please use `v0.0.0` instead.

tiraboschi:
> @tiraboschi as Anrdrej requested, we will post this PR again with all the code for supporting the crypto policy, but can you just take a look and confirm the API is right?

Yes, the API looks fine

borod108:
> 

Just because it will look cleaner - this PR generates a lot of code, it is easier in my mind to have those PRs separate. I would prefer it this way, but it is up to you.

akrejcir:
Ok. Can you rebase? Then it can be merged.

api/go.mod
```diff
@@ -3,6 +3,7 @@ module kubevirt.io/ssp-operator/api
 go 1.18
 
 require (
+	github.com/openshift/api v0.0.0
```
akrejcir, line 6: I'm sorry for confusion, but I found out that you can simply `require` here the needed api version.
I thought there was some problem with modules so we needed to use this hack with `require` `v0.0.0` and then `replace` the needed version.

Can you change this line to:
```
github.com/openshift/api v0.0.0-20220124143425-d74727069f6f // release-4.10
```

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/go-logr/logr v1.2.3
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.19.0
-	github.com/openshift/api v0.0.0-20220124143425-d74727069f6f // release-4.10
+	github.com/openshift/api v0.0.0
```
akrejcir, line 13: No need to do this change.

go.mod
```diff
@@ -108,8 +108,8 @@ require (
 replace (
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
 	github.com/gorilla/websocket => github.com/gorilla/websocket v1.4.2
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20220124143425-d74727069f6f
```
akrejcir, line 111: This is not needed.

akrejcir (changes requested):

api/go.mod
borod108: when I did that (same for other comments) it actually failed go mod tidy :/

go.mod
```diff
@@ -110,6 +110,5 @@ replace (
 	github.com/gorilla/websocket => github.com/gorilla/websocket v1.4.2
 	k8s.io/client-go => k8s.io/client-go v0.24.2
 	k8s.io/cluster-bootstrap => k8s.io/cluster-bootstrap v0.23.5
-
```
akrejcir, line 113: This change is not needed, when nothing else is changed in this file.

akrejcir (commented):
I checked-out this code locally and run:
```bash
cd api
go mod vendor
go mod tidy

cd ..
go mod vendor
go mod tidy
```

And the file `api/go.sum` has some changes. Can you commit them?

go.mod
borod108: ok

ksimon1 (approved):

PR #401, opened 2022-07-19
Add missing webhook test and improve tests

akrejcir:
Added missing test for webhook functionality, and improved test code.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/webhook_test.go
+++ b/tests/webhook_test.go
@@ -142,6 +142,14 @@ var _ = Describe("Validation webhook", func() {
 						"created SSP CR with invalid template-validator placement fields")
 				})
 			})
+
+			It("[test_id:TODO] should fail when DataImportCronTemplate does not have a name", func() {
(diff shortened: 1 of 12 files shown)


akrejcir:
I will add more changes

/hold

ksimon1:
@akrejcir what is the status of this PR?

tests/webhook_test.go
```diff
@@ -164,7 +172,7 @@ var _ = Describe("Validation webhook", func() {
 					foundSsp = getSsp()
 					foundSsp.Spec.TemplateValidator.Placement = &placementAPIValidationValidPlacement
 					return apiClient.Update(ctx, foundSsp, client.DryRunAll)
-				}, twentySecondTimeout, time.Second).ShouldNot(HaveOccurred(), "failed to update SSP CR with valid template-validator placement fields")
```
ksimon1, line 167: Why do you remove this variable? Lets define it in env package

akrejcir: I thought the variable was not important. It has no other meaning than 20 seconds, so it can be inlined.

PR #402, opened 2022-07-20
tests: RHEL: wait for DataVolume import and increase timeout

akrejcir:
Waits specifically for DataVolume import, and increases timeout.

**Release note**:
```release-note
None
```


dominikholler:
@akrejcir thanks for the quick PR!

akrejcir:
Upgrade functests will fail until we release a new version.

PR #406, opened 2022-07-26
CI: Move versions to a separate file

akrejcir:
When a new release branch is created, the versions should be fixed.

**Release note**:
```release-note
None
```


akrejcir:
Upgrade functests will fail until we release a new version.

Issue #407, opened 2022-08-04
Comply with cluster-wide crypto policy


borod108:
SSP and Template Validator should support cluster-wide crypto policy. There should be an API that allows setting the desired TLS configuration, including ciphers and min TLS version.
More details on the general design can be found here:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1717

akrejcir:
The PRs that implement this are merged.

/close

PR #408, opened 2022-08-04
Support cluster-wide crypto policy

borod108:
All webhook servers will now follow the TLS configuration defined in the
SSP CR.
    
This addresses: https://github.com/kubevirt/ssp-operator/issues/407

Co-authored-by: Felix Matouschek <fmatouschek@redhat.com>
```release-note
SSP-Operator now supports TLS configuration through the SSP CR
```


internal/common/crypto_policy.go
```diff
@@ -0,0 +1,126 @@
...
+type SSPTLSOptions struct {
+	MinTLSVersion      string
+	OpenSSLCipherNames []string
+}
+
+func (s *SSPTLSOptions) IsEmpty() bool {
+	return len(s.OpenSSLCipherNames) == 0 && s.MinTLSVersion == ""
+}
+
+func (s *SSPTLSOptions) MinTLSVersionId() (minTlsVersionId uint16) {
```
0xFelix, line 26: The named return is not used and could be simplified to just the type that is returned.

internal/common/crypto_policy.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	return
+}
+
+func emptyTLSProfile() *SSPTLSOptions {
+	return &SSPTLSOptions{"", []string{}}
+}
+
+var (
+	setupLog = ctrl.Log.WithName("setup")
```
0xFelix, line 70: If `setupLog` is only used in `GetCiperSuitesAndMinTLSVersion` we should declare it just in this function.

internal/common/crypto_policy.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	namespace, err := GetOperatorNamespace(setupLog)
+	if err != nil {
+		return emptyTLSProfile(), err
+	}
+
+	var sspList v1beta1.SSPList
+	if err := apiReader.List(context.TODO(), &sspList, &client.ListOptions{Namespace: namespace}); err != nil {
+		return emptyTLSProfile(), fmt.Errorf("error getCipherSuitesAndMinTLSVersion: %w", err)
```
0xFelix, line 87: Do we need to wrap the error? Does it have a benefit?

main.go
```diff
@@ -25,18 +26,20 @@ import (
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	//"k8s.io/apimachinery/pkg/runtime/schema"
```
0xFelix, line 29: Commented imports?

main.go
```diff
@@ -82,6 +94,20 @@ func runPrometheusServer(metricsAddr string) {
 	}()
 }
 
+func getWebhookServer(sspTLSOptions common.SSPTLSOptions) *webhook.Server {
+	if sspTLSOptions.IsEmpty() {
```
0xFelix, line 98: What is the default behavior of other parts of SSP when there is no TLS configuration set? Should it run without TLS or not at all? At least all endpoints should behave the same. IIUC the template validator and the prometheus endpoint start with empty TLS config values while the SSP webhook server will refuse to run?

internal/common/crypto_policy.go
borod108: I would like to because it is a public method, what do you think?

main.go
borod108: Yes they are, tnx :P

internal/common/crypto_policy.go
borod108: I like to keep the log separate cause the next developer adding a log in some other method will find it easy, even if we use it only there currently, does it make sense? if you find it is not worth it - I will move it.

internal/common/crypto_policy.go
borod108: right!

main.go
borod108: we decided it will all run if the tls config is not provided. If the webhookserver is set to nil the manager will crate a default one.

internal/common/crypto_policy.go
akrejcir: Wrapping error is ok, but the text should provide more information about the function that failed. In this case something like: `"failed listing SSP resources: %w"`.
The calling function of `GetCipherSuitesAndMinTLSVersion()` will know that the failure was in that function, so text `"error getCipherSuitesAndMinTLSVersion"` is useless.

internal/common/crypto_policy.go
0xFelix: I would stick to declaring it only in the function because this makes it a "global" variable to the whole `common` package and it is not used anywhere else.

internal/common/crypto_policy.go
0xFelix: Ok, but then the other errors which can be returned should be wrapped too with a more explanative error message. You agree?

main.go
0xFelix: Ah, I forgot that an `nil` webhookserver creates a default one. So everything starts with some default TLS configuration values? That's fine.

internal/common/crypto_policy.go
akrejcir: Yes. If we want to wrap errors, then it makes sens to wrap all of them. Unless it is obvious where the error happened without wrapping. But in this case there are multiple calls to the `apiReader` and they could fail with simillar error messages, so wrapping is useful.

main.go
```diff
@@ -17,26 +17,24 @@ limitations under the License.
...
+	"kubevirt.io/ssp-operator/webhooks"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
-
-	"kubevirt.io/ssp-operator/controllers"
-	"kubevirt.io/ssp-operator/internal/common"
-	"kubevirt.io/ssp-operator/webhooks"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
```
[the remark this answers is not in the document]
akrejcir: If this happens when you edit the text or when you save, then it is an IDE setting. If it happens during `make generate`, then you probably removed an empty line separating groups of imports.

internal/operands/template-validator/reconcile.go
```diff
@@ -155,7 +155,13 @@ func reconcileDeployment(request *common.Request) (common.ReconcileResult, error
 	if request.IsSingleReplicaTopologyMode() && (numberOfReplicas > 1) {
 		numberOfReplicas = 1
 	}
-	deployment := newDeployment(request.Namespace, numberOfReplicas, image)
+
+	sspTLSOptions, err := common.NewSSPTLSOptions(request.Instance.Spec.TLSSecurityProfile, nil)
+	if err != nil {
+		return common.ReconcileResult{}, nil
```
akrejcir, line 161: Can this return the error?

internal/common/crypto_policy.go
```diff
@@ -0,0 +1,128 @@
...
+func NewSSPTLSOptions(tlsSecurityProfile *ocpv1.TLSSecurityProfile, logger *logr.Logger) (*SSPTLSOptions, error) {
+	ciphers, tlsProfile := selectCipherSuitesAndMinTLSVersion(tlsSecurityProfile)
+
+	if logger != nil {
+		logger.Info("Got Ciphers and tlsProfile:", "ciphers: ", ciphers, "tlsProfile: ", tlsProfile)
+	}
+
+	minVersion, err := tlsVersionToHumanReadable(tlsProfile)
+	if err != nil {
+		return nil, nil
```
akrejcir, line 51: Can this return the error?

internal/common/crypto_policy.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if len(sspList.Items) == 0 {
+		return nil, nil
+	}
+
+	ssp := sspList.Items[0]
+
+	sspTLSOptions, err := NewSSPTLSOptions(ssp.Spec.TLSSecurityProfile, &setupLog)
+	if err != nil {
+		return nil, nil
```
akrejcir, line 97: Can this return error?

internal/common/crypto_policy.go
```diff
@@ -0,0 +1,128 @@
...
+	case ocpv1.VersionTLS10:
+		return "1.0", nil
+	case ocpv1.VersionTLS11:
+		return "1.1", nil
+	case ocpv1.VersionTLS12:
+		return "1.2", nil
+	case ocpv1.VersionTLS13:
+		return "1.3", nil
+	default:
+		return "0", fmt.Errorf("invalid ocpv1.VersionTLS %v", version)
```
akrejcir, line 126: The first return value should be `""`, to be consistent.

internal/operands/template-validator/reconcile_test.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 var log = logf.Log.WithName("validator_operand")
+var emptySSPTLSConfig, err = common.NewSSPTLSOptions(nil, nil)
```
akrejcir, line 30: Instead of calling the function, you can directly create an empty object:
```go
var emptySSPTLSConfig = &common.SSPTLSOptions{}
```

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -171,4 +188,15 @@ func (ti *TLSInfo) CrateTlsConfig() *tls.Config {
 			return cert, nil
 		},
 	}
+
+	if !ti.sspTLSOptions.IsEmpty() {
+		tlsConfig.CipherSuites = ti.sspTLSOptions.CipherIDs()
+		minVersion, err := ti.sspTLSOptions.MinTLSVersionId()
+		if err != nil {
+			os.Exit(1)
```
akrejcir, line 196: Can you return the error instead?

go.mod
```diff
@@ -10,8 +10,9 @@ require (
 	github.com/go-logr/logr v1.2.3
 	github.com/onsi/ginkgo v1.16.5
 	github.com/onsi/gomega v1.19.0
-	github.com/openshift/api v0.0.0-20220124143425-d74727069f6f // release-4.10
+	github.com/openshift/api v0.0.0-20220525145417-ee5b62754c68 // release-4.10
```
akrejcir, line 13: This commit points to the `master` branch. Can you revert this change?

go.mod
```diff
@@ -87,7 +88,7 @@ require (
 	go.uber.org/zap v1.19.1 // indirect
 	golang.org/x/crypto v0.0.0-20220214200702-86341886e292 // indirect
 	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
-	golang.org/x/oauth2 v0.0.0-20220223155221-ee480838109b // indirect
+	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5 // indirect
```
akrejcir, line 91: Is this change needed?

akrejcir (commented):
/lgtm

akrejcir:
/ltgm

Please remove the `WIP` from title.

PR #409, opened 2022-08-08
Update CDI version to v1.53.0

akrejcir:
The new version introduced a required field to `DataSource` resource.

**Release note**:
```release-note
None
```


akrejcir:
The upgrade functests cannot pass, because they test upgrade from the latest released version, which does not work with latest CDI.

PR #410, opened 2022-08-09
add debug lines to DataImportCron template tests

ksimon1:
add debug lines to DataImportCron template tests
These debug lines should help to find out whats wrong with e2e
dataImportCron tests.

**Release note**:
```
NONE
```

### DIFF
--- a/tests/dataSources_test.go
+++ b/tests/dataSources_test.go
@@ -852,39 +852,44 @@ var _ = Describe("DataSources", func() {
 				}, shortTimeout, time.Second).Should(HaveKeyWithValue(cdiLabel, cronName))
 			})
 
-			It("[test_id:8112] should restore DataSource if DataImportCron removed from SSP CR", func() {
-				// Wait until DataImportCron imports PVC and changes data source
-				Eventually(func() bool {
-					cron := &cdiv1beta1.DataImportCron{}
-					Expect(apiClient.Get(ctx, dataImportCron.GetKey(), cron)).To(Succeed())
-
-					return cron.Status.LastImportTimestamp.IsZero()
-				}, timeout, time.Second).Should(BeFalse(), "DataImportCron did not finish importing.")
-
-				managedDataSource := &cdiv1beta1.DataSource{}
-				Expect(apiClient.Get(ctx, dataSource.GetKey(), managedDataSource)).To(Succeed())
-
-				// Remove the DataImportCron
-				updateSsp(func(foundSsp *ssp.SSP) {
-					foundSsp.Spec.CommonTemplates.DataImportCronTemplates = nil
+			Context("restore dataSource", func() {
+				JustAfterEach(func() {
+					logObject(dataImportCron.GetKey(), &cdiv1beta1.DataImportCron{})
 				})
-				waitUntilDeployed()
 
-				// Check if the DataSource has been reverted
-				revertedDataSource := &cdiv1beta1.DataSource{}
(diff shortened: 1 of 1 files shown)


ksimon1:
cc @akrejcir

tests/dataSources_test.go
```diff
@@ -857,7 +857,8 @@ var _ = Describe("DataSources", func() {
 				Eventually(func() bool {
 					cron := &cdiv1beta1.DataImportCron{}
 					Expect(apiClient.Get(ctx, dataImportCron.GetKey(), cron)).To(Succeed())
-
+					fmt.Println("dataImportCron " + dataImportCron.GetKey().String() + " status")
+					fmt.Printf("%#v \n", cron.Status)
```
akrejcir, line 861: This will be very noisy in log.
Instead, can you add a `JustAfterEach()` block, with a call to `logObject()` function which is exactly for this use case:
https://github.com/kubevirt/ssp-operator/blob/da413a117962247d41c1134b4914643a189543dd/tests/misc_test.go#L360

Here is an example: https://github.com/kubevirt/ssp-operator/blob/da413a117962247d41c1134b4914643a189543dd/tests/misc_test.go#L224-L241

It uses `GinkgoWriter` which will only output to log if a test fails.
The function is currently local in `misc_test.go`, but you can move it to `tests_common_test.go`.

ksimon1: If the tests will not fail, then there will be no record in log. It will log only in case of error. But your approach with JustAfterEach is better. I will update my PR.

akrejcir: Ok, I thought that a regular `fmt.Printf()` always outputs, even when test is successful. Good to know.

tests/dataSources_test.go
```diff
@@ -852,39 +852,45 @@ var _ = Describe("DataSources", func() {
...
 
-			It("[test_id:8112] should restore DataSource if DataImportCron removed from SSP CR", func() {
-				// Wait until DataImportCron imports PVC and changes data source
-				Eventually(func() bool {
-					cron := &cdiv1beta1.DataImportCron{}
-					Expect(apiClient.Get(ctx, dataImportCron.GetKey(), cron)).To(Succeed())
+			Context("restore dataSource", func() {
+				var cron *cdiv1beta1.DataImportCron
+				JustAfterEach(func() {
+					logObject(client.ObjectKeyFromObject(cron), &cdiv1beta1.DataImportCron{})
```
akrejcir, line 858: The `cron` temporary variable is nil at this point. You can get the key directly from `dataImportCron.GetKey()`:
```go
logObject(dataImportCron.GetKey(), &cdiv1beta1.DataImportCron{})
```

tests/dataSources_test.go
```diff
@@ -991,6 +997,8 @@ var _ = Describe("DataSources", func() {
 				Eventually(func() cdiv1beta1.DataVolumePhase {
 					foundDv := &cdiv1beta1.DataVolume{}
 					Expect(apiClient.Get(ctx, client.ObjectKeyFromObject(dataVolume), foundDv)).To(Succeed())
+					fmt.Println("datavolume " + client.ObjectKeyFromObject(dataVolume).String() + " status")
+					fmt.Printf("%#v \n", dataVolume.Status)
```
akrejcir, line 1001: The same approach can be used here.

ksimon1: does justAfterEach gets triggered if beforeEach fails?

akrejcir: Yes it does.

PR #417, opened 2022-09-27
Watch required CRDs, and restart operator if they are removed

akrejcir:
Previously, the operator did not start before all needed CRDs were installed. This meant that it was not possible to remove an SSP resource if CDI was not installed.

This PR adds a `CrdWatch` component, which restarts the operator, if some needed CRDs are deleted while it is running.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2122236

**Release note**:
```release-note
Operator properly handles missing CRDs.
```


akrejcir:
CI will fail until #418 is merged.

/hold

PR #418, opened 2022-09-28
Fix tests waiting for DVs to succeed

0xFelix:

This PR does two things:

1. Fix tests waiting for DVs to succeed in misc_test.go

CDI now garbage collects succeeded DVs by default, so relying on a DV
having status phase succeeded does not work anymore. When the importer
Pod of a PVC succeeded and the corresponding DV no longer exists it can
be treated as a successful import instead.

2. Disable CDI garbage collection for now in dataSources_test.go

CDI currently does not handle DataImportCrons properly when garbage
collecting DataVolumes, so as a temporary fix the garbage collection is
disabled via an annotation.

The following tests still fail:

- [[test_id:8112] should restore DataSource if DataImportCron removed from SSP CR](https://github.com/kubevirt/ssp-operator/blob/f5c7be21c7c11e0b4f29bb816ce914094384eb23/tests/dataSources_test.go#L861)
- [[test_id:8298] should restore DataSource, when CDI label is removed](https://github.com/kubevirt/ssp-operator/blob/f5c7be21c7c11e0b4f29bb816ce914094384eb23/tests/dataSources_test.go#L1118)

This is related to:
- https://github.com/kubevirt/containerized-data-importer/issues/2437
- https://bugzilla.redhat.com/show_bug.cgi?id=2130509

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/dataSources_test.go
+++ b/tests/dataSources_test.go
@@ -737,6 +737,8 @@ var _ = Describe("DataSources", func() {
 			pullMethod        = cdiv1beta1.RegistryPullNode
 			commonAnnotations = map[string]string{
 				"cdi.kubevirt.io/storage.bind.immediate.requested": "true",
+				// Remove this annotation once CDI handles DataImportCron and garbage collection properly
+				"cdi.kubevirt.io/storage.deleteAfterCompletion": "false",
 			}
 
 			cronTemplate   ssp.DataImportCronTemplate
--- a/tests/dataSources_test.go
+++ b/tests/dataSources_test.go
@@ -859,7 +859,7 @@ var _ = Describe("DataSources", func() {
 					logObject(dataImportCron.GetKey(), &cdiv1beta1.DataImportCron{})
 				})
 
-				It("[test_id:8112] should restore DataSource if DataImportCron removed from SSP CR", func() {
+				PIt("[QUARANTINE][test_id:8112] should restore DataSource if DataImportCron removed from SSP CR", func() {
 					// Wait until DataImportCron imports PVC and changes data source
 					Eventually(func() bool {
 						cron := &cdiv1beta1.DataImportCron{}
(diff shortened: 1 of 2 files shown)


tests/dataSources_test.go
```diff
@@ -737,6 +737,7 @@ var _ = Describe("DataSources", func() {
 			pullMethod        = cdiv1beta1.RegistryPullNode
 			commonAnnotations = map[string]string{
 				"cdi.kubevirt.io/storage.bind.immediate.requested": "true",
+				"cdi.kubevirt.io/storage.deleteAfterCompletion":    "false",
```
akalenyu, line 740: maybe include a note here that we can get rid of this once we fix https://bugzilla.redhat.com/show_bug.cgi?id=2130509?

0xFelix:
This will be the fix for the two remaining tests: https://github.com/kubevirt/containerized-data-importer/pull/2439

arnongilboa:
@0xFelix note the DataSource DV GC issue is solved in https://github.com/kubevirt/containerized-data-importer/pull/2441, so you either wait or it or revert your relevant cvhange when it's in.

tests/dataSources_test.go
```diff
@@ -857,7 +859,7 @@ var _ = Describe("DataSources", func() {
 					logObject(dataImportCron.GetKey(), &cdiv1beta1.DataImportCron{})
 				})
 
-				It("[test_id:8112] should restore DataSource if DataImportCron removed from SSP CR", func() {
+				It("[QUARANTINE][test_id:8112] should restore DataSource if DataImportCron removed from SSP CR", func() {
```
akrejcir, line 862: I'm not sure `[QUARANTINE]` has any effect on tests in ssp-operator. You can mark the test as pending to disable it (use `PIt()`).

PR #425, opened 2022-10-24
Add e2e tests for total_restored_common_templates metric

machadovilaca:
Signed-off-by: João Vilaça <jvilaca@redhat.com>

This PR adds an e2e test to ensure the `total_restored_common_templates` metric is not incremented if a template is created/edited during and SSP operator upgrade

**Release note**:

```release-note
Add e2e tests for total_restored_common_templates metric
```

### DIFF
--- a/tests/metrics_test.go
+++ b/tests/metrics_test.go
@@ -2,15 +2,21 @@ package tests
 
 import (
 	"fmt"
+	templatev1 "github.com/openshift/api/template/v1"
+	"kubevirt.io/ssp-operator/tests/env"
 	"net/http"
 	"reflect"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	apps "k8s.io/api/apps/v1"
 	rbac "k8s.io/api/rbac/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/operands/metrics"
@@ -24,6 +30,13 @@ func mergeMaps(maps ...map[string]string) map[string]string {
 		}
 	}
 	return targetMap
+}
+
+func getDeployment(name string, namespace string) *apps.Deployment {
+	deployment := &apps.Deployment{}
+	err := apiClient.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, deployment)
+	Expect(err).ToNot(HaveOccurred())
+	return deployment
 }
 
 var _ = Describe("Metrics", func() {
@@ -153,6 +166,88 @@ var _ = Describe("Metrics", func() {
 		)
 	})
 
+	Context("SSP metrics", func() {
+		var originalVersion string
+
+		waitForVersionUpdate := func(version string) {
+			Eventually(func() error {
+				rsList := &apps.ReplicaSetList{}
+				err := apiClient.List(ctx, rsList, client.InNamespace(sspDeploymentNamespace),
+					client.MatchingLabels{"name": "ssp-operator"})
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, rs := range rsList.Items {
+					for _, env := range rs.Spec.Template.Spec.Containers[0].Env {
+						if env.Name == common.OperatorVersionKey && env.Value == version && rs.Status.ReadyReplicas == *rs.Spec.Replicas {
+							return nil
+						}
+					}
+				}
+
+				return fmt.Errorf("operator version not yet updated")
+			}, env.ShortTimeout(), time.Second).Should(Succeed())
+		}
+
+		updateVersion := func(version string) {
+			Eventually(func() error {
+				deployment := getDeployment(sspDeploymentName, sspDeploymentNamespace)
+
+				envs := deployment.Spec.Template.Spec.Containers[0].Env
+
+				for i, env := range envs {
+					if env.Name == common.OperatorVersionKey {
+						envs[i].Value = version
+						break
+					}
+				}
+
+				return apiClient.Update(ctx, deployment)
+			}, env.ShortTimeout(), time.Second).ShouldNot(HaveOccurred())
+
+			waitForVersionUpdate(version)
+		}
+
+		BeforeEach(func() {
+			strategy.SkipSspUpdateTestsIfNeeded()
+
+			By("saving original version")
+			originalVersion = ""
+			deployment := getDeployment(sspDeploymentName, sspDeploymentNamespace)
+			envs := deployment.Spec.Template.Spec.Containers[0].Env
+			for _, env := range envs {
+				if env.Name == common.OperatorVersionKey {
+					originalVersion = env.Value
+				}
+			}
+
+			waitUntilDeployed()
+		})
+
+		AfterEach(func() {
+			updateVersion(originalVersion)
+			strategy.RevertToOriginalSspCr()
+		})
+
+		It("[test_id:TODO]should not increment total_restored_common_templates during upgrades", func() {
+			testTemplate := createTestTemplate()
+
+			pauseSsp()
+			version := "test-" + rand.String(5)
+			updateVersion(version)
+
+			template := &templatev1.Template{}
+			Expect(apiClient.Get(ctx, testTemplate.GetKey(), template)).To(Succeed())
+			testTemplate.Update(template)
+			Expect(apiClient.Update(ctx, template)).To(Succeed())
+
+			unpauseSsp()
+			waitUntilDeployed()
+
+			newRestoredCount := totalRestoredTemplatesCount()
+			Expect(newRestoredCount).To(BeZero())
+		})
+	})
+
 	Context("SSP alerts rules", func() {
 		var promRule promv1.PrometheusRule
 
@@ -160,7 +255,7 @@ var _ = Describe("Metrics", func() {
 			Expect(apiClient.Get(ctx, prometheusRuleRes.GetKey(), &promRule)).To(Succeed())
 		})
 
-		It("[test_id:7851]should have all the requried annotations", func() {
+		It("[test_id:7851]should have all the required annotations", func() {
 			for _, group := range promRule.Spec.Groups {
 				for _, rule := range group.Rules {
 					if rule.Alert != "" {
@@ -176,7 +271,7 @@ var _ = Describe("Metrics", func() {
 			}
 		})
 
-		It("[test_id:8955]should have all the requried labels", func() {
+		It("[test_id:8955]should have all the required labels", func() {
 			for _, group := range promRule.Spec.Groups {
 				for _, rule := range group.Rules {
 					if rule.Alert != "" {


tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
 		)
 	})
 
+	Context("SSP metrics", func() {
+		getDeployment := func(name string, namespace string) *apps.Deployment {
```
akrejcir, line 162: This function is only used in one place, so it can be inlined.

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			deployment := &apps.Deployment{}
+			err := apiClient.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, deployment)
+			Expect(err).ToNot(HaveOccurred())
+			return deployment
+		}
+
+		getOperatorPod := func() (k8sv1.Pod, uint16) {
+			pods, port := operatorPodsWithMetricsPort()
+			return pods[0], port
```
akrejcir, line 171: When the operator starts, it does a leader election to determine which pod should be active. The inactive pods will not update the metrics. So the fist returned pod may be inactive, and the test would fail.

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			for _, env := range envs {
+				if env.Name != common.OperatorVersionKey {
+					newEnvs = append(newEnvs, env)
+				}
+			}
+
+			newEnvs = append(newEnvs, k8sv1.EnvVar{Name: common.OperatorVersionKey, Value: version})
+			deployment.Spec.Template.Spec.Containers[0].Env = newEnvs
+
+			err := apiClient.Update(ctx, deployment)
```
akrejcir, line 189: This update may fail, because of a concurrent update operation. Can you wrap the update logic in an `Eventually()` ?

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+		}
+
+		BeforeEach(func() {
```
akrejcir, line 193: These tests modify the SSP deployment, so we need to disable them if modification is not allowed.
Can you add `strategy.SkipSspUpdateTestsIfNeeded()` at the beginning of `BeforeEach()`?

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			updateVersion("devel")
```
akrejcir, line 194: The SSP version may not always be `devel`. The functional tests can be run on a deployment created by HCO, where the version will have some other value.
Instead of a hardcoded update, can you get the current version and store it in a variable, and then restore it back in `AfterEach()`?

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			waitUntilDeployed()
+		})
+
+		AfterEach(func() {
+			updateVersion("devel")
+		})
```
akrejcir, line 200: One of the test uses `pauseSsp()`  function which modifies the SSP CR, so we need to restore it to its original definition in `AfterEach()`. Can you call `strategy.RevertToOriginalSspCr()` here?

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+
+		It("should increment total_restored_common_templates when not in upgrade", func() {
```
akrejcir, line 202: This test already exists here:
https://github.com/kubevirt/ssp-operator/blob/b3a78fcf04bfe1cb0c21fbb5895138ee4e7bb603/tests/commonTemplates_test.go#L193-L195

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			By("creating test template")
+			testTemplate := createTestTemplate()
+
+			By("getting restored count after creating test template")
+			expectRestoreAfterUpdate(&testTemplate)
+			newRestoredCount := intMetricValue("total_restored_common_templates", port, &sspPod)
+			Expect(newRestoredCount).To(Equal(restoredCount + 1))
+		})
+
+		It("should not increment total_restored_common_templates during upgrades", func() {
```
akrejcir, line 217: The test is missing `[test_id:]`. Once this PR is good to merge, we can ask QE for an ID. In the meantime, can you add `[test_id:TODO]` here, so we know it is missing?

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			By("getting initial restored count")
+			restoredCount := intMetricValue("total_restored_common_templates", port, &sspPod)
+
+			By("updating SSP operator version and creating test template")
+			pauseSsp()
+
+			testTemplate := createTestTemplate()
+
+			updateVersion("test-" + rand.String(5))
+			waitUntilDeployed()
```
akrejcir, line 229: At this point the operator is paused, so the SSP will never be in `Deployed` state. Waiting for it does not make sense.

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			time.Sleep(30 * time.Second)
```
akrejcir, line 230: Please avoid `time.Sleep()` in tests. Can we instead wait for something?

tests/metrics_test.go
```diff
@@ -153,14 +158,95 @@ var _ = Describe("Metrics", func() {
...
+			sspPod, port = getOperatorPod()
+
+			unpauseSsp()
+
+			By("getting restored count after updating SSP operator version")
+			expectRestoreAfterUpdate(&testTemplate)
```
akrejcir, line 236: This template update is executed after SSP has finished updating and is in `Deployed` state, so the metric will probably be increased.

akrejcir (changes requested):
The test does not work as expected.

akrejcir:
Can you also add more text to commit message and PR's description?

tests/metrics_test.go
machadovilaca: using it twice after changes

tests/metrics_test.go
machadovilaca: I'm now iterating the list to make sure I get the value

tests/metrics_test.go
machadovilaca: done

tests/metrics_test.go
machadovilaca: done

tests/metrics_test.go
machadovilaca: done

tests/metrics_test.go
machadovilaca: removed

tests/metrics_test.go
machadovilaca: added

tests/metrics_test.go
machadovilaca: removed

tests/metrics_test.go
machadovilaca: changed

tests/metrics_test.go
machadovilaca: Since it's paused I think it's only updated with the version update

akrejcir:
The CI is currently broken. No need to retest.

tests/metrics_test.go
```diff
@@ -153,14 +159,119 @@ var _ = Describe("Metrics", func() {
...
+			deployment.Spec.Template.Spec.Containers[0].Env = newEnvs
+
+			Eventually(func() error {
+				return apiClient.Update(ctx, deployment)
+			}, 1*time.Minute, time.Second).ShouldNot(HaveOccurred())
+		}
+
+		waitForVersionUpdate := func(version string) {
+			Eventually(func() error {
+				operatorPods, _ := operatorPodsWithMetricsPort()
```
akrejcir, line 216: I think you don't need to check pods here. It should be enough to test status of deployment.
These fields:
- `status.updatedReplicas` - number of running pods with the desired spec
- `status.readyReplicas` - number of pods with Ready condition

If both of these values are equal to `spec.replicas`, we know that all pods have been updated and are ready.

tests/metrics_test.go
```diff
@@ -153,14 +159,119 @@ var _ = Describe("Metrics", func() {
...
+			waitUntilDeployed()
+		})
+
+		AfterEach(func() {
+			updateVersion(originalVersion)
+			waitForVersionUpdate(originalVersion)
+			strategy.RevertToOriginalSspCr()
+		})
+
+		It("[test_id:TODO]should not increment total_restored_common_templates during upgrades", func() {
```
akrejcir, line 248: This test does not work as expected. I think these should be the steps:
- `pauseSsp()` - this pauses the reconciliation, and so any updated templates will not be restored
- Change the version environment variable on the SSP deployment - this simulates version update
- Update a template
- `unpauseSsp()` - when operator is unpaused it detects that the template was changed, but because it is in the middle of version update, it should not increase the metric.
- wait until SSP is in `deployed` state - then we know that version update is finished
- test that the metric was not increased.

tests/metrics_test.go
```diff
@@ -153,14 +159,119 @@ var _ = Describe("Metrics", func() {
...
+		var originalVersion string
+
+		getDeployment := func(name string, namespace string) *apps.Deployment {
+			deployment := &apps.Deployment{}
+			err := apiClient.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, deployment)
+			Expect(err).ToNot(HaveOccurred())
+			return deployment
+		}
+
+		totalRestoredTemplatesCount := func() int {
```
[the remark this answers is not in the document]
machadovilaca: removed

tests/metrics_test.go
```diff
@@ -153,14 +159,119 @@ var _ = Describe("Metrics", func() {
...
+			sum := 0
+			operatorPods, operatorMetricsPort := operatorPodsWithMetricsPort()
+			for _, sspOperator := range operatorPods {
+				sum += intMetricValue("total_restored_common_templates", operatorMetricsPort, &sspOperator)
+			}
+			return sum
+		}
+
+		saveOriginalVersion := func() {
```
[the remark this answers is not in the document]
machadovilaca: done

tests/metrics_test.go
```diff
@@ -153,14 +159,119 @@ var _ = Describe("Metrics", func() {
...
+			for _, env := range envs {
+				if env.Name != common.OperatorVersionKey {
+					newEnvs = append(newEnvs, env)
+				}
+			}
+
+			newEnvs = append(newEnvs, k8sv1.EnvVar{Name: common.OperatorVersionKey, Value: version})
+			deployment.Spec.Template.Spec.Containers[0].Env = newEnvs
+
+			Eventually(func() error {
```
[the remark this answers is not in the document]
machadovilaca: done

tests/metrics_test.go
machadovilaca: this is what is being done, except for the waiting ` until SSP is in deployed state` I think. Is this comment just to add that step?

tests/metrics_test.go
machadovilaca: does not work as is, considers the previous deployment as update and succeeds instantly

tests/metrics_test.go
machadovilaca: updated

tests/metrics_test.go
0xFelix: IIUC a new ReplicaSet should be created after the update, right? What if you check the ReplicaSet for Desired == Ready Pods?

machadovilaca: updated

tests/metrics_test.go
```diff
@@ -153,14 +158,104 @@ var _ = Describe("Metrics", func() {
...
+
+		updateVersion := func(version string) {
+			Eventually(func() error {
+				deployment := getDeployment(sspDeploymentName, sspDeploymentNamespace)
+
+				envs := deployment.Spec.Template.Spec.Containers[0].Env
+				var newEnvs []k8sv1.EnvVar
+
+				for _, env := range envs {
+					if env.Name != common.OperatorVersionKey {
```
0xFelix, line 179: Is there a reason why the existing `EnvVar` cannot be changed in-place?

tests/metrics_test.go
```diff
@@ -153,14 +158,104 @@ var _ = Describe("Metrics", func() {
...
+
+				newEnvs = append(newEnvs, k8sv1.EnvVar{Name: common.OperatorVersionKey, Value: version})
+				deployment.Spec.Template.Spec.Containers[0].Env = newEnvs
+
+				return apiClient.Update(ctx, deployment)
+			}, 1*time.Minute, time.Second).ShouldNot(HaveOccurred())
+		}
+
+		waitForVersionUpdate := func(version string) {
+			Eventually(func() error {
```
0xFelix, line 192: What about returning a bool instead of an error, to keep it simple?

tests/metrics_test.go
```diff
@@ -153,14 +158,104 @@ var _ = Describe("Metrics", func() {
...
+						if env.Name == common.OperatorVersionKey && env.Value == version {
+							if rs.Status.ReadyReplicas == *rs.Spec.Replicas {
+								return nil
+							}
+						}
+					}
+				}
+
+				return fmt.Errorf("no replicaset with version %s found", version)
+			}, 1*time.Minute, time.Second).ShouldNot(HaveOccurred())
```
0xFelix, line 209: If possible avoid double negation.

tests/metrics_test.go
```diff
@@ -153,14 +158,104 @@ var _ = Describe("Metrics", func() {
...
+		waitForVersionUpdate := func(version string) {
+			Eventually(func() error {
+				rsList := &apps.ReplicaSetList{}
+				err := apiClient.List(ctx, rsList, client.InNamespace(sspDeploymentNamespace),
+					client.MatchingLabels{"name": "ssp-operator"})
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, rs := range rsList.Items {
+					for _, env := range rs.Spec.Template.Spec.Containers[0].Env {
+						if env.Name == common.OperatorVersionKey && env.Value == version {
```
0xFelix, line 200: This can be reduced to a single if-statement.

tests/metrics_test.go
machadovilaca: how so?

tests/metrics_test.go
machadovilaca: i think it makes it easier to understand to have the message from the error

tests/metrics_test.go
machadovilaca: done

tests/metrics_test.go
machadovilaca: done

machadovilaca: i think it is clearer and more easier to understand in case of failure this way

0xFelix: Then the error message is misleading, as it can not only occur when the RS was not found but also when the `rs.Status.ReadyReplicas == *rs.Spec.Replicas` was not met.

tests/metrics_test.go
0xFelix: I meant like this:

```
for i, env := range envs {
    if env.Name == common.OperatorVersionKey {
        envs[i].Version = version
        break
    }
}
```

Also, does the update need the surrounding `Eventually`?

tests/metrics_test.go
machadovilaca: message updated

tests/metrics_test.go
akrejcir: I meant that the test should look like this:
```go
It("[test_id:TODO]should not increment total_restored_common_templates during upgrades", func() {
	testTemplate := createTestTemplate()

	pauseSsp()
	version := "test-" + rand.String(5)
	updateVersion(version)
	waitForVersionUpdate(version)

	template := &templatev1.Template{}
	Expect(apiClient.Get(ctx, testTemplate.GetKey(), template)).To(Succeed())
	testTemplate.Update(template)
	Expect(apiClient.Update(ctx, template)).To(Succeed())

	unpauseSsp()
	waitUntilDeployed()

	newRestoredCount := totalRestoredTemplatesCount()
	Expect(newRestoredCount).To(BeZero())
})
```

The current test does not do what is expected.

tests/metrics_test.go
```diff
@@ -153,14 +157,99 @@ var _ = Describe("Metrics", func() {
...
+						envs[i].Value = version
+						break
+					}
+				}
+
+				return apiClient.Update(ctx, deployment)
+			}, 1*time.Minute, time.Second).ShouldNot(HaveOccurred())
+		}
+
+		waitForVersionUpdate := func(version string) {
```
akrejcir, line 187: This functionality can be moved into `updateVersion()` function, because it is always useful to wait until containers with new version are deployed.

tests/metrics_test.go
```diff
@@ -153,14 +157,99 @@ var _ = Describe("Metrics", func() {
...
+			Eventually(func() error {
+				rsList := &apps.ReplicaSetList{}
+				err := apiClient.List(ctx, rsList, client.InNamespace(sspDeploymentNamespace),
+					client.MatchingLabels{"name": "ssp-operator"})
+				Expect(err).ToNot(HaveOccurred())
+
+				for _, rs := range rsList.Items {
+					for _, env := range rs.Spec.Template.Spec.Containers[0].Env {
+						if env.Name == common.OperatorVersionKey && env.Value == version && rs.Status.ReadyReplicas == *rs.Spec.Replicas {
+							return nil
```
akrejcir, line 197: It is better to wait until all pods in the ReplicaSet are updated to new version, not only a single pod.

tests/metrics_test.go
```diff
@@ -153,14 +157,99 @@ var _ = Describe("Metrics", func() {
...
+
+				for i, env := range envs {
+					if env.Name == common.OperatorVersionKey {
+						envs[i].Value = version
+						break
+					}
+				}
+
+				return apiClient.Update(ctx, deployment)
+			}, 1*time.Minute, time.Second).ShouldNot(HaveOccurred())
```
akrejcir, line 184: Instead of 1 minute timeout, can you use `env.ShortTimeout()` ?

tests/metrics_test.go
```diff
@@ -153,14 +157,99 @@ var _ = Describe("Metrics", func() {
...
+				for _, rs := range rsList.Items {
+					for _, env := range rs.Spec.Template.Spec.Containers[0].Env {
+						if env.Name == common.OperatorVersionKey && env.Value == version && rs.Status.ReadyReplicas == *rs.Spec.Replicas {
+							return nil
+						}
+					}
+				}
+
+				return fmt.Errorf("operator version not yet updated")
+			}, 1*time.Minute, time.Second).Should(Succeed())
```
akrejcir, line 203: Similarly, can you use `env.ShortTimeout()` ?

tests/metrics_test.go
```diff
@@ -153,14 +157,99 @@ var _ = Describe("Metrics", func() {
 		)
 	})
 
+	Context("SSP metrics", func() {
+		var originalVersion string
+
+		getDeployment := func(name string, namespace string) *apps.Deployment {
```
akrejcir, line 163: This can be a free function, because it does not depend on any variables inside the scope.

akrejcir (changes requested):

tests/metrics_test.go
akrejcir: It's a good practice to have `Eventually` around update, because the resource in the cluster may be concurrently updated by another component. Then update would fail.

tests/metrics_test.go
machadovilaca: moved

tests/metrics_test.go
machadovilaca: done

tests/metrics_test.go
machadovilaca: done

tests/metrics_test.go
machadovilaca: moved

tests/metrics_test.go
akrejcir: Right, sorry, I missed that.

tests/metrics_test.go
machadovilaca: mhh, did't get it, updated

PR #427, opened 2022-10-26
add dockerfile for ci builds

ksimon1:
The new go 1.19 ci image does not contain packages which ssp operator needs for tests. This commit adds new dockerfile, which installs all necessary packages needed for ssp operator

**Release note**:
```
NONE
```

### DIFF
--- a/ci-builder/Dockerfile
+++ b/ci-builder/Dockerfile
@@ -1,0 +1,12 @@
+# This Dockerfile is used in openshift CI
+FROM quay.io/fedora/fedora:37
+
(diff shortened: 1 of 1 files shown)


ci-builder/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# This Dockerfile is used in openshift CI
+FROM quay.io/fedora/fedora:latest
```
akrejcir, line 2: Can you fix the fedora version, instead of `latest`?

PR #429, opened 2022-10-27
update controller-gen and kustomize

ksimon1:
update controller-gen and kustomize

**Release note**:
```
NONE
```


akrejcir:
`preserveUnknownFields` is not needed for us. From the [documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning), the field is useful when we want to update CRD definition from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`. But SSP operator uses `v1` CRD since the first released version, so there is no update.

It is safe to just remove the `preserveUnknownFields=false`.

PR #437, opened 2022-11-03
Allow running the SSP Operator on plain kubernetes

nunnatsa:
SSP can't run on a plain kubernetes, because it requires the `Infrastructure` kind to be present on the cluster.

When deploying KubeVirt with HCO using OLM on plain kubernetes, the deployment is never completed because SSP operator never becomes ready. More details and full description of the issue can be found here: https://github.com/kubevirt/hyperconverged-cluster-operator/issues/2129

This PR disables the reconcilers and all the watches, when running in a plain K8s cluster, instead of killing the process.

That way, the SSP operator is runing and responding to the health and ready checks, but does nothing else. The result is that SSP is still not working on plain kubernetest, but the OLM deployment is successfully completed.

Fixes kubevirt/hyperconverged-cluster-operator#2129

Signed-off-by: Nahshon Unna-Tsameret <nunnatsa@redhat.com>

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/setup.go
+++ b/controllers/setup.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"github.com/go-logr/logr"
 	"path/filepath"
 
 	"kubevirt.io/ssp-operator/internal/common"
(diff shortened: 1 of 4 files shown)


0xFelix:
Maybe you could actively check if it is running on OpenShift instead of relying on `GetInfrastructureTopology` to fail / modifying the `topologyMode` variabele / `TopologyMode` field, like here?

https://github.com/kubevirt/tekton-tasks-operator/blob/ac169d3372d33f9d20fe93af70b8eba06e53f9ce/pkg/tekton-bundle/tekton-bundle.go#L102

nunnatsa:
> Maybe you could actively check if it is running on OpenShift instead of relying on `GetInfrastructureTopology` to fail / modifying the `topologyMode` variabele / `TopologyMode` field, like here?
> 
> https://github.com/kubevirt/tekton-tasks-operator/blob/ac169d3372d33f9d20fe93af70b8eba06e53f9ce/pkg/tekton-bundle/tekton-bundle.go#L102

Thanks @0xFelix. I'm not so sure about it, because we're using the same algorithm here. In the `runningOnOpenshift` function does almost the same thing: it is trying to read the version CR, and check if it's supported, as we're doing here, for the `Infrastructure` CR, but here we need the `InfrastructureTopology` value, if exists. I'm not sure I see the big different.

0xFelix:
The algorithm might be the same but from looking at the code it's not obvious why ssp should behave differently when `len(r.topologyMode) > 0`. IMO two things are mixed up here. Separation of getting the infrastructure topology and disabling the reconciler would greatly improve the understanding of the code.

controllers/setup.go
```diff
@@ -36,64 +37,75 @@ func CreateAndStartReconciler(ctx context.Context, mgr controllerruntime.Manager
...
 		return err
 	}
 
-	if missingCrds := crdWatch.MissingCrds(); len(missingCrds) > 0 {
-		mgr.GetLogger().Error(nil, "Some required crds are missing. The operator will not create any new resources.",
-			"missingCrds", missingCrds,
-		)
-	}
+	if runningOnOpenShift {
+		var requiredCrds []string
```
0xFelix, line 50: Maybe it makes sense to move the body of the if clause to its own function?

0xFelix (commented):
This approach seems much cleaner to me, just one question left.

controllers/setup.go
nunnatsa: Thanks @0xFelix - done

0xFelix:
Thanks! Can you update the commit message / pr description accordingly please? And another question: What about the status conditions? Are they set in some way?

nunnatsa:
> Can you update the commit message / pr description accordingly please? 

Done

0xFelix:
Thank you! I'm waiting for the CI to succeed before approving it.

akrejcir:
The CI is currently broken.

akrejcir:
/lgtm

I will retrigger CI once it is fixed.

akrejcir:
@nunnatsa Do you want this in `4.12`?

nunnatsa:
yes, it would be nice. But it also needs #443 to be installed with OLM.

akrejcir:
Ok

/cherry-pick release-v0.16

PR #448, opened 2022-11-18
crypto_policy_test.go: Do not hardcode SSP deployment NS

0xFelix:

Use the appropriate getter instead of hardcoding the deployment namespace to 'kubevirt'.

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/crypto_policy_test.go
+++ b/tests/crypto_policy_test.go
@@ -183,10 +183,10 @@ func tryToAccessEndpoint(pod core.Pod, subpath string, port uint16, tlsConfig cl
 		subpath = "/" + subpath
 	}
 
-	// ssp-webhook-service.kubevirt.svc is used to access the endpoints we are testing.
+	// ssp-webhook-service.${deploymentNamespace}.svc is used to access the endpoints we are testing.
 	// It is used here for the metrics as well for simplicity, as it is the CN in the ca_cert
 	// and the metrics just sit on a different port on the same pod.
-	attemptedUrl = fmt.Sprintf("https://ssp-webhook-service.kubevirt.svc:%d%s", port, subpath)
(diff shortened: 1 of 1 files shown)


tests/crypto_policy_test.go
```diff
@@ -183,10 +183,10 @@ func tryToAccessEndpoint(pod core.Pod, subpath string, port uint16, tlsConfig cl
 		subpath = "/" + subpath
 	}
 
-	// ssp-webhook-service.kubevirt.svc is used to access the endpoints we are testing.
+	// ssp-webhook-service.${deploymentNamespace}.svc is used to access the endpoints we are testing.
 	// It is used here for the metrics as well for simplicity, as it is the CN in the ca_cert
 	// and the metrics just sit on a different port on the same pod.
-	attemptedUrl = fmt.Sprintf("https://ssp-webhook-service.kubevirt.svc:%d%s", port, subpath)
+	attemptedUrl = fmt.Sprintf("https://ssp-webhook-service.%s.svc:%d%s", strategy.GetSSPDeploymentNameSpace(), port, subpath)
```
dominikholler, line 189: I am unsure about the expected URL in OpenShift, is it
ssp-operator-service.openshift-cnv.svc
or
ssp-webhook-service.openshift-cnv.svc
?

0xFelix: `ssp-webhook-service` is the only `Service` created by the manifests found in `config/webhook`, so `ssp-webhook-service.${deploymentNamespace}.svc` should be right.

lyarwood:
/lgtm

@ksimon1 do you want to approve this as @akrejcir is out today?

tests/crypto_policy_test.go
akrejcir: I think when the webhook is created by OLM, the service may have a different name. But I haven't tried it.

0xFelix: Yes, this is also an issue. I will introduce another getter similar to the NS shortly.

PR #468, opened 2022-12-13
kubevirtci: Move to _cluster-up

lyarwood:

Avoiding the situation where make thinks there's nothing to do because the directory of the same name as the target exists:

```
$ make cluster-up
make: 'cluster-up' is up to date.
```

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,6 @@ bundle/*
 cover.out
 bundle.Dockerfile
 config/manager/manager.yaml
-cluster-up/
+_cluster-up/
 _kubevirt/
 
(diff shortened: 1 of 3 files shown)


0xFelix (commented):
I would have suggested renaming to `_cluster-up` and additionally declaring the `cluster-up` target as `.PHONY`.

lyarwood:
@dollierp @0xFelix ACK thanks both I've added a commit marking the targets as phonies.

0xFelix:
Thanks!

/approve
/lgtm

PR #474, opened 2022-12-27
Add `operator_health_impact` label to SSP alerts

assafad:

This PR adds to each of SSP alerts a label named `operator_health_impact`, which indicates how the alerts impact the operator health. This will enable us filtering alerts based on their impact on the operator health, and to later have an operator health metric, which will tell the user what is the aggregated operator health.

This label gets one of the following values:
- `critical` indicates that the alert fires due to an issue that impacts the operator's functionality badly, and an action must be taken immediately in order to solve it.
-  `warning` indicates that the alert fires due to an issue that soon might impact the operator's functionality badly, and the user should be warned in order to solve the issue.
- `none` indicates that the alert fires due to an issue that doesn't affect the operator health, and in most cases is related to the workload rather than the operator itself.

https://issues.redhat.com/browse/CNV-18923

**Release note**:

```release-note
None
```

### DIFF
--- a/internal/operands/metrics/resources.go
+++ b/internal/operands/metrics/resources.go
@@ -13,6 +13,7 @@ const (
 	MonitorNamespace             = "openshift-monitoring"
 	runbookURLBasePath           = "https://kubevirt.io/monitoring/runbooks/"
 	severityAlertLabelKey        = "severity"
+	healthImpactAlertLabelKey    = "operator_health_impact"
 	partOfAlertLabelKey          = "kubernetes_operator_part_of"
 	partOfAlertLabelValue        = "kubevirt"
 	componentAlertLabelKey       = "kubernetes_operator_component"
@@ -85,9 +86,10 @@ var alertRulesList = []promv1.Rule{
 			"runbook_url": runbookURLBasePath + "SSPOperatorDown",
 		},
 		Labels: map[string]string{
-			severityAlertLabelKey:  "critical",
-			partOfAlertLabelKey:    partOfAlertLabelValue,
-			componentAlertLabelKey: componentAlertLabelValue,
+			severityAlertLabelKey:     "critical",
+			healthImpactAlertLabelKey: "critical",
+			partOfAlertLabelKey:       partOfAlertLabelValue,
+			componentAlertLabelKey:    componentAlertLabelValue,
 		},
 	},
 	{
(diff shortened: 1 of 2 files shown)


internal/operands/metrics/resources.go
```diff
@@ -99,9 +101,10 @@ var alertRulesList = []promv1.Rule{
 			"runbook_url": runbookURLBasePath + "SSPTemplateValidatorDown",
 		},
 		Labels: map[string]string{
-			severityAlertLabelKey:  "critical",
-			partOfAlertLabelKey:    partOfAlertLabelValue,
-			componentAlertLabelKey: componentAlertLabelValue,
+			severityAlertLabelKey:     "critical",
+			healthImpactAlertLabelKey: "critical",
```
sradco, line 105: @ksimon1 @akrejcir should this alert have a critical severity?  Would a template validator require waking someone in the middle if the night?

sradco: Based on the runboks, https://github.com/kubevirt/monitoring/blob/main/docs/runbooks/SSPTemplateValidatorDown.md , I would argue that the severity is warning. But the health impact is critical. What do you think?

ksimon1: AFAIR when template validator is down, it will prevent creation of any VM from template.

sradco: Who can answer this? Is this indeed critical on the cluster level and require handling immediately?

akrejcir: Template validator is a validating webhook, that gets called for every VM create or update. 
If it is down, then no VMs can be created or updated, which also means a VM cannot be started.

assafad:
Hi @akrejcir, can you please review this PR?

internal/operands/metrics/resources.go
sradco: @akrejcir is it critical on the cluster level ? or only on the operator health level?

akrejcir: I'm not sure. What is the difference between the levels?

akrejcir:
Looks good, but holding until the above conversation is resolved.

/lgtm
/hold

internal/operands/metrics/resources.go
akrejcir: Thanks. According to this part of the docs: 

> Reserve critical level alerts only for reporting conditions that may lead to loss of data or inability to deliver service for the cluster as a whole.

the severity does not need to be critical. Without the Template Validator, VMs cannot be created and started, but other cluster functionality works.

sradco: @akrejcir  I was just wondering about this part, "inability to deliver service for the cluster as a whole." 
Is creating and updating VMs considered part of the must have for every cluster with CNV installed?
I think the answer is No, but I want to make sure..

akrejcir: The same reason can probably be applied to other alerts in SSP.

akrejcir: I don't know the answer.

sradco: I consulted with Ronen. Lets ignore this for this PR. 
I will open a different story/bug in case we need to update the severity for these alerts.

PR #475, opened 2023-01-03
api: Make all SSPSpec attributes optional

lyarwood:

This change makes all SSPSpec attributes optional and omits any when they are empty. This is required as the operator can now be deployed without operands such as the template-validator operand and as a result the associated TemplateValidator attribute of the SSPSpec is not always required anymore.

Fixes #

**Release note**:

```release-note
NONE
```


akrejcir:
I think this is not a backward compatible change. Old operator will fail to decode the new JSON into the old object.

I've read somewhere that when removing fields from API object, the version should be incremented. But I cannot find it now.

lyarwood:
> I think this is not a backward compatible change. Old operator will fail to decode the new JSON into the old object.

AFAIK k8s will reject the resource as it doesn't conform to the older schema where some of these attributes were required yes.

> I've read somewhere that when removing fields from API object, the version should be incremented. But I cannot find it now.

Yeah apologies as this isn't a full removal and just a move to optional I didn't think about this, I blame the two weeks away from my keyboard over the holidays! 

I guess that for something like this to land then we will need to bump the version to `v1beta2`, moving these attributes to optional etc then. Lots of churn for a simple clean up but I don't mind taking a look. Moving this to draft for the time being.

PR #476, opened 2023-01-04
data_sources: Always run the operand

lyarwood:

The operand itself deploys DataImportCrons defined by the SSP CR and DataSources based on the provided common-templates bundle. While the latter functionality should likely reside in the common-templates operand the DataImportCrons functionaility is useful in non-OpenShift environments alongside that of the common-instancetypes operand.

This change now ensures the data_sources operand runs on both OpenShift and non-OpenShift environments.

Fixes #

**Release note**:

```release-note
The `data_sources` operand is now always deployed, regardless of the underlying environment being OpenShift or OKD.
```

### DIFF
--- a/controllers/setup.go
+++ b/controllers/setup.go
@@ -57,6 +57,7 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 
 	sspOperands := []operands.Operand{
 		common_instancetypes_operand,
+		data_sources.New(templatesBundle.DataSources),
 	}
 
 	runningOnOpenShift, err := common.RunningOnOpenshift(ctx, mgr.GetAPIReader())
@@ -69,7 +70,6 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 			metrics.New(),
 			template_validator.New(),
 			common_templates.New(templatesBundle.Templates),
-			data_sources.New(templatesBundle.DataSources),
 			node_labeller.New(),
 		)
 	}


lyarwood:
Actually I think I'm going to propose a different approach here with a new operand that isn't dependent on common-templates.

0xFelix:
Other than that `DataImportCronTemplates` are defined below `CommonTemplates` in the SSP CR I don't see any dependency between the data_sources operand and common-templates. Is there really that much to rework here?

See https://github.com/kubevirt/ssp-operator/blob/master/api/v1beta1/ssp_types.go#L48

/approve

lyarwood:
> Other than that `DataImportCronTemplates` are defined below `CommonTemplates` in the SSP CR I don't see any dependency between the data_sources operand and common-templates. Is there really that much to rework here?
> 
> See https://github.com/kubevirt/ssp-operator/blob/master/api/v1beta1/ssp_types.go#L48
> 
> /approve

Yeah apologies, it isn't obvious from the operand itself but we pass `templatesBundle.DataSources` into `New()` :

https://github.com/kubevirt/ssp-operator/blob/781a27fb7c248a3772812b54bcefab88e65cda90/controllers/setup.go#L60

This is created by the following `template-bundle` code so it's not a direct dependency between the operands but definitely between the template functionality and the `data_sources` operand:

https://github.com/kubevirt/ssp-operator/blob/781a27fb7c248a3772812b54bcefab88e65cda90/internal/template-bundle/bundle.go#L60-L93

I'd really like to understand the reasoning behind this and the creation of the `DataSources` outside of CDI, removing the latter if possible allowing us to break this linkage between the two.

0xFelix:
/lgtm

Then let's enable the operand for now as is and do the clean up in a separate PR.

PR #482, opened 2023-01-13
commonTemplates_test.go: Wait until restored templates increases

0xFelix:

This moves the check for the increased restored templates count into an Eventually block with a short timeout. This makes the test more robust by giving SSP some time to increase the count.

**Release note**:

```release-note
None
```


0xFelix:
Closing this for now

PR #485, opened 2023-01-19
fix: Do not set GOARCH to amd64

akrejcir:
Remove the `GOARCH` variable when building container. It may be build for a different architecture.

**Release note**:
```release-note
None
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,8 @@ COPY hack/boilerplate.go.txt hack/boilerplate.go.txt
 COPY hack/csv-generator.go hack/csv-generator.go
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on make manager
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on make csv-generator
+RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on make manager
+RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on make csv-generator
 
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal
--- a/validator.Dockerfile
+++ b/validator.Dockerfile
@@ -22,7 +22,7 @@ COPY api/ api/
 COPY controllers/ controllers/
 COPY internal/ internal/
 
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -ldflags="-X 'kubevirt.io/ssp-operator/internal/template-validator/version.COMPONENT=$COMPONENT'\
+RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -a -ldflags="-X 'kubevirt.io/ssp-operator/internal/template-validator/version.COMPONENT=$COMPONENT'\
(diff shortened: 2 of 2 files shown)


lyarwood (commented):
What's the justification for this if the overall Dockerfile is still pinned to amd64?

lyarwood:
> What's the justification for this if the overall Dockerfile is still pinned to amd64?

Ah apologies ubi is a multi-platform image so we pull down the relevent image for our hosts arch during build, still we install the `amd64` version of golang regardless, I would assume that would need to change for builds to work on `arm64` etc?

lyarwood:
I'm not sure if `buildah` supports this but `docker` passes in a `TARGETPLATFORM`  arg that we could use to install the correct version of golang FWIW:

https://nielscautaerts.xyz/making-dockerfiles-architecture-independent.html
https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

lyarwood:
Okay it does, so I can guess we can drop `GOARCH` and use `TARGETPLATFORM` to install the correct version of golang?
```
$ cat Dockerfile 
FROM alpine
ARG TARGETPLATFORM
RUN echo "I'm building for $TARGETPLATFORM"

$ buildah build -f ~/Dockerfile 
WARN[0000] missing "TARGETPLATFORM" build argument. Try adding "--build-arg TARGETPLATFORM=<VALUE>" to the command line 
STEP 1/3: FROM alpine
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
Getting image source signatures
Copying blob 8921db27df28 done  
Copying config 042a816809 done  
Writing manifest to image destination
Storing signatures
STEP 2/3: ARG TARGETPLATFORM
STEP 3/3: RUN echo "I'm building for $TARGETPLATFORM"
I'm building for linux/amd64
COMMIT
Getting image source signatures
Copying blob 8e012198eea1 skipped: already exists  
Copying blob 351d582fc7be done  
Copying config 912026a5e9 done  
Writing manifest to image destination
Storing signatures
--> 912026a5e9a
912026a5e9a90572bf95263ecc766da140095936d9164788ef91b526edbaa825
```

0xFelix:
The arch of the toolchain also depends on the arch of the build host. We might build for aarch64 but the building host is amd64?

lyarwood:
> The arch of the toolchain also depends on the arch of the build host. We might build for aarch64 but the building host is amd64?

Both `docker buildx` [1] and `buildah` [2] handle this by providing a suitable target build env if the arch differs from the host:

[1] https://github.com/docker/buildx#building-multi-platform-images
[2] https://danmanners.com/posts/2022-01-buildah-multi-arch/

0xFelix:
Should we add more changes to properly support multiarch builds (looking at the example Dockerfile in [1]) or is this enough for the time being?

[1] https://github.com/docker/buildx#building-multi-platform-images

akrejcir:
I missed that `Dockerfile` downloads and installs golang `amd64`. 
I will need to change this. Thanks for the comments.

akrejcir:
Closing, because the default branch is moving to `main`. We can return to this in a future.
I've opened an issue to track it: https://github.com/kubevirt/ssp-operator/issues/605

PR #486, opened 2023-01-20
Add alert for rhel6 running vms

fossedihelm:
RHEL6 guests will be not supported on RHEL9 hosts anymore [1].
And CNV 4.13 is planned to be on RHCOS 9, so RHEL6 cannot be supported in CNV 4.13 anymore.
Adding a vm controller that watch for rhel 6 vms.
Rhel6 template adds a label in the vm in the format: `rhel6-<workload>-<flavor>`.
This label is checked to determine whether a vm is rhel6 or not.

[1] https://access.redhat.com/articles/973163#rhelkvm

Fixes #

**Release note**:

```release-note
RHEL6 guests will be not supported on RHEL9 hosts anymore. Add alert for rhel6 running vms
```
/cc @akrejcir

### DIFF
--- a/controllers/vm_controller.go
+++ b/controllers/vm_controller.go
@@ -1,0 +1,111 @@
+package controllers
+
+import (
+	"context"
+	"github.com/go-logr/logr"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"k8s.io/apimachinery/pkg/api/errors"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"strings"
+)
+
+const (
+	vmControllerName = "vm-controller"
+	rhel6MetricName  = "kubevirt_vm_rhel6"
+)
+
+var (
+	VmRhel6 = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: rhel6MetricName,
+			Help: "Indication for a VirtualMachine that is based on rhel6 template",
+		},
+		[]string{
+			"namespace", "name",
+		},
+	)
+)
+
+// Annotation to generate RBAC roles to read virtualmachines
+// +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachines,verbs=get;list;watch
+
+func CreateVmController(mgr ctrl.Manager) (*vmReconciler, error) {
+	return newVmReconciler(mgr)
+}
+
+func (r *vmReconciler) Name() string {
+	return vmControllerName
+}
+
+func (r *vmReconciler) Start(ctx context.Context, mgr ctrl.Manager) error {
+	return r.setupController(mgr)
+}
+
+func (r *vmReconciler) setupController(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("vm-controller").
+		For(&kubevirtv1.VirtualMachine{}, builder.WithPredicates(predicate.NewPredicateFuncs(
+			func(object client.Object) bool {
+				return hasRhel6TemplateLabel(object)
+			}))).
+		Complete(r)
+}
+
+// vmReconciler watches the vms in the cluster
+type vmReconciler struct {
+	client client.Client
+	log    logr.Logger
+}
+
+func newVmReconciler(mgr ctrl.Manager) (*vmReconciler, error) {
+	logger := ctrl.Log.WithName("controllers").WithName("VirtualMachines")
+	reconciler := &vmReconciler{
+		client: mgr.GetClient(),
+		log:    logger,
+	}
+
+	return reconciler, nil
+}
+
+func (r *vmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
+	r.log.V(1).Info("Starting vm reconciliation...", "request", req.String())
+	vm := kubevirtv1.VirtualMachine{}
+	err = r.client.Get(ctx, req.NamespacedName, &vm)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			VmRhel6.WithLabelValues(req.Namespace, req.Name).Set(0)
+			r.log.Info("VM not found", "vm", req.NamespacedName)
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return ctrl.Result{}, nil
+		}
+		r.log.V(1).Info("Error retrieving the VM", "vm", req.NamespacedName)
+		// Error reading the object - requeue the request.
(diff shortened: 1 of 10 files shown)


controllers/vm_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	reconciler := &vmReconciler{
+		client: mgr.GetClient(),
+		log:    logger,
+	}
+
+	return reconciler, nil
+}
+
+func (r *vmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
+	r.log.Info("Starting vm reconciliation...", "request", req.String())
```
akrejcir, line 77: This log will be to verbose, because it will be printed for every VM creation, update and delete.
Can you change it to debug level using `.V(1)`? Like here: https://github.com/kubevirt/ssp-operator/blob/aabcf8d8d1c61ca04b97b4c830c2ff4c14f88458/controllers/ssp_controller.go#L212

controllers/vm_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	vm := kubevirtv1.VirtualMachine{}
+	err = r.client.Get(ctx, req.NamespacedName, &vm)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			VmRhel6.WithLabelValues(req.Namespace, req.Name).Set(0)
+			r.log.Info("VM not found", "vm", req.NamespacedName)
```
akrejcir, line 83: Similarly, this should be debug level.

tests/monitoring_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"k8s.io/utils/pointer"
+	kubevirtv1 "kubevirt.io/api/core/v1"
```
akrejcir, line 9: Can you move these lines to the second last block?

tests/monitoring_test.go
```diff
@@ -154,6 +156,41 @@ var _ = Describe("Prometheus Alerts", func() {
...
+			waitForAlertToActivate(metrics.Rhel6AlertName)
+		})
+
+		It("Should deactivate the DeprecatedRHEL6Vm alert if the rhel6 running vm is stopped", func() {
+			waitForAlertToActivate(metrics.Rhel6AlertName)
+			foundVm := &kubevirtv1.VirtualMachine{}
+			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), foundVm)
+			Expect(err).ToNot(HaveOccurred())
+			foundVm.Spec.Running = pointer.Bool(false)
+			Expect(apiClient.Update(ctx, foundVm)).ToNot(HaveOccurred())
```
akrejcir, line 190: These lines should be wrapped in an `Eventually()` block, because the update may fail when VM status is updated concurrently.

akrejcir (commented):
Looks good. I have only small comments.

tests/monitoring_test.go
fossedihelm: I was thinking to use `Patch` instead of `Update`, avoiding the `Eventually` block. WDYT?

tests/monitoring_test.go
fossedihelm: Sure!

controllers/vm_controller.go
fossedihelm: Totally agree, thanks

fossedihelm:
@akrejcir I saw that a `metric_test`  is failing since there's still no runbook for the new alert.

tests/monitoring_test.go
akrejcir: Ok, you can use `Patch`, but it will probably be more code than using `Eventually()`.

akrejcir:
> @akrejcir I saw that a `metric_test` is failing since there's still no runbook for the new alert.

Can you add code to the failing test, that will skip this specific alert? Then later we can remove it, when the runbook exists.

tests/monitoring_test.go
fossedihelm: At the end I prefer to use your suggestion :)

fossedihelm:
Thanks @akrejcir for your review! I pushed the changes to address your comments. PTAL! Thanks

0xFelix (commented):
Do we actually need a new controller for this or could this be implemented as an operand like we already do for other parts of ssp?

akrejcir:
> Do we actually need a new controller for this or could this be implemented as an operand like we already do for other parts of ssp?

We discussed this before, and I suggested creating a controller. The reason is that the operands in ssp-controller are only active if the SSP object is present in the cluster, and they can only watch objects they own. In the case of this PR, the VMs are not owned by ssp, so operand would not work.

Also, it would have worse performance, because on every update of a VM, the whole ssp-controller logic would run.

internal/operands/metrics/resources.go
```diff
@@ -152,6 +153,20 @@ var alertRulesList = []promv1.Rule{
 			componentAlertLabelKey:    componentAlertLabelValue,
 		},
 	},
+	{
+		Alert: Rhel6AlertName,
+		Expr:  intstr.FromString("sum by (namespace, name) (kubevirt_vm_rhel6) > 0"),
+		Annotations: map[string]string{
+			"summary": "VM {{ $labels.namespace }}/{{ $labels.name }} is based on RHEL6 template, and this will not be supported in the next release",
+			//"runbook_url": runbookURLBasePath + Rhel6AlertName,
```
0xFelix, line 161: How can the runbook / `runbook_url` be created?

lyarwood:
I'm not against this approach tbh, however I'd like to see more justification in the PR/ReleaseNote/Commits about why we need this (RHEL 6 is ELS downstream but what does that mean here?) and some docs somewhere in the tree about the controller before we merge this if possible.

internal/operands/metrics/resources.go
fossedihelm: Thanks! The runbook will be created under [kubevirt/monitoring](https://github.com/kubevirt/monitoring) project

fossedihelm:
@lyarwood Thanks for the offline discussion.
Can you have a look? @0xFelix @akrejcir Thanks

internal/operands/metrics/resources.go
0xFelix: Can it be created in advance? Or only after the metric is available? (Sorry, I don't know much about runbooks)

fossedihelm: I don't know much neither :) In general, I think the two things can go in parallel. Personally, I think that documentation that says there is something, but in fact there is not, is worse than providing something without saying it. But I am open to other points of view. If you prefer to firstly update the documentation, no problem at all. WDYT?

fossedihelm:
/hold
Evaluating some alternative that can avoid the new vm controller.

fossedihelm:
/unhold
The alternative will create a tight coupling between CMO and CNV, which is not good

fossedihelm:
@akrejcir Can you have a look again, thank!

internal/operands/metrics/resources.go
sradco: @fossedihelm every alerts must have a runbook in advance. Changing this test just to get the alert in the code is not good practice. The test was set in place for a reason.
Please revert the change to the test and add a runbook and runbook_url annotation to the alert asap. Thank you.

fossedihelm: Thanks @sradco https://github.com/kubevirt/monitoring/pull/177 for the runbook.
ssp changes ready as soon the monitoring pr is merged. Thanks and sorry

PR #493, opened 2023-02-08
common-instancetypes: Add support to reconcile from a kustomize repo URI

lyarwood:

This change introduces support within the operand for reconciling from a
kustomize repo URI. Existing bundled resources or previously generated
kustomize resources are removed entirely when the URI is modified.

The bulk of the change is caused by the vendoring of the kustomize
krusty entry point module used to render resources from the given URI.

Fixes #

**Release note**:

```release-note
The `common-instancetypes` operand now supports reconciling resources from a `Kustomize` remote target through the `CommonInstancetypes.URI` attributes of the `SSP` `CustomResource`. 
```


internal/operands/common-instancetypes/reconcile.go
```diff
@@ -58,50 +97,121 @@ func fetchClusterResources[C clusterType](bundlePath string) ([]C, error) {
...
+	return virtualMachineClusterInstancetypes, virtualMachineClusterPreferences, nil
 }
 
-func (c *commonInstancetypes) WatchTypes() []operands.WatchType {
-	return nil
+func (c *commonInstancetypes) reconcileFromURI(request *common.Request) ([]common.ReconcileResult, error) {
+	request.Logger.Info(fmt.Sprintf("Reconciling common-instancetypes from URI %s", *request.Instance.Spec.CommonInstancetypes.URI))
+
+	// Remove any existing resources if the URI is new or has changed
+	if c.resourceURI == "" || c.resourceURI != *request.Instance.Spec.CommonInstancetypes.URI {
```
lyarwood, line 184: This is definitely wrong and will cleanup resources on restart, I'll look again tomorrow.

api/v1beta1/ssp_types.go
```diff
@@ -66,6 +71,9 @@ type SSPSpec struct {
 
 	// TLSSecurityProfile is a configuration for the TLS.
 	TLSSecurityProfile *ocpv1.TLSSecurityProfile `json:"tlsSecurityProfile,omitempty"`
+
+	// CommonInstancetypes is the confgiruation of the common instance types operand
```
0xFelix, line 75: Typo: confgiruation

config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -39,6 +39,17 @@ spec:
           spec:
             description: SSPSpec defines the desired state of SSP
             properties:
+              commonInstancetypes:
+                description: CommonInstancetypes is the confgiruation of the common
+                  instance types operand
+                properties:
+                  uri:
+                    description: URI of common-instancetypes kustomise repo from which
```
0xFelix, line 47: `kustomise` -> `kustomize`

go.mod
```diff
@@ -38,6 +38,11 @@ require (
 	sigs.k8s.io/controller-runtime v0.13.1
 )
 
+require (
```
0xFelix, line 41: New require block?

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -38,16 +49,45 @@ type clusterType interface {
...
+
+func New(virtualMachineClusterInstancetypeBundlePath, virtualMachineClusterPreferenceBundlePath string) *commonInstancetypes {
+	return &commonInstancetypes{
+		virtualMachineClusterInstancetypeBundle: virtualMachineClusterInstancetypeBundlePath,
+		virtualMachineClusterPreferenceBundle:   virtualMachineClusterPreferenceBundlePath,
+	}
+}
+
+func decodeResources[C clusterType](b []byte) ([]C, error) {
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(b), 1024)
```
0xFelix, line 86: I asked the same question in `vm-console-proxy`, do we want to support both YAML and JSON as input formats?

config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -39,6 +39,17 @@ spec:
           spec:
             description: SSPSpec defines the desired state of SSP
             properties:
+              commonInstancetypes:
+                description: CommonInstancetypes is the confgiruation of the common
+                  instance types operand
+                properties:
+                  uri:
```
dominikholler, line 46: What is the URI? Does it have to be http/https? Would it work with a github repo, or a plain git repo?

lyarwood: Updated with more context.

internal/operands/common-instancetypes/reconcile.go
lyarwood: The support for JSON is incidental here, we want the features provided by the decoder over the pure YAML reader. Mainly that it decodes per object instead of the entire YAML document.

Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on make csv-generator
 FROM registry.access.redhat.com/ubi9/ubi-minimal
 LABEL org.kubevirt.hco.csv-generator.v1="/csv-generator"
 
-RUN microdnf update -y && microdnf clean all
+RUN microdnf update -y && microdnf install git -y && microdnf clean all
```
0xFelix, line 42: What is git needed for in the container image?

api/v1beta1/ssp_types.go
```diff
@@ -53,6 +53,27 @@ type NodeLabeller struct {
...
 }
 
+type CommonInstancetypes struct {
+	// URL of a Kustomize remote target from which to generate and deploy resources.
+	//
+	// The following caveats apply to the provided URL:
+	//
+	// * 'file://' URLs are not supported and will be rejected by the SSP validation webhook.
+	//
+	// * The URL must include '?ref=$ref' pinning it to a specific reference. It is recommended
```
0xFelix, line 63: Can you add that referencing a branch will have unpredictable reconcile behavior?

internal/operands/common-instancetypes/reconcile.go
0xFelix: Can you use `NewYAMLToJSONDecoder` instead? It should have the desired behavior while only allowing YAML.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -38,16 +49,45 @@ type clusterType interface {
...
 	var bundle []C
 	for {
 		resource := new(C)
```
0xFelix, line 89: There is a collision with an imported package name and this variable.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -58,52 +98,196 @@ func fetchClusterResources[C clusterType](bundlePath string) ([]C, error) {
...
 
+func (c *commonInstancetypes) reconcileFromBundle(request *common.Request) ([]common.ReconcileResult, error) {
+	request.Logger.Info("Reconciling common-instancetypes from internal bundle")
+	virtualMachineClusterInstancetypes, virtualMachineClusterPreferences, err := c.fetchResourcesFromBundle()
+	if err != nil {
+		return nil, err
+	}
+
+	c.virtualMachineClusterInstancetypes = virtualMachineClusterInstancetypes
+	c.virtualMachineClusterPreferences = virtualMachineClusterPreferences
```
0xFelix, line 280: I think we need to reconcile removed resources here too.

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -165,6 +171,93 @@ var _ = Describe("Common-Instancetypes operand", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(virtualMachineClusterInstancetypesFromURL).To(BeEmpty())
+		Expect(virtualMachineClusterPreferencesFromURL).ToNot(BeEmpty())
+		assertResoucesExist(request, virtualMachineClusterInstancetypesFromURL, virtualMachineClusterPreferencesFromURL)
+
+	})
+
+	It("should cleanup resources no longer provided after updating the URL", func() {
+		request.Instance.Spec.CommonInstancetypes = &ssp.CommonInstancetypes{
+			URL: pointer.String("https://github.com/kubevirt/common-instancetypes//VirtualMachineClusterPreferences"),
```
0xFelix, line 234: Can this work without ref if the validation webhook denies URL without ref?

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -165,6 +171,93 @@ var _ = Describe("Common-Instancetypes operand", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		virtualMachineClusterInstancetypes, virtualMachineClusterPreferences, err := FetchResourcesFromURL(*request.Instance.Spec.CommonInstancetypes.URL)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(virtualMachineClusterInstancetypes).To(BeEmpty())
+		Expect(virtualMachineClusterPreferences).ToNot(BeEmpty())
+		assertResoucesExist(request, virtualMachineClusterInstancetypes, virtualMachineClusterPreferences)
+
+		request.Instance.Spec.CommonInstancetypes = &ssp.CommonInstancetypes{
+			URL: pointer.String("https://github.com/kubevirt/common-instancetypes//VirtualMachineClusterInstancetypes"),
```
0xFelix, line 246: Same as above.

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -165,6 +171,93 @@ var _ = Describe("Common-Instancetypes operand", func() {
...
+		Expect(virtualMachineClusterInstancetypesFromBundle).ToNot(BeEmpty())
+
+		virtualMachineClusterPreferencesFromBundle, err := FetchBundleResource[instancetypev1alpha2.VirtualMachineClusterPreference](preferencePath)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(virtualMachineClusterPreferencesFromBundle).ToNot(BeEmpty())
+		assertResoucesExist(request, virtualMachineClusterInstancetypesFromBundle, virtualMachineClusterPreferencesFromBundle)
+
+		// Switch over to the URL and only install VirtualMachineClusterPreferences from common-instancetypes
+		request.Instance.Spec.CommonInstancetypes = &ssp.CommonInstancetypes{
+			URL: pointer.String("https://github.com/kubevirt/common-instancetypes//VirtualMachineClusterPreferences"),
```
0xFelix, line 214: Same as below.

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -165,6 +171,93 @@ var _ = Describe("Common-Instancetypes operand", func() {
 		Expect(cleanupResult).To(BeNil())
 		Expect(err).ToNot(HaveOccurred())
 	})
+
+	It("should create and cleanup common-instancetypes resources from an external URL", func() {
+		request.Instance.Spec.CommonInstancetypes = &ssp.CommonInstancetypes{
+			URL: pointer.String("https://github.com/kubevirt/common-instancetypes"),
```
0xFelix, line 177: Same as below.

webhooks/ssp_webhook.go
```diff
@@ -174,6 +183,23 @@ func validateDataImportCronTemplates(ssp *sspv1beta1.SSP) error {
...
+	if ssp.Spec.CommonInstancetypes == nil || ssp.Spec.CommonInstancetypes.URL == nil {
+		return nil
+	}
+
+	url := *ssp.Spec.CommonInstancetypes.URL
+	if strings.HasPrefix(url, "https://") && strings.HasPrefix(url, "ssh://") {
+		return fmt.Errorf("%s is invalid, only https:// or ssh:// are supported as a remote kustomize target for commonInstancetypes", url)
+	}
+
+	if !strings.Contains(url, "?ref=") {
```
0xFelix, line 196: `version=` is valid too, see https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md.

Dockerfile
lyarwood: Referenced in the commit but it's actually a poorly documented dependency of `kustomize` itself:

https://github.com/kubernetes-sigs/kustomize/blob/183928e6baab759f61c033892e32ab5759f442e6/api/internal/git/gitrunner.go#L22-L38

api/v1beta1/ssp_types.go
lyarwood: I've basically said as much but I'll make it more explicit yeah.

api/v1beta1/ssp_types.go
```diff
@@ -53,6 +53,27 @@ type NodeLabeller struct {
...
+	//   not change over time.
+	//
+	// * Only VirtualMachineClusterPreference and VirtualMachineClusterInstancetype
+	//   resources generated from the URL are deployed by the operand.
+	//
+	// See the following Kustomize documentation for more details:
+	//
+	// remote targets
+	// https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
+	URL *string `json:"url",omitempty`
```
akrejcir, line 74: This is a typo, the tag should be: `` `json:"url,omitempty"` ``

webhooks/ssp_webhook.go
```diff
@@ -174,6 +183,23 @@ func validateDataImportCronTemplates(ssp *sspv1beta1.SSP) error {
 	return nil
 }
 
+func validateCommonInstancetypes(ssp *sspv1beta1.SSP) error {
+	if ssp.Spec.CommonInstancetypes == nil || ssp.Spec.CommonInstancetypes.URL == nil {
+		return nil
+	}
+
+	url := *ssp.Spec.CommonInstancetypes.URL
+	if strings.HasPrefix(url, "https://") && strings.HasPrefix(url, "ssh://") {
```
akrejcir, line 192: This condition is incorrect, it should be:
```
if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "ssh://") {
```

internal/common/labels.go
```diff
@@ -61,3 +65,11 @@ func addInstanceLabels(requestInstance *v1beta1.SSP, to map[string]string) {
 func copyLabel(from, to map[string]string, key string) {
 	to[key] = from[key]
 }
+
+func GetAppNameSelector(name string) labels.Selector {
+	appNameRequirement, err := labels.NewRequirement(AppKubernetesNameLabel, selection.Equals, []string{name})
+	if err != nil {
+		panic(fmt.Sprintf("Failed creating label selector for '%s=%s'", AppKubernetesNameLabel, name))
```
akrejcir, line 72: I think this function should not panic on error, because the error can be caused by an incorrect value of `name` parameter.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -58,52 +98,196 @@ func fetchClusterResources[C clusterType](bundlePath string) ([]C, error) {
...
+	virtualMachineClusterPreferences, err := FetchBundleResource[instancetypev1alpha2.VirtualMachineClusterPreference](c.virtualMachineClusterPreferenceBundle)
+	if err != nil {
+		return nil, nil, err
+	}
+	return virtualMachineClusterInstancetypes, virtualMachineClusterPreferences, err
+}
+
+func generateKustomization(URL string) (filesys.FileSystem, string, error) {
+	fSys := filesys.MakeFsOnDisk()
+	tmpDir, err := filesys.NewTmpConfirmedDir()
```
akrejcir, line 123: I think the temporary directory should be removed when it's no longer needed, otherwise it is a leak.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -58,52 +98,196 @@ func fetchClusterResources[C clusterType](bundlePath string) ([]C, error) {
...
-		virtualMachineClusterPreferences:   virtualMachineClusterPreferences,
-	}, nil
+	k := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	return k.Run(fSys, tmpDirPath)
 }
 
-func (c *commonInstancetypes) Name() string {
-	return operandName
+func decodeResMapResources[C clusterType](r *resource.Resource) ([]C, error) {
+	bytes, err := r.AsYAML()
```
akrejcir, line 143: Can you encode it to json, instead of yaml? The `AsYAML()` method first encodes it to json and then translates json to yaml. We can skip the translation step, because the `decodeResources()` below can load json.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -58,52 +98,196 @@ func fetchClusterResources[C clusterType](bundlePath string) ([]C, error) {
...
+	if c.resourceURL != "" && c.resourceURL == *request.Instance.Spec.CommonInstancetypes.URL {
+		request.Logger.Info(fmt.Sprintf("Skipping reconcile of common-instancetypes from URL %s, force with a restart of the service.", *request.Instance.Spec.CommonInstancetypes.URL))
+		return nil, nil
+	}
+
+	// Stash the URL
+	c.resourceURL = *request.Instance.Spec.CommonInstancetypes.URL
+
+	request.Logger.Info(fmt.Sprintf("Reconciling common-instancetypes from URL %s", *request.Instance.Spec.CommonInstancetypes.URL))
+	clusterInstancetypesFromURL, clusterPreferencesFromURL, err := FetchResourcesFromURL(*request.Instance.Spec.CommonInstancetypes.URL)
```
akrejcir, line 242: This line fetches the contents of the URL on every reconcile cycle. Do we want to support reloading the contents of the URL if they change? If not, can you cache the resources, so the URL is accessed only when the `.spec.commonInstancetypes.URL` changes?

internal/operands/common-instancetypes/reconcile_test.go
akrejcir, line 177: In my opinion it's unusual that unit tests depend on the contents of an external URL.
But in this case, I'm not sure how it can be avoided.

internal/operands/common-instancetypes/reconcile.go
akrejcir: I would say yes for both formats. In `vm-console-proxy` it was special case, because the config file was yaml. But here, we would need to unnecessarily translate json to yaml when encoding, and then back from yaml to json when decoding.

internal/operands/common-instancetypes/reconcile.go
lyarwood: See the conditional above, it's only triggered when the operand is first called or when the URL changes at the moment, we skip the rest of the time given the constraints that the URL needs to target a fixed `?ref=` or `?target=`.

internal/operands/common-instancetypes/reconcile_test.go
lyarwood: Fully agree, the only thing I can do here is create a nested kustomize target within the tree and use `file://` as we aren't going through the webhooks here. This doesn't exercise as much code but works offline.

internal/operands/common-instancetypes/reconcile_test.go
lyarwood: We don't exercise the webhooks from these unit tests so yes it works.

internal/operands/common-instancetypes/reconcile.go
akrejcir: Ok, I missed that condition.

lyarwood:
```
error: unable to read URL "https://github.com/kubevirt/ssp-operator/releases/download/null/ssp-operator.yaml", server reported 404 Not Found, status code=404
```

I'll look into this now.

lyarwood:
> ```
> error: unable to read URL "https://github.com/kubevirt/ssp-operator/releases/download/null/ssp-operator.yaml", server reported 404 Not Found, status code=404
> ```
> 
> I'll look into this now.

Fixed by removing the `v0.17.0` branch and documented under https://github.com/kubevirt/ssp-operator/issues/499

/retest-required

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -38,72 +50,269 @@ type clusterType interface {
...
+func New(virtualMachineClusterInstancetypeBundlePath, virtualMachineClusterPreferenceBundlePath string) *commonInstancetypes {
+	return &commonInstancetypes{
+		virtualMachineClusterInstancetypeBundle: virtualMachineClusterInstancetypeBundlePath,
+		virtualMachineClusterPreferenceBundle:   virtualMachineClusterPreferenceBundlePath,
 	}
-	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(file), 1024)
+}
+
+func decodeResources[C clusterType](b []byte) ([]C, error) {
+	decoder := yaml.NewYAMLToJSONDecoder(bytes.NewReader(b))
```
akrejcir, line 87: This will only decode `b` as yaml, but the function `decodeResMapResources()` calls it with json argument.
Can you leave the json decoder here?

lyarwood: https://github.com/kubevirt/ssp-operator/compare/37939ca1995be819c5fb9c0226aec01bd553c396..a5965d7db2de19ef34ce756528e0d4366b62582c is what you are suggesting right @akrejcir ?

akrejcir: yes

webhooks/ssp_webhook.go
```diff
@@ -174,6 +183,21 @@ func validateDataImportCronTemplates(ssp *sspv1beta1.SSP) error {
...
+func validateCommonInstancetypes(ssp *sspv1beta1.SSP) error {
+	if ssp.Spec.CommonInstancetypes == nil || ssp.Spec.CommonInstancetypes.URL == nil {
+		return nil
+	}
+
+	url := *ssp.Spec.CommonInstancetypes.URL
+	if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "ssh://") {
+		return fmt.Errorf("%s is invalid, only https:// or ssh:// are supported as a remote kustomize target for commonInstancetypes", url)
+	}
+	if !strings.Contains(url, "?ref=") && !strings.Contains(url, "?version=") {
```
0xFelix, line 195: Use `strings.HasSuffix` instead of `strings.Contains`?

webhooks/ssp_webhook.go
```diff
@@ -174,6 +183,21 @@ func validateDataImportCronTemplates(ssp *sspv1beta1.SSP) error {
 	return nil
 }
 
+func validateCommonInstancetypes(ssp *sspv1beta1.SSP) error {
+	if ssp.Spec.CommonInstancetypes == nil || ssp.Spec.CommonInstancetypes.URL == nil {
+		return nil
+	}
+
+	url := *ssp.Spec.CommonInstancetypes.URL
+	if !strings.HasPrefix(url, "https://") && !strings.HasPrefix(url, "ssh://") {
```
0xFelix, line 192: Can/should we support `http://` too?

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -38,72 +50,269 @@ type clusterType interface {
...
+		}
+	}
 	return nil
 }
 
-func (c *commonInstancetypes) RequiredCrds() []string {
-	return []string{
-		virtualMachineClusterInstancetypeCrd,
-		virtualMachineClusterPreferenceCrd,
+func reconcileRemovedPreferences(request *common.Request, existingResources []instancetypev1alpha2.VirtualMachineClusterPreference, resourcesFromURL []instancetypev1alpha2.VirtualMachineClusterPreference) error {
```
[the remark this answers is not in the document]
lyarwood: Yes they are but you can't access common attributes still so looking up `resourceFromURL.Name` isn't possible.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -38,72 +50,269 @@ type clusterType interface {
...
+	// Stash the URL
+	c.resourceURL = *request.Instance.Spec.CommonInstancetypes.URL
+
+	request.Logger.Info(fmt.Sprintf("Reconciling common-instancetypes from URL %s", *request.Instance.Spec.CommonInstancetypes.URL))
+	clusterInstancetypesFromURL, clusterPreferencesFromURL, err := FetchResourcesFromURL(*request.Instance.Spec.CommonInstancetypes.URL)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use the cached slices of resources unless this is our first time calling reconcileFromURL
```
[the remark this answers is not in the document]
lyarwood: As discussed offline I've removed the cache for now to simplify things.

webhooks/ssp_webhook.go
lyarwood: That assumes this is the only argument we end up supporting and is thus always a suffix, I'd rather stay with contains.

webhooks/ssp_webhook.go
lyarwood: The krusty/kustomize docs only highlight https so that's all I'm including for now.

lyarwood:
I've reworked things a little introducing a simple `KustomizeWrapper` that wraps the `krusty` entry point while allowing us to also mock things out in the unit tests. It's awkward but hopefully acceptable and makes the tests much better IMHO.

tests/common_instancetypes_test.go
```diff
@@ -0,0 +1,79 @@
...
+var _ = Describe("Common Instance Types", func() {
+	BeforeEach(func() {
+		strategy.SkipSspUpdateTestsIfNeeded()
+	})
+
+	AfterEach(func() {
+		strategy.RevertToOriginalSspCr()
+	})
+
+	FContext("operand", func() {
```
akrejcir, line 24: Focused context.

internal/operands/common-instancetypes/kustomize.go
```diff
@@ -0,0 +1,182 @@
...
+	"sigs.k8s.io/kustomize/api/resource"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/resid"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// This is a simple wrapper around the Krusty kustomize entry point allowing us to mock things out in unit tests
+type KustomizeWrapper struct {
```
0xFelix, line 15: I'd prefer if you make that an interface with just the `Run(fSys filesys.FileSystem, path string) (resmap.ResMap, error)` method. This way you can pass in the real `krusty.Kustomizer` or a `MockKustomizer` instance and we don't need the current proxy `Run` method.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -21,90 +30,270 @@ const (
...
+}
+
+func (c *CommonInstancetypes) reconcileFromBundle(request *common.Request) ([]common.ReconcileResult, error) {
+	request.Logger.Info("Reconciling common-instancetypes from internal bundle")
+	clusterInstancetypesFromBundle, clusterPreferencesFromBundle, err := c.fetchResourcesFromBundle()
+	if err != nil {
+		return nil, err
+	}
+
+	existingClusterInstancetypes, existingClusterPreferences, err := c.fetchExistingResources(request)
```
0xFelix, line 271: Move removal of existing resources into separate func to avoid duplication? Or what do you think of moving L271-L286 into `Reconcile` func and returning the new resources from `reconcileFromURL` and `reconcileFromBundle`?

webhooks/ssp_webhook.go
0xFelix: Agreed.

webhooks/ssp_webhook.go
0xFelix: Agreed.

internal/operands/common-instancetypes/reconcile.go
lyarwood: I'd rather have a shared call in `reconcileFromBundle` and `reconcileFromURL` tbh, so we remove first and then create/update.

tests/common_instancetypes_test.go
lyarwood: Well that's embarrassing, I'm tempted to also configure https://pre-commit.com/ to run `make lint` once https://github.com/kubevirt/ssp-operator/pull/495 lands tbh to catch silly mistakes like this.

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -138,32 +177,173 @@ var _ = Describe("Common-Instancetypes operand", func() {
...
+		ExpectResourceExists(preference, request)
+	})
+
+	It("should create and cleanup resources from an external URL", func() {
+		// Generate a mock ResMap and resources for the test
+		mockResMap, virtualMachineClusterInstancetypes, virtualMachineClusterPreferences, err := newMockResources(10, 10)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Use a mock Run function to return our fake ResMap
+		operand.KustomizeRunFunc = func(fSys filesys.FileSystem, path string) (resmap.ResMap, error) {
```
0xFelix, line 255: You could mark the func args as unused (`_`).

internal/operands/common-instancetypes/kustomize.go
```diff
@@ -0,0 +1,164 @@
+package common_instancetypes
```
0xFelix, line 1: Maybe rename this file to `mockresmap.go` to make it more obvious this is only test code?

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -138,32 +177,173 @@ var _ = Describe("Common-Instancetypes operand", func() {
...
+		Expect(err).ToNot(HaveOccurred())
+		Expect(originalInstancetypes).To(HaveLen(0))
+		Expect(originalPreferences).To(HaveLen(10))
 
-		// Assert that the CRDs are not present before we call Cleanup
-		Expect(request.CrdWatch.CrdExists(virtualMachineClusterInstancetypeCrd)).To(BeFalse())
-		Expect(request.CrdWatch.CrdExists(virtualMachineClusterPreferenceCrd)).To(BeFalse())
-		Expect(request.CrdWatch.MissingCrds()).To(HaveLen(2))
-		Expect(request.CrdWatch.MissingCrds()).To(ContainElements(virtualMachineClusterInstancetypeCrd, virtualMachineClusterPreferenceCrd))
+		operand.KustomizeRunFunc = func(fSys filesys.FileSystem, path string) (resmap.ResMap, error) {
```
0xFelix, line 285: Same as above.

akrejcir:
Looks good.
/lgtm
/approve

PR #498, opened 2023-02-15
test: add vm-console-proxy functional tests

codingben:

This PR will add vm-console-proxy functional tests.

**Release note**:

```release-note
Add vm-console-proxy functional tests
```

### DIFF
--- a/internal/operands/vm-console-proxy/reconcile.go
+++ b/internal/operands/vm-console-proxy/reconcile.go
@@ -14,10 +14,11 @@ import (
 )
 
 const (
-	kubeVirtNamespace = "kubevirt"
-	operandName       = "vm-console-proxy"
-	operandComponent  = "vm-console-proxy"
-	enableAnnotation  = "ssp.kubevirt.io/enable-vm-console-proxy"
+	EnableAnnotation                  = "ssp.kubevirt.io/vm-console-proxy-enabled"
+	VmConsoleProxyNamespaceAnnotation = "ssp.kubevirt.io/vm-console-proxy-namespace"
+
+	operandName      = "vm-console-proxy"
+	operandComponent = "vm-console-proxy"
 )
 
 // Define RBAC rules needed by this operand:
@@ -29,6 +30,17 @@ const (
 // +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachineinstances,verbs=get;list;watch
 // +kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
 // +kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
+
+func WatchClusterTypes() []operands.WatchType {
+	return []operands.WatchType{
+		{Object: &rbac.ClusterRole{}},
+		{Object: &rbac.ClusterRoleBinding{}},
+		{Object: &core.ServiceAccount{}},
+		{Object: &core.Service{}},
+		{Object: &apps.Deployment{}, WatchFullObject: true},
+		{Object: &core.ConfigMap{}},
+	}
+}
 
 type vmConsoleProxy struct {
 	serviceAccount     *core.ServiceAccount
@@ -61,14 +73,7 @@ func (v *vmConsoleProxy) WatchTypes() []operands.WatchType {
 }
 
 func (v *vmConsoleProxy) WatchClusterTypes() []operands.WatchType {
-	return []operands.WatchType{
-		{Object: &rbac.ClusterRole{}},
-		{Object: &rbac.ClusterRoleBinding{}},
-		{Object: &core.ServiceAccount{}},
-		{Object: &core.Service{}},
-		{Object: &apps.Deployment{}},
-		{Object: &core.ConfigMap{}},
-	}
+	return WatchClusterTypes()
 }
 
 func (v *vmConsoleProxy) RequiredCrds() []string {
@@ -88,12 +93,12 @@ func (v *vmConsoleProxy) Reconcile(request *common.Request) ([]common.ReconcileR
 		return []common.ReconcileResult{}, nil
 	}
 
-	reconcileFunc = append(reconcileFunc, reconcileServiceAccountsFuncs(*v.serviceAccount))
-	reconcileFunc = append(reconcileFunc, reconcileClusterRoleFuncs(*v.clusterRole))
-	reconcileFunc = append(reconcileFunc, reconcileClusterRoleBindingFuncs(*v.clusterRoleBinding))
-	reconcileFunc = append(reconcileFunc, reconcileConfigMapFuncs(*v.configMap))
-	reconcileFunc = append(reconcileFunc, reconcileServiceFuncs(*v.service))
-	reconcileFunc = append(reconcileFunc, reconcileDeploymentFuncs(*v.deployment))
+	reconcileFunc = append(reconcileFunc, reconcileServiceAccountsFuncs(*v.serviceAccount.DeepCopy()))
+	reconcileFunc = append(reconcileFunc, reconcileClusterRoleFuncs(*v.clusterRole.DeepCopy()))
+	reconcileFunc = append(reconcileFunc, reconcileClusterRoleBindingFuncs(*v.clusterRoleBinding.DeepCopy()))
+	reconcileFunc = append(reconcileFunc, reconcileConfigMapFuncs(*v.configMap.DeepCopy()))
+	reconcileFunc = append(reconcileFunc, reconcileServiceFuncs(*v.service.DeepCopy()))
+	reconcileFunc = append(reconcileFunc, reconcileDeploymentFuncs(*v.deployment.DeepCopy()))
 
 	reconcileBundleResults, err := common.CollectResourceStatus(request, reconcileFunc...)
 	if err != nil {
@@ -118,7 +123,7 @@ func (v *vmConsoleProxy) Cleanup(request *common.Request) ([]common.CleanupResul
 
 func reconcileServiceAccountsFuncs(serviceAccount core.ServiceAccount) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		serviceAccount.Namespace = kubeVirtNamespace
+		serviceAccount.Namespace = getVmConsoleProxyNamespace(request)
 		return common.CreateOrUpdate(request).
 			ClusterResource(&serviceAccount).
 			WithAppLabels(operandName, operandComponent).
@@ -157,17 +162,24 @@ func reconcileClusterRoleBindingFuncs(clusterRoleBinding rbac.ClusterRoleBinding
 
 func reconcileConfigMapFuncs(configMap core.ConfigMap) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		configMap.Namespace = kubeVirtNamespace
+		configMap.Namespace = getVmConsoleProxyNamespace(request)
 		return common.CreateOrUpdate(request).
 			ClusterResource(&configMap).
 			WithAppLabels(operandName, operandComponent).
+			UpdateFunc(func(newRes, foundRes client.Object) {
+				newConfigMap := newRes.(*core.ConfigMap)
+				foundConfigMap := foundRes.(*core.ConfigMap)
+				foundConfigMap.Immutable = newConfigMap.Immutable
+				foundConfigMap.Data = newConfigMap.Data
+				foundConfigMap.BinaryData = newConfigMap.BinaryData
+			}).
 			Reconcile()
 	}
 }
 
 func reconcileServiceFuncs(service core.Service) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		service.Namespace = kubeVirtNamespace
+		service.Namespace = getVmConsoleProxyNamespace(request)
 		return common.CreateOrUpdate(request).
 			ClusterResource(&service).
 			WithAppLabels(operandName, operandComponent).
@@ -183,8 +195,9 @@ func reconcileServiceFuncs(service core.Service) common.ReconcileFunc {
 }
 
 func reconcileDeploymentFuncs(deployment apps.Deployment) common.ReconcileFunc {
+	numberOfReplicas := *deployment.Spec.Replicas
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		deployment.Namespace = kubeVirtNamespace
+		deployment.Namespace = getVmConsoleProxyNamespace(request)
 		deployment.Spec.Template.Spec.Containers[0].Image = getVmConsoleProxyImage()
 		return common.CreateOrUpdate(request).
 			ClusterResource(&deployment).
@@ -195,13 +208,10 @@ func reconcileDeploymentFuncs(deployment apps.Deployment) common.ReconcileFunc {
 			StatusFunc(func(res client.Object) common.ResourceStatus {
 				dep := res.(*apps.Deployment)
 				status := common.ResourceStatus{}
-				numberOfReplicas := *deployment.Spec.Replicas
-
 				if numberOfReplicas > 0 && dep.Status.AvailableReplicas == 0 {
 					msg := fmt.Sprintf("No vm-console-proxy pods are running. Expected: %d", dep.Status.Replicas)
 					status.NotAvailable = &msg
 				}
-
 				if dep.Status.AvailableReplicas != numberOfReplicas {
 					msg := fmt.Sprintf(
 						"Not all template vm-console-proxy pods are running. Expected: %d, running: %d",
@@ -211,7 +221,6 @@ func reconcileDeploymentFuncs(deployment apps.Deployment) common.ReconcileFunc {
 					status.Progressing = &msg
 					status.Degraded = &msg
 				}
-
 				return status
 			}).
 			Reconcile()
@@ -222,7 +231,7 @@ func isEnabled(request *common.Request) bool {
 	if request.Instance.GetAnnotations() == nil {
 		return false
 	}
-	enabledStr, ok := request.Instance.GetAnnotations()[enableAnnotation]
+	enabledStr, ok := request.Instance.GetAnnotations()[EnableAnnotation]
 	if !ok {
 		return false
 	}
@@ -231,6 +240,18 @@ func isEnabled(request *common.Request) bool {
 		return false
 	}
 	return isEnabled
+}
+
+func getVmConsoleProxyNamespace(request *common.Request) string {
+	const defaultNamespace = "kubevirt"
+	if request.Instance.GetAnnotations() == nil {
+		return defaultNamespace
+	}
+	namespaceStr, ok := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
+	if !ok {
+		return defaultNamespace
+	}
+	return namespaceStr
 }
 
 func getVmConsoleProxyImage() string {
--- a/tests/vm_console_proxy_test.go
+++ b/tests/vm_console_proxy_test.go
@@ -1,0 +1,190 @@
+package tests
+
+import (
+	"reflect"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	apps "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/pointer"
+	ssp "kubevirt.io/ssp-operator/api/v1beta1"
+	vm_console_proxy "kubevirt.io/ssp-operator/internal/operands/vm-console-proxy"
+)
+
+var _ = Describe("VM Console Proxy Operand", func() {
+	var (
+		clusterRoleResource        testResource
+		clusterRoleBindingResource testResource
+		serviceAccountResource     testResource
+		serviceResource            testResource
+		deploymentResource         testResource
+		configMapResource          testResource
+	)
+
+	BeforeEach(func() {
+		strategy.SkipSspUpdateTestsIfNeeded()
+
+		updateSsp(func(foundSsp *ssp.SSP) {
+			if foundSsp.GetAnnotations() == nil {
+				foundSsp.Annotations = make(map[string]string)
+			}
+
+			namespace := strategy.GetVmConsoleProxyNamespace()
+
+			foundSsp.Annotations[vm_console_proxy.EnableAnnotation] = "true"
+			foundSsp.Annotations[vm_console_proxy.VmConsoleProxyNamespaceAnnotation] = namespace
+		})
+
+		expectedLabels := expectedLabelsFor("vm-console-proxy", "vm-console-proxy")
+		clusterRoleResource = testResource{
+			Name:           "vm-console-proxy",
+			Resource:       &rbac.ClusterRole{},
+			ExpectedLabels: expectedLabels,
+			UpdateFunc: func(role *rbac.ClusterRole) {
+				role.Rules[0].Verbs = []string{"watch"}
+			},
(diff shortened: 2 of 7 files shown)


tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,135 @@
...
+			Namespace:      strategy.GetVmConsoleProxyNamespace(),
+			Resource:       &core.Service{},
+			ExpectedLabels: expectedLabels,
+		}
+		deploymentResource = testResource{
+			Name:           "vm-console-proxy",
+			Namespace:      strategy.GetVmConsoleProxyNamespace(),
+			Resource:       &apps.Deployment{},
+			ExpectedLabels: expectedLabels,
+		}
```
akrejcir, line 50: For the update tests, these `testResource` structures need to have the `UpdateFunc` and `EqualsFunc` defined. They are used to do the update and then test that the resource was reverted.

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+
+	Context("Resource creation", func() {
+		DescribeTable("created cluster resource", func(res *testResource) {
+			resource := res.NewResource()
+			err := apiClient.Get(ctx, res.GetKey(), resource)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(hasOwnerAnnotations(resource.GetAnnotations())).To(BeTrue(), "owner annotations are missing")
+		},
+			Entry("[test_id:1000] cluster role", &clusterRoleResource),
```
akrejcir, line 68: These `test_id` are created by QE, and correspond to test plans in their system.
For now can you write `TODO` here, until the IDs are created?

codingben: Okay, who will responsible then to monitor the creation of the test plans and update them here? Will be there a ticket to update IDs?

tests/vm_console_proxy_test.go
codingben: Done, thanks!

tests/vm_console_proxy_test.go
akrejcir: You can ask QE to create the test IDs.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -118,7 +123,7 @@ func (v *vmConsoleProxy) Cleanup(request *common.Request) ([]common.CleanupResul
 
 func reconcileServiceAccountsFuncs(serviceAccount core.ServiceAccount) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		serviceAccount.Namespace = kubeVirtNamespace
+		serviceAccount.Namespace = getVmConsoleProxyNamespace(request)
 		return common.CreateOrUpdate(request).
 			ClusterResource(&serviceAccount).
 			WithAppLabels(operandName, operandComponent).
```
akrejcir, line 129: Can you add `UpdateFunc()` here, and copy all the fields from the `newRes` to `foundRes`? Except `TypeMeta` and `ObjectMeta`. It is needed to correctly reconcile the reosurce.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -157,7 +162,7 @@ func reconcileClusterRoleBindingFuncs(clusterRoleBinding rbac.ClusterRoleBinding
 
 func reconcileConfigMapFuncs(configMap core.ConfigMap) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		configMap.Namespace = kubeVirtNamespace
+		configMap.Namespace = getVmConsoleProxyNamespace(request)
 		return common.CreateOrUpdate(request).
 			ClusterResource(&configMap).
 			WithAppLabels(operandName, operandComponent).
```
akrejcir, line 168: Similarly here, can you add `UpdateFunc()`?

akrejcir (commented):
Can you make 2 commits in this PR? One will fix some bugs from the previous PR, like missing `UpdateFunc`s, and the other will add functional tests?

internal/operands/vm-console-proxy/reconcile.go
codingben: Yep, done.

internal/operands/vm-console-proxy/reconcile.go
codingben: Done.

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,177 @@
...
+					reflect.DeepEqual(old.Subjects, new.Subjects)
+			},
+		}
+		serviceAccountResource = testResource{
+			Name:           "vm-console-proxy",
+			Namespace:      strategy.GetVmConsoleProxyNamespace(),
+			Resource:       &core.ServiceAccount{},
+			ExpectedLabels: expectedLabels,
+			UpdateFunc: func(serviceAccount *core.ServiceAccount) {
+				serviceAccount.Secrets = nil
```
akrejcir, line 56: I'm not sure we can do any reasonable update test for SA, because it's hard to change any of the fields for a test. What secret would we add?
Can you skip the update test for now? So these functions are not needed.

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,177 @@
...
+				return reflect.DeepEqual(old.Spec, new.Spec)
+			},
+		}
+		deploymentResource = testResource{
+			Name:           "vm-console-proxy",
+			Namespace:      strategy.GetVmConsoleProxyNamespace(),
+			Resource:       &apps.Deployment{},
+			ExpectedLabels: expectedLabels,
+			UpdateFunc: func(deployment *apps.Deployment) {
+				deployment.Spec.Replicas = pointer.Int32Ptr(0)
```
akrejcir, line 81: Use non-deprecated function `pointer.Int32()`.

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,177 @@
...
+				return reflect.DeepEqual(old.Spec, new.Spec)
+			},
+		}
+		configMapResource = testResource{
+			Name:           "vm-console-proxy",
+			Namespace:      strategy.GetVmConsoleProxyNamespace(),
+			Resource:       &core.ConfigMap{},
+			ExpectedLabels: expectedLabels,
+			UpdateFunc: func(configMap *core.ConfigMap) {
+				configMap.Data = nil
```
akrejcir, line 93: It's better to add some unique data, because it may already be `nil` and we want to test an update.

codingben:
@akrejcir Thanks for the review, I'll take care of it on Sunday. Also I see this PR is affected by some issue:
```
error: unable to read URL "https://github.com/kubevirt/ssp-operator/releases/download/null/ssp-operator.yaml", server reported 404 Not Found
```

lyarwood:
> @akrejcir Thanks for the review, I'll take care of it on Sunday. Also I see this PR is affected by some issue:
> 
> ```
> error: unable to read URL "https://github.com/kubevirt/ssp-operator/releases/download/null/ssp-operator.yaml", server reported 404 Not Found
> ```

Should be resolved, see https://github.com/kubevirt/ssp-operator/issues/499/

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,177 @@
...
+	var (
+		clusterRoleResource        testResource
+		clusterRoleBindingResource testResource
+		serviceAccountResource     testResource
+		serviceResource            testResource
+		deploymentResource         testResource
+		configMapResource          testResource
+	)
+
+	BeforeEach(func() {
```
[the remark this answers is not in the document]
codingben: If I understand correctly, [here](https://github.com/kubevirt/ssp-operator/blob/9973ad5d20d778652925281f3d4072ba75e00c64/tests/tests_suite_test.go#L115) we already have relevant annotations, and in case they are not there, we should add them here in `vm_console_proxy_test.go`?

tests/vm_console_proxy_test.go
codingben: Yep, thanks for your attention. :)

tests/vm_console_proxy_test.go
codingben: Sure, changed to `pointer.Int32(0)`. Thanks!

tests/vm_console_proxy_test.go
codingben: Ack.

tests/vm_console_proxy_test.go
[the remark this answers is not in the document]
akrejcir: Correct. The linked part of code only adds the annotations in the `newSspStrategy`. If the tests are run with an existing SSP operator deployment (as is done by QE), then the annotations may not be there.

akrejcir:
The CI is timeouting. I've created a PR to increase the timeout: https://github.com/openshift/release/pull/36544

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -233,6 +252,17 @@ func isEnabled(request *common.Request) bool {
 	return isEnabled
 }
 
+func getVmConsoleProxyNamespace(request *common.Request) string {
+	if request.Instance.GetAnnotations() == nil {
+		return "kubevirt"
+	}
+	namespaceStr, ok := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
+	if !ok {
+		return "kubevirt"
```
akrejcir, line 261: Can you extract the `"kubevirt"` string literal to a constant? Then we can be sure that both branches return the same string, and there is not a typo.

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,188 @@
...
+		strategy.SkipSspUpdateTestsIfNeeded()
+
+		updateSsp(func(foundSsp *ssp.SSP) {
+			if foundSsp.GetAnnotations() == nil {
+				foundSsp.Annotations = make(map[string]string)
+				foundSsp.Annotations[vm_console_proxy.EnableAnnotation] = "true"
+			} else {
+				foundSsp.Annotations[vm_console_proxy.EnableAnnotation] = "true"
```
akrejcir, line 35: This line can be moved out of the `if` statement, because it is in both brnaches.

Can you also add the `VmConsoleProxyNamespaceAnnotation` here?

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,188 @@
...
+			Entry("[test_id:TODO] deployment", &deploymentResource),
+			Entry("[test_id:TODO] config map", &configMapResource),
+		)
+	})
+
+	Context("Resource change", func() {
+		DescribeTable("should restore modified resource", expectRestoreAfterUpdate,
+			Entry("[test_id:TODO] cluster role", &clusterRoleResource),
+			Entry("[test_id:TODO] cluster role binding", &clusterRoleBindingResource),
+			Entry("[test_id:TODO] service account", &serviceAccountResource),
```
akrejcir, line 154: The update test for service account currently does not do any update, so it can be removed.

internal/operands/vm-console-proxy/reconcile.go
codingben: Yep, done.

tests/vm_console_proxy_test.go
codingben: Good point, thanks!

tests/vm_console_proxy_test.go
codingben: Also in `should restore modified resource with pause`, right?

akrejcir: yes

codingben:
@akrejcir Looks like it failed again because of timeout? I see it here [1]:
```
�[38;5;9m�[1mRan 61 of 266 Specs in 3600.077 seconds�[0m
�[38;5;9m�[1mFAIL! - Interrupted by Timeout�[0m -- �[38;5;10m�[1m60 Passed�[0m | �[38;5;9m�[1m1 Failed�[0m | �[38;5;11m�[1m1 Pending�[0m | �[38;5;14m�[1m204 Skipped�[0m
```
[1] https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/498/pull-ci-kubevirt-ssp-operator-master-e2e-functests/1627959301203038208

akrejcir:
Let's try again, maybe it takes a while until the configuration is applied.

/retest

codingben:
@akrejcir Okay. I see all *-apply jobs passed https://github.com/openshift/release/commit/232d237b1000e6051da30c45c1c7b3956491bbda.

akrejcir:
The timeout was probably not set correctly. I've created another PR to fix it: https://github.com/openshift/release/pull/36609

lyarwood:
I think the timeout here is coming from ginkgo itself right?

https://onsi.github.io/ginkgo/#recommended-continuous-integration-configuration

```
--timeout allows you to specify a timeout for the ginkgo run. The default duration is one hour, which may or may not be enough!
```

The fact we are still using `go test` to launch these tests and pass a timeout of `0` (unlimited?) is likely causing this to default to the 1 hour we are seeing.

```
go test -v -coverprofile cover.out -timeout 0 ./tests/...
```

lyarwood:
https://github.com/kubevirt/ssp-operator/pull/504 - I wonder if this could help?

lyarwood:
Okay I'm pretty sure this is a change in behaviour in v2 thanks to the following docs:

https://onsi.github.io/ginkgo/MIGRATING_TO_V2#timeout-behavior

```
Timeout Behavior

In Ginkgo V1.x, Ginkgo's timeout was managed by go test. This meant that timeouts exited the test suite abruptly with 
no opportunity for custom reporters or clean up code (e.g. AfterEach, AfterSuite) to run. This is fixed in V2. Ginkgo now 
manages its own timeout and when a timeout triggers the test winds down gracefully. In fact, a timeout is now 
functionally equivalent to a user-initiated interrupt.

In addition, in V1.x when running multiple test suites Ginkgo would give each suite the full timeout allotment (so ginkgo -
r -timeout=1h would give each test suite one hour to complete). In V2 the timeout now applies to the entire test suite run 
so that ginkgo -r -timeout=1h is now guaranteed to exit after (about) one hour.
```

codingben:
@lyarwood Thanks! I think we'll be affected again when we'll add KubeVirt Tekton Tasks here.

codingben:
@dominikholler @akrejcir Do we really need "EnableAnnotation"? Why it shouldn't be enabled by default? Will the customer need to change it somehow?

akrejcir:
> Okay I'm pretty sure this is a change in behaviour in v2 thanks to the following docs:
> 
> https://onsi.github.io/ginkgo/MIGRATING_TO_V2#timeout-behavior

Thanks for finding this. I've added the `-timeout 0` parameter, because the default timeout was 10 minutes, but that was probably timeout from `go test` and not ginkgo.

akrejcir:
> @dominikholler @akrejcir Do we really need "EnableAnnotation"? Why it shouldn't be enabled by default? Will the customer need to change it somehow?

Yes, it's safer to have it disabled by default. It is a new feature that has not been tested by QE yet. There may be unexpected bugs.

codingben:
@lyarwood Done, rebased after it was merged now. @akrejcir Ack.

codingben:
I see again in the logs [1]:
```
1mFAIL! - Interrupted by Timeout
```
Seems like 2 hours of timeout isn't enough?

[1] https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/498/pull-ci-kubevirt-ssp-operator-master-e2e-functests/1628055834716016640

tests/vm_console_proxy_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	BeforeEach(func() {
+		strategy.SkipSspUpdateTestsIfNeeded()
+
+		updateSsp(func(foundSsp *ssp.SSP) {
+			if foundSsp.GetAnnotations() == nil {
+				foundSsp.Annotations = make(map[string]string)
+			}
+
+			namespace := strategy.GetVmConsoleProxyNamespace()
```
akrejcir, line 35: This line panics for update tests. They use the `existingSspStrategy`, and the annotation is not defined at the time that this line is called.

codingben: Thanks, I see. I'll investigate tomorrow to see where to define that annotation.

codingben: Isn't `existingSspStrategy` used only when ssp-operator is deployed by HCO? I don't see in the logs any indication that the tests are deployed by HCO, is it done on QE side?

If so, then I think we don't need to panic and just return the default namepsace:
```
	namespaceStr, ok := s.ssp.ObjectMeta.GetAnnotations()[vm_console_proxy.VmConsoleProxyNamespaceAnnotation]
	if !ok {
		return vm_console_proxy.DefaultNamespace // "kubevirt"
	}
	return namespaceStr
```
Am I right?

akrejcir: The `existingSspStrategy` should be used when the SSP CR already exists in the cluster. We use it for the update tests. Here is the script run by CI: https://github.com/kubevirt/ssp-operator/blob/be8d6b510301a9658018e6796a702082dbd3eafd/automation/e2e-upgrade-functests/run.sh

It does these steps:
- Install the latest released ssp-operator version
- Create an SSP CR
- Wait until ssp-operator is finished deploying all resources
- Install the ssp-operator build from the current PR
- Run the functional test suite

This way we can test if there are any problems when updating from the latest released version to the PR version.

codingben: Thanks! I see how it works now, learned something new. So in this PR we should also add new annotations to `ssp-sample.yaml` [1] right?

[1] https://github.com/kubevirt/tekton-tasks-operator/blob/main/scripts/ssp-sample.yaml

akrejcir: That file is in TTO, do you mean `config/samples/ssp_v1beta1_ssp.yaml` ? It only contains mandatory parts, so I would say don't add it there.

Can you add code to read the kubevirt namespace for tests from an env variable? This file contains env variables used by tests: https://github.com/kubevirt/ssp-operator/blob/3719aa2cf60a2f93117f02c57a380455fef56bdd/tests/env/env.go
You could add a new one and then define it in the update tests script in `automation/e2e-upgrade-functests/run.sh`

codingben: @akrejcir Ah no, I took a look on how it was implemented in TTO and here in ssp-operator. I meant to [1] of course, wrong copy-paste.

[1] https://github.com/kubevirt/ssp-operator/blob/master/automation/e2e-upgrade-functests/ssp-cr-template.yaml

codingben: So `func (s *existingSspStrategy) GetVmConsoleProxyNamespace() string` should read it from env var and not from an annotation?

akrejcir: No, in case the annotation is not defined, then it should read env variable.

akrejcir: > @akrejcir Ah no, I took a look on how it was implemented in TTO and here in ssp-operator. I meant to [1] of course, wrong copy-paste.
> 
> [1] https://github.com/kubevirt/ssp-operator/blob/master/automation/e2e-upgrade-functests/ssp-cr-template.yaml

Right. But no, don't edit this file. We still need the logic with the env variable for QE tests.

codingben: Ack, thanks. Done. Thanks a lot for helping here, really appreciated. :)

codingben:
Okay so it failed again due timeout:
```
Process did not finish before 2h0m0s timeout
```

akrejcir:
I will increase the timeout to a large value and then we can lower it later.

codingben:
> I will increase the timeout to a large value and then we can lower it later.

Ack, sounds good.

akrejcir:
https://github.com/openshift/release/pull/36695

codingben:
Looks like it failed because of some internal error [1]:
```
Status: "Failure",
              Message: "Internal error occurred: failed calling webhook \"validation.ssp.kubevirt.io\": failed to call webhook: Post \"https://ssp-webhook-service.kubevirt.svc:443/validate-ssp-kubevirt-io-v1beta1-ssp?timeout=10s\": no endpoints available for service \"ssp-webhook-service\"",
              Reason: "InternalError",
````
[1] https://storage.googleapis.com/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/498/pull-ci-kubevirt-ssp-operator-master-e2e-upgrade-functests/1628771659840229376/build-log.txt

File: internal/operands/vm-console-proxy/reconcile.go [hunk 13]
```diff
@@ -118,10 +123,17 @@ func (v *vmConsoleProxy) Cleanup(request *common.Request) ([]common.CleanupResul
 
 func reconcileServiceAccountsFuncs(serviceAccount core.ServiceAccount) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		serviceAccount.Namespace = kubeVirtNamespace
+		serviceAccount.Namespace = getVmConsoleProxyNamespace(request)
 		return common.CreateOrUpdate(request).
 			ClusterResource(&serviceAccount).
 			WithAppLabels(operandName, operandComponent).
+			UpdateFunc(func(newRes, foundRes client.Object) {
```
akrejcir, line 130: I was mistaken about this change. The cluster adds a secret to the service account by default. This default field causes a reconciliation loop, and is removed again here in this function.

The loop means that SSP is always in a `Deploying` state. Curiously, the functional tests should fail after 10 minutes if `Deployed` state is not reached, but because of a race, sometimes the `Deployed` state is seen and that is enough for the test to progress.

Please remove this `UpdateFunc()`.

Makefile
```diff
@@ -93,7 +93,7 @@ build-functests:
 	go test -c ./tests
 
 GINKGO_VERSION ?= v2.1.4
-GINKGO_TIMEOUT ?= 2h
+GINKGO_TIMEOUT ?= 8h
```
akrejcir, line 96: This increase is not needed.

internal/operands/vm-console-proxy/reconcile.go
codingben: Done.

codingben:
It's failing now because of some weird error:
```
{  could not initialize namespace: failed to wait for authentication cache to warm up after 15s: projects.project.openshift.io "ci-op-ngsfcfg0" is forbidden: User "system:serviceaccount:ci:ci-operator" cannot get resource "projects" in API group "project.openshift.io" in the namespace "ci-op-ngsfcfg0"}
```

akrejcir:
Retest will not help. I'm looking into it.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -30,6 +31,17 @@ const (
 // +kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
 // +kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
 
+func WatchClusterTypes() []operands.WatchType {
+	return []operands.WatchType{
+		{Object: &rbac.ClusterRole{}},
+		{Object: &rbac.ClusterRoleBinding{}},
+		{Object: &core.ServiceAccount{}},
+		{Object: &core.Service{}},
+		{Object: &apps.Deployment{}},
```
akrejcir, line 40: The Deployment needs to watch full resource, because it reacts to `.spec` changes:
```
{Object: &apps.Deployment{}, WatchFullObject: true},
```

akrejcir:
And just to be safe of unexpected changes, can you make a DeepCopy() of all the objects used in `Reconcile()` function?
The shallow copies that are used now contain pointers, and can save some state from one reconciliation loop to another.

internal/operands/vm-console-proxy/reconcile.go
codingben: Done.

codingben:
@akrejcir Thanks!

PR #504, opened 2023-02-21
functest: Switch to calling ginkgo using go run

lyarwood:
/cc @akrejcir 
/cc @codingben 
/cc @0xFelix 

Additional `ginkgo` and `getginkgo` targets are also introduce to help with updating and vendoring ginkgo in the future. They also ensure the expected version is present before running func tests.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,9 +92,19 @@ unittest: generate fmt vet manifests
 build-functests:
 	go test -c ./tests
 
+GINKGO_VERSION ?= v2.1.4
+GINKGO_TIMEOUT ?= 2h
+
+.PHONY: ginkgo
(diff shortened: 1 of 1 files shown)


lyarwood:
If this works for people I can also switch over `unittest`, seems there's some redundant stuff in there anyway as we don't have any tests under `./api/...` etc.

codingben (approved):
/lgtm

lyarwood:
/retest-required

Some very weird and unseen AFAIK failures in the failed functest runs.

akrejcir:
Thanks for finding the timeout issue.

/lgtm

PR #509, opened 2023-03-01
test: add and cleanup vm-console-proxy operand

codingben:

Add and cleanup operand unit tests.

**Release note**:

```release-note
Add and cleanup vm-console-proxy operand unit tests
```

### DIFF
--- a/internal/operands/vm-console-proxy/reconcile_test.go
+++ b/internal/operands/vm-console-proxy/reconcile_test.go
@@ -17,13 +17,16 @@ import (
 	kubevirt "kubevirt.io/api/core"
 	ssp "kubevirt.io/ssp-operator/api/v1beta1"
 	"kubevirt.io/ssp-operator/internal/common"
+	. "kubevirt.io/ssp-operator/internal/test-utils"
 	vm_console_proxy_bundle "kubevirt.io/ssp-operator/internal/vm-console-proxy-bundle"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
+	namespace              = "kubevirt"
 	name                   = "vm-console-proxy"
 	vmConsoleProxyName     = "vm-console-proxy"
 	clusterRoleName        = "vm-console-proxy"
@@ -35,33 +38,168 @@ const (
 )
 
 var _ = Describe("VM Console Proxy Operand", func() {
-	var v *vmConsoleProxy
-	var mockedRequest *common.Request
+	const (
+		replicas int32 = 1
+	)
+
+	var (
+		bundle  *vm_console_proxy_bundle.Bundle
+		operand *vmConsoleProxy
+		request common.Request
+	)
 
 	BeforeEach(func() {
-		v = getMockedVmConsoleProxyOperand()
-		mockedRequest = getMockedRequest()
-	})
-
-	It("should return new with test bundle correctly", func() {
-		vmConsoleProxy := New(getMockedTestBundle())
-		Expect(vmConsoleProxy.serviceAccount).ToNot(BeNil(), "service account should not be nil")
-		Expect(vmConsoleProxy.clusterRole).ToNot(BeNil(), "cluster role should not be nil")
-		Expect(vmConsoleProxy.clusterRoleBinding).ToNot(BeNil(), "cluster role binding should not be nil")
-		Expect(vmConsoleProxy.service).ToNot(BeNil(), "service should not be nil")
-		Expect(vmConsoleProxy.deployment).ToNot(BeNil(), "deployment should not be nil")
-		Expect(vmConsoleProxy.configMap).ToNot(BeNil(), "config map should not be nil")
+		bundle = getMockedTestBundle()
+		operand = New(bundle)
+		request = getMockedRequest()
 	})
 
 	It("should return name correctly", func() {
-		name := v.Name()
+		name := operand.Name()
 		Expect(name).To(Equal(operandName), "should return correct name")
 	})
 
 	It("should return functions from reconcile correclty", func() {
-		functions, err := v.Reconcile(mockedRequest)
+		functions, err := operand.Reconcile(&request)
 		Expect(err).ToNot(HaveOccurred(), "should not throw err")
 		Expect(len(functions)).To(Equal(7), "should return correct number of reconcile functions")
+	})
+
+	It("should create vm-console-proxy resources", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(&bundle.ServiceAccount, request)
+		ExpectResourceExists(&bundle.ClusterRole, request)
+		ExpectResourceExists(&bundle.ClusterRoleBinding, request)
+		ExpectResourceExists(&bundle.ConfigMap, request)
+		ExpectResourceExists(&bundle.Service, request)
+		ExpectResourceExists(&bundle.Deployment, request)
+		ExpectResourceExists(newRoute(namespace, serviceName), request)
+	})
+
+	It("should remove cluster resources on cleanup", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(&bundle.ServiceAccount, request)
+		ExpectResourceExists(&bundle.ClusterRole, request)
+		ExpectResourceExists(&bundle.ClusterRoleBinding, request)
+		ExpectResourceExists(&bundle.ConfigMap, request)
+		ExpectResourceExists(&bundle.Service, request)
+		ExpectResourceExists(&bundle.Deployment, request)
+		ExpectResourceExists(newRoute(namespace, serviceName), request)
+
+		_, err = operand.Cleanup(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceNotExists(&bundle.ServiceAccount, request)
+		ExpectResourceNotExists(&bundle.ClusterRole, request)
+		ExpectResourceNotExists(&bundle.ClusterRoleBinding, request)
+		ExpectResourceNotExists(&bundle.ConfigMap, request)
+		ExpectResourceNotExists(&bundle.Service, request)
+		ExpectResourceNotExists(&bundle.Deployment, request)
+		ExpectResourceNotExists(newRoute(namespace, serviceName), request)
+	})
+
+	It("should not update service cluster IP", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		key := client.ObjectKeyFromObject(&bundle.Service)
+		service := &core.Service{}
+		Expect(request.Client.Get(request.Context, key, service)).ToNot(HaveOccurred())
+
+		// This address is from a range of IP addresses reserved for documentation.
+		const testClusterIp = "198.51.100.42"
+
+		service.Spec.ClusterIP = testClusterIp
+		Expect(request.Client.Update(request.Context, service)).ToNot(HaveOccurred())
+
+		_, err = operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		updatedService := &core.Service{}
+		Expect(request.Client.Get(request.Context, key, updatedService)).ToNot(HaveOccurred())
+		Expect(updatedService.Spec.ClusterIP).To(Equal(testClusterIp))
+	})
+
+	It("should report status", func() {
+		reconcileResults, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Set status for deployment
+		key := client.ObjectKeyFromObject(&bundle.Deployment)
+		updateDeployment(key, &request, func(deployment *apps.Deployment) {
+			deployment.Status.Replicas = replicas
+			deployment.Status.ReadyReplicas = 0
+			deployment.Status.AvailableReplicas = 0
+			deployment.Status.UpdatedReplicas = 0
+			deployment.Status.UnavailableReplicas = replicas
+		})
+
+		reconcileResults, err = operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Only deployment should be progressing
+		for _, reconcileResult := range reconcileResults {
+			if _, ok := reconcileResult.Resource.(*apps.Deployment); ok {
+				Expect(reconcileResult.Status.NotAvailable).ToNot(BeNil())
+				Expect(reconcileResult.Status.Progressing).ToNot(BeNil())
+				Expect(reconcileResult.Status.Degraded).ToNot(BeNil())
+			} else {
+				Expect(reconcileResult.Status.NotAvailable).To(BeNil())
+				Expect(reconcileResult.Status.Progressing).To(BeNil())
+				Expect(reconcileResult.Status.Degraded).To(BeNil())
+			}
+		}
+
+		updateDeployment(key, &request, func(deployment *apps.Deployment) {
+			deployment.Status.Replicas = replicas
+			deployment.Status.ReadyReplicas = replicas
+			deployment.Status.AvailableReplicas = replicas
+			deployment.Status.UpdatedReplicas = replicas
+			deployment.Status.UnavailableReplicas = 0
+		})
+
+		reconcileResults, err = operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		// All resources should be available
+		for _, reconcileResult := range reconcileResults {
+			Expect(reconcileResult.Status.NotAvailable).To(BeNil())
+			Expect(reconcileResult.Status.Progressing).To(BeNil())
+			Expect(reconcileResult.Status.Degraded).To(BeNil())
+		}
+	})
+
+	It("should deploy resources in namespace provided by annotation", func() {
+		namespace := "some-namespace"
+		request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation] = namespace
+
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectNamespacedResourceExists(&bundle.ServiceAccount, request, namespace)
+		ExpectNamespacedResourceExists(&bundle.ConfigMap, request, namespace)
+		ExpectNamespacedResourceExists(&bundle.Service, request, namespace)
+		ExpectNamespacedResourceExists(&bundle.Deployment, request, namespace)
+		ExpectNamespacedResourceExists(newRoute(getVmConsoleProxyNamespace(&request), serviceName), request, namespace)
+	})
+
+	It("should delete resources when enabled annotation is removed", func() {
+		delete(request.Instance.Annotations, EnableAnnotation)
+
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceNotExists(&bundle.ServiceAccount, request)
+		ExpectResourceNotExists(&bundle.ClusterRole, request)
+		ExpectResourceNotExists(&bundle.ClusterRoleBinding, request)
+		ExpectResourceNotExists(&bundle.ConfigMap, request)
+		ExpectResourceNotExists(&bundle.Service, request)
+		ExpectResourceNotExists(&bundle.Deployment, request)
+		ExpectResourceNotExists(newRoute(namespace, serviceName), request)
 	})
 })
 
@@ -70,14 +208,14 @@ func TestVmConsoleProxyBundle(t *testing.T) {
 	RunSpecs(t, "VM Console Proxy Operand Suite")
 }
 
-func getMockedRequest() *common.Request {
-	var log = logf.Log.WithName("metrics_operand")
+func getMockedRequest() common.Request {
+	var log = logf.Log.WithName("vm_console_proxy_operand")
 	client := fake.NewClientBuilder().WithScheme(common.Scheme).Build()
 
-	return &common.Request{
+	return common.Request{
 		Request: reconcile.Request{
 			NamespacedName: types.NamespacedName{
-				Namespace: "kubevirt",
+				Namespace: namespace,
 				Name:      name,
 			},
 		},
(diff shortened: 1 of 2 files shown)


codingben:
@akrejcir FYI, it can be reviewed except an issue in `should create vm-console-proxy resources`.

internal/operands/vm-console-proxy/resources.go
```diff
@@ -0,0 +1,226 @@
+package vm_console_proxy
```
akrejcir, line 1: The resources for `vm-console-proxy` are loaded from a yaml file, so they should not be hardcoded here.
I think you can remove this file.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -203,7 +203,9 @@ func reconcileDeploymentFuncs(deployment apps.Deployment) common.ReconcileFunc {
 			ClusterResource(&deployment).
 			WithAppLabels(operandName, operandComponent).
 			UpdateFunc(func(newRes, foundRes client.Object) {
-				foundRes.(*apps.Deployment).Spec = newRes.(*apps.Deployment).Spec
+				newDeployment := newRes.(*apps.Deployment)
+				foundDeployment := foundRes.(*apps.Deployment)
+				foundDeployment.Spec = newDeployment.Spec
```
akrejcir, line 208: This change is not needed. The one line was clear enough.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -228,30 +230,28 @@ func reconcileDeploymentFuncs(deployment apps.Deployment) common.ReconcileFunc {
...
-	if err != nil {
-		return false
+	if request.Instance.GetAnnotations() != nil {
+		enable, isFound := request.Instance.GetAnnotations()[EnableAnnotation]
+		if isFound {
+			isEnabled, err := strconv.ParseBool(enable)
+			if err == nil {
+				return isEnabled
+			}
+		}
```
akrejcir, line 240: I think this change is not needed. Early `return`s are easier to read than nested `if`s.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -228,30 +230,28 @@ func reconcileDeploymentFuncs(deployment apps.Deployment) common.ReconcileFunc {
...
-		return defaultNamespace
-	}
-	namespaceStr, ok := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
-	if !ok {
-		return defaultNamespace
+	if request.Instance.GetAnnotations() != nil {
+		namespace, isFound := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
+		if isFound {
+			return namespace
+		}
```
akrejcir, line 251: This change is ok, I think.

akrejcir (commented):
Can you also cleanup the names of functions?

internal/operands/vm-console-proxy/reconcile.go
codingben: I agree, but isn't makes us to add another unit test, for that early return?

akrejcir: In general, unit tests should check the **functionality** of a code "unit", usually a `struct` and it's public methods.
So a code change that does not change functionality, should not change unit tests. In fact unit tests are useful to verify that the changed code has the same functionality.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -252,30 +252,28 @@ func reconcileRoute(serviceName string) common.ReconcileFunc {
...
 func getVmConsoleProxyNamespace(request *common.Request) string {
-	const defaultNamespace = "kubevirt"
-	if request.Instance.GetAnnotations() == nil {
-		return defaultNamespace
-	}
-	namespaceStr, ok := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
-	if !ok {
-		return defaultNamespace
+	if request.Instance.GetAnnotations() != nil {
+		namespace, isFound := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
```
0xFelix, line 270: You could merge L270 and 271.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -54,31 +56,236 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
 	})
 
 	It("should return functions from reconcile correclty", func() {
-		functions, err := v.Reconcile(mockedRequest)
+		functions, err := operand.Reconcile(&request)
 		Expect(err).ToNot(HaveOccurred(), "should not throw err")
 		Expect(len(functions)).To(Equal(7), "should return correct number of reconcile functions")
 	})
+
+	It("should create vm-console-proxy resources", func() {
```
0xFelix, line 69: Strictly speaking this is not cleaning up existing code but adding new test code. Could you please put it into a separate commit?

akrejcir: Ok, then please use this PR to add new unit tests and you can drop the refactoring changes.

codingben: Currently some new unit tests depend one refactored functions, `isEnabled` and `getVmConsoleProxyNamespace`. Otherwise, unit tests won't fit there.

internal/operands/vm-console-proxy/reconcile.go
codingben: Do you mean:
```
if namespace, isFound := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]; isFound {
			return namespace
		}
```
If yes, is it common code style? Because I don't like it and feel it's line too long for good code readability.

akrejcir: It is common, but It's up to you to decide. We don't have guidelines about this.

codingben:
I'll rename that PR to add only new tests, and open another PR to have refactored functions.

internal/operands/vm-console-proxy/resources.go
codingben: Done, thanks!

codingben:
This PR is ready for review, except some issue with the commented lines:
```
It("should remove cluster resources on cleanup", func() {
...
		// ExpectResourceNotExists(bundle.serviceAccount, request)
		ExpectResourceNotExists(bundle.clusterRole, request)
		ExpectResourceNotExists(bundle.clusterRoleBinding, request)
		// ExpectResourceNotExists(bundle.configMap, request)
		// ExpectResourceNotExists(bundle.service, request)
		// ExpectResourceNotExists(bundle.deployment, request)
		ExpectResourceNotExists(newRoute(namespace, "vm-console-proxy"), request)
}
```
It throws nil error when I try to test it.

/cc @akrejcir @0xFelix

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -17,13 +18,17 @@ import (
...
+	. "kubevirt.io/ssp-operator/internal/test-utils"
 	vm_console_proxy_bundle "kubevirt.io/ssp-operator/internal/vm-console-proxy-bundle"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const (
+	containerPort          = 8768
```
akrejcir, line 30: This const is not used.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,12 +40,20 @@ const (
...
 
 var _ = Describe("VM Console Proxy Operand", func() {
-	var v *vmConsoleProxy
-	var mockedRequest *common.Request
+	const (
+		replicas int32 = 1
+	)
+
+	var (
+		bundle  *vmConsoleProxy
```
akrejcir, line 48: `bundle` should have type `*vm_console_proxy_bundle.Bundle`

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,12 +40,20 @@ const (
...
+		operand *vmConsoleProxy
+		request common.Request
+	)
 
 	BeforeEach(func() {
-		v = getMockedVmConsoleProxyOperand()
-		mockedRequest = getMockedRequest()
+		bundle = New(getMockedTestBundle())
+		operand = getMockedVmConsoleProxyOperand()
```
akrejcir, line 55: Please use `New()` to create the operand. It doesn't make sense to mock and object that is being tested.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,12 +40,20 @@ const (
...
+		request = getMockedRequest()
 	})
 
 	It("should return new with test bundle correctly", func() {
```
akrejcir, line 59: Please remove this test, because it only tests private fields. Usually unit tests are only for exported functionality.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -54,30 +67,234 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+
+		// All resources should be available
+		for _, reconcileResult := range reconcileResults {
+			Expect(reconcileResult.Status.NotAvailable).To(BeNil())
+			Expect(reconcileResult.Status.Progressing).To(BeNil())
+			Expect(reconcileResult.Status.Degraded).To(BeNil())
+		}
+	})
+
+	Describe("GetVmConsoleProxyImage", func() {
```
akrejcir, line 188: Unit tests should not test private functions. Can you rewrite this in terms of the created deployment?
Or you can just remove it.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -54,30 +67,234 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+		})
+
+		It("should return the default value if VM_CONSOLE_PROXY_IMAGE not set", func() {
+			defaultImage := "quay.io/kubevirt/vm-console-proxy:v0.1.0"
+			actualImage := getVmConsoleProxyImage()
+			Expect(actualImage).To(Equal(defaultImage))
+		})
+	})
+
+	Describe("GetVmConsoleProxyNamespace", func() {
```
akrejcir, line 204: Similarly, these tests also call private function.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -54,30 +67,234 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+						},
+					},
+				},
+			}
+			actualNamespace := getVmConsoleProxyNamespace(request)
+			Expect(actualNamespace).To(Equal(customNamespace))
+		})
+	})
+
+	Describe("IsEnabled", func() {
```
akrejcir, line 245: The same here.

akrejcir (commented):
A partial review. 
In general, unit tests don't need to test every single private function. Private functions can change. Tests should cover visible behavior of a module:
- Exported functions and methods
- Exported structs
- Exported fields of structs

In my opinion, an exception to this guideline is if a private function has complex implementation and we want to make sure that it works.

codingben:
@akrejcir Ah okay :) I'm so sorry. I never wrote before operator tests. I thought we should test all operand functions because it plays critical role in deploying `vm-console-proxy`, it's not just another library. Thanks!

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -54,30 +67,234 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+		ExpectResourceExists(bundle.clusterRoleBinding, request)
+		ExpectResourceExists(bundle.configMap, request)
+		ExpectResourceExists(bundle.service, request)
+		ExpectResourceExists(bundle.deployment, request)
+		ExpectResourceExists(newRoute(namespace, "vm-console-proxy"), request)
+
+		_, err = operand.Cleanup(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		// ExpectResourceNotExists(bundle.serviceAccount, request)
```
0xFelix, line 108: Can you please investigate what makes these calls fail? Commented code IMO is unclean.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -54,30 +67,234 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(bundle.serviceAccount, request)
+		ExpectResourceExists(bundle.clusterRole, request)
+		ExpectResourceExists(bundle.clusterRoleBinding, request)
+		ExpectResourceExists(bundle.configMap, request)
+		ExpectResourceExists(bundle.service, request)
+		ExpectResourceExists(bundle.deployment, request)
+		ExpectResourceExists(newRoute(namespace, "vm-console-proxy"), request)
```
0xFelix, line 103: You might use a const for the route name in this spec.

internal/operands/vm-console-proxy/reconcile_test.go
0xFelix: IMO if a PR gets large, multiple commits help to reduce the cognitive load of reviewing the PR significantly. Too many changes in a single commit make PRs very hard to review.

internal/operands/vm-console-proxy/reconcile_test.go
codingben: Yep, thanks to Andrej's review, I saw that I was using wrong resource reference.

codingben:
@akrejcir @0xFelix Thanks a lot for your reviews, really appreciated. It should be okay now. :)

codingben:
> Can you also cleanup the names of functions?

Sorry, what do you mean?

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,49 +38,153 @@ const (
...
+	It("should create vm-console-proxy resources", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(operand.serviceAccount, request)
+		ExpectResourceExists(operand.clusterRole, request)
+		ExpectResourceExists(operand.clusterRoleBinding, request)
+		ExpectResourceExists(operand.configMap, request)
+		ExpectResourceExists(operand.service, request)
+		ExpectResourceExists(operand.deployment, request)
```
akrejcir, line 75: It's better if these tests check the resources from the mocked bundle, and don't access private fields of the operand. Because private fields can change, and unit tests should not depend on them.

internal/operands/vm-console-proxy/reconcile_test.go
akrejcir: +1

akrejcir:
> > Can you also cleanup the names of functions?
> 
> @akrejcir Sorry, what do you mean?

I meant that function names in `reconcile.go` can be made clearer. For example `reconcileServiceAccountsFuncs()` can just be `reconcileServiceAccounts()`. But now that you've split this PR into 2, that comment is relevant to the other PR.

akrejcir:
Can you add unit tests for the annotations functionality? To test these cases:
- proxy is deployed to the namespace specified by the `vm-console-proxy-namespace` annotation
- when `vm-console-proxy-enabled` annotation is removed, the proxy resources are deleted

internal/operands/vm-console-proxy/reconcile_test.go
codingben: @akrejcir I already changed it. The parameter `vm-console-proxy` in `newRoute(namespace string, serviceName string)` isn't route name but service name.

akrejcir: Right, sorry. I just realized that service name is irrelevant for this test.

internal/operands/vm-console-proxy/reconcile_test.go
codingben: What do you mean by "resources from the mocked bundle", by loading it from `vm-console-proxy-bundle-test.yaml` or just a new function?

akrejcir: By "mocked bundle", I mean the `*vm_console_proxy_bundle.Bundle` returned by `getMockedTestBundle()`.
So for example, in `BeforeEach()` you could store the bundle into a variable, and then here check that the resource from the bundle was created by the operand:
```go
ExpectResourceExists(bundle.Deployment, request)
```

This way the test does not access any private fields.

codingben: @akrejcir Yes, got it. I think this is what I did initially but then changed it yesterday.

codingben: Done.

codingben:
> proxy is deployed to the namespace specified by the vm-console-proxy-namespace annotation

Done, please review.

> when vm-console-proxy-enabled annotation is removed, the proxy resources are deleted

@akrejcir Done, please review as well.

But, what is the purpose of that test? Isn't similar to `should remove cluster resources on cleanup`? I mean, it sounds like it's the same test as `should remove cluster resources on cleanup` and we just need to remove annotation by:
```
request.Instance.Annotations = map[string]string{
    VmConsoleProxyNamespaceAnnotation: namespace,
}
```
Before we call to Cleanup?

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,49 +38,187 @@ const (
...
+			Expect(reconcileResult.Status.Progressing).To(BeNil())
+			Expect(reconcileResult.Status.Degraded).To(BeNil())
+		}
+	})
+
+	It("should deploy resources in namespace provided by annotation", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		namespace := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]
```
akrejcir, line 180: Instead of getting the namespace, can you change the value of the annotation to something different?
And then after calling `operand.Reconcile()`, test that the resources are deployed in that namespace.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,49 +38,187 @@ const (
...
+		ExpectNamespacedResourceExists(&bundle.ServiceAccount, request, namespace)
```
akrejcir, line 181: Function `ExpectNamespacedResourceExists()` is not defined anywhere.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,49 +38,187 @@ const (
...
+		ExpectNamespacedResourceExists(newRoute(getVmConsoleProxyNamespace(&request), serviceName), request, namespace)
+	})
+
+	It("should delete resources when enabled annotation is removed", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		request.Instance.Annotations = map[string]string{
+			VmConsoleProxyNamespaceAnnotation: namespace,
+		}
```
akrejcir, line 194: Instead of setting the annotation map, can you just delete the `EnableAnnotation`?
```go
delete(request.Instance.Annotations, EnableAnnotation)
```

Then it will be more obvious, that this test removes an annotation.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -35,49 +38,187 @@ const (
...
+
+		_, err = operand.Cleanup(&request)
+		Expect(err).ToNot(HaveOccurred())
```
akrejcir, line 197: `Cleanup()` is normally called when the SSP CR is removed from the cluster.
This test only tests removing an annotation, not the whole CR, so it should call `Reconcile()` here.

internal/operands/vm-console-proxy/reconcile_test.go
codingben: It wasn't added to the commit, but `ExpectNamespacedResourceExists is part of test_utils.go`.

internal/test-utils/test_utils.go
```diff
@@ -19,3 +19,10 @@ func ExpectResourceNotExists(resource client.Object, request common.Request) {
 	ExpectWithOffset(1, err).To(HaveOccurred())
 	ExpectWithOffset(1, errors.IsNotFound(err)).To(BeTrue())
 }
+
+func ExpectNamespacedResourceExists(resource client.Object, request common.Request, namespace string) {
+	key := client.ObjectKeyFromObject(resource)
+	err := request.Client.Get(request.Context, key, resource)
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
```
0xFelix, line 26: IMO either use `...WithOffset` in all assertions or in none. I'd vote for using the standard `Expect`.

akrejcir: +1 for the consistency. 
But in this case using no offset will generate useless error message, because it will point to this line without any information about the calling function or the values of parameters passed.

internal/operands/vm-console-proxy/reconcile_test.go
codingben: Since I'm doing: `ExpectNamespacedResourceExists(&bundle.Deployment, request, namespace)` it doesn't work, I think this is because `&bundle.Deployment` contains value of `namespace` from constant variable.

codingben: Then we should do: `ExpectNamespacedResourceExists(operand.serviceAccount, request, namespace)` to use changed resource in `operand.serviceAccount`?

internal/operands/vm-console-proxy/reconcile_test.go
codingben: Done, thanks!

internal/operands/vm-console-proxy/reconcile_test.go
codingben: Done, thanks! :)

internal/operands/vm-console-proxy/reconcile_test.go
codingben: I'm trying to do it, but for some reason it doesn't work, I'm spending some time to debug it.

codingben: Done, thanks @akrejcir.

codingben:
@akrejcir Seems like approval is missing and it wasn't merged because of it.

PR #510, opened 2023-03-01
fix: Add validation for vm-console-proxy annotations to webhook

akrejcir:
Webhook needs to validate the annotations:
- `ssp.kubevirt.io/vm-console-proxy-enabled`
- `ssp.kubevirt.io/vm-console-proxy-namespace`

**Release note**:
```release-note
None
```


akrejcir:
Is not needed anymore, because the validation webhook will be part or API change.

/close

PR #513, opened 2023-03-07
fix: Wait for SSP operator to be ready after TLS policy tests

akrejcir:
The TLS policy tests cause the operator to enter crash loop. This commit adds code that waits until it is ready.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/crypto_policy_test.go
+++ b/tests/crypto_policy_test.go
@@ -7,11 +7,14 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
 	ocpv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/crypto"
+	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	sspv1beta1 "kubevirt.io/ssp-operator/api/v1beta1"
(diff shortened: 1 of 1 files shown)


tests/crypto_policy_test.go
```diff
@@ -103,6 +106,17 @@ var _ = Describe("Crypto Policy", func() {
 
 	AfterEach(func() {
 		strategy.RevertToOriginalSspCr()
+
+		// The SSP operator may enter crash loop backoff state, so we need to wait until it is running
```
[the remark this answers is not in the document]
akrejcir: Done

tests/crypto_policy_test.go
```diff
@@ -103,6 +106,19 @@ var _ = Describe("Crypto Policy", func() {
 
 	AfterEach(func() {
 		strategy.RevertToOriginalSspCr()
+
+		// Because of bug[1], the SSP operator will move to CrashLoopBackOff state,
```
lyarwood, line 110: supernit - I forget what the rules are for FIXMEs in this codebase but this strikes me as something we could mark as FIXME once that bug is resolved?

akrejcir: There are no rules. Do you have a suggestion for what is a good way to mark as FIXME? Or simply prefix the comment with `FIXME:` ?

dominikholler:
@akrejcir Please do not mention the bug in the commit message or in the PR description, because this would make the automation assume that this PR/commit would contain already the fix the the bug.

0xFelix:
I think we have to reconsider how the automation works. If we exactly know the issue (speaking of a known bug), we should be able to add a direct reference to it. Does the automation really pick up links in the code too? Or just in the description of a commit?

0xFelix (approved):
/approve

PR #495, opened 2023-02-10
lint: Introduce Makefile target, introduce pre-commit config and fix issues

lyarwood:
**What this PR does / why we need it**:

Adds a `lint` Makefile target using `golangci-lint` initially with the following plugins enabled:

* whitespace
* misspell
* ginkgolinter
* errcheck
* ineffassign
* staticcheck
* unused
* errname

Additionally provides a `pre-commit` config users can opt into using to help catch issues before committing changes locally.

**Which issue(s) this PR fixes**: 

Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #514, opened 2023-03-08
fix: Use defer when unlocking a lock

akrejcir:
A mutex was incorrectly unlocked right after locking it.

**Release note**:
```release-note
None
```


lyarwood:
It's already fixed in https://github.com/kubevirt/ssp-operator/pull/495 so I'm not sure why you've created this PR?

PR #516, opened 2023-03-09
fix: Update Kubevirt CR used in CI

akrejcir:
Enable `KubevirtSeccompProfile` feature gate, so VM can be started on OCP 4.13.

**Release note**:
```release-note
None
```

### DIFF
--- a/automation/common/deploy-kubevirt-and-cdi.sh
+++ b/automation/common/deploy-kubevirt-and-cdi.sh
@@ -15,9 +15,9 @@ EOF
 # Deploying kuebvirt
 oc apply -n $NAMESPACE -f "https://github.com/kubevirt/kubevirt/releases/download/${KUBEVIRT_VERSION}/kubevirt-operator.yaml"
 
-# Using KubeVirt CR from version v0.35.0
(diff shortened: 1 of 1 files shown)


automation/common/deploy-kubevirt-and-cdi.sh
```diff
@@ -15,27 +15,28 @@ EOF
...
   namespace: kubevirt
 spec:
   certificateRotateStrategy: {}
   configuration:
-    selinuxLauncherType: "virt_launcher.process"
     developerConfiguration:
       featureGates:
         - DataVolumes
         - CPUManager
         - LiveMigration
```
lyarwood, line 32: nit - Do we really need `LiveMigration`?

lyarwood (approved):
/lgtm

Quick nit about the need for the LM feature gate but this LGTM in the short term.

Longer term I'd prefer it if we used the [example CR provided with the release](https://github.com/kubevirt/kubevirt/releases/download/0.59.0/kubevirt-cr.yaml) and patched in anything missing like this but that's something for a follow up so we can unblock 4.13 CI.

PR #524, opened 2023-03-14
Use `RUNBOOK_URL_TEMPLATE` env for the runbooks URL template

assafad:
When `RUNBOOK_URL_TEMPLATE` env variable is set, its value will be used as the runbooks URL template. Otherwise, keep using the default runbook URL template `https://kubevirt.io/monitoring/runbooks/%s`. 
This allows using different runbooks versions, in accordance with the installation.

Jira-ticket - https://issues.redhat.com/browse/CNV-14382

**Release note**:

```release-note
None
```

### DIFF
--- a/internal/operands/metrics/reconcile_test.go
+++ b/internal/operands/metrics/reconcile_test.go
@@ -2,6 +2,9 @@ package metrics
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"strings"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -30,7 +33,7 @@ var _ = Describe("Metrics operand", func() {
 		operand = New()
 	)
 
-	It("should create metrics resources", func() {
+	BeforeEach(func() {
 		client := fake.NewClientBuilder().WithScheme(common.Scheme).Build()
 		request = common.Request{
 			Request: reconcile.Request{
@@ -54,14 +57,55 @@ var _ = Describe("Metrics operand", func() {
 			Logger:       log,
 			VersionCache: common.VersionCache{},
 		}
+	})
 
+	AfterEach(func() {
+		os.Unsetenv(runbookURLTemplateEnv)
+	})
+
+	It("should create metrics resources", func() {
 		_, err := operand.Reconcile(&request)
 		Expect(err).ToNot(HaveOccurred())
-		ExpectResourceExists(newPrometheusRule(namespace), request)
+
+		prometheusRule, err := newPrometheusRule(namespace)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(prometheusRule, request)
 		ExpectResourceExists(newServiceMonitorCR(namespace), request)
 		ExpectResourceExists(newMonitoringClusterRole(), request)
 		ExpectResourceExists(newMonitoringClusterRoleBinding(), request)
 	})
+
+	DescribeTable("runbook URL template",
+		func(template string) {
+			if template != defaultRunbookURLTemplate {
+				os.Setenv(runbookURLTemplateEnv, template)
+			}
+
+			prometheusRule, err := newPrometheusRule(namespace)
+
+			if strings.Count(template, "%s") != 1 || strings.Count(template, "%") != 1 {
+				Expect(err).To(HaveOccurred())
+				return
+			}
+
+			Expect(err).ToNot(HaveOccurred())
+
+			for _, group := range prometheusRule.Spec.Groups {
+				for _, rule := range group.Rules {
+					if rule.Alert != "" {
+						if rule.Annotations["runbook_url"] != "" {
+							Expect(rule.Annotations["runbook_url"]).To(Equal(fmt.Sprintf(template, rule.Alert)))
+						}
+					}
+				}
+			}
+		},
+		Entry("should use the default template when no ENV variable is set", defaultRunbookURLTemplate),
+		Entry("should use the ENV variable when a valid value is set", "valid/runbookURL/template/%s"),
+		Entry("should throw an error when the ENV variable value doesn't contain a string format placeholder", "invalid/runbookURL/template/"),
+		Entry("should throw an error when the ENV variable value contains an integer format placeholder", "invalid/runbookURL/template/%d"),
+	)
 })
 
 func TestMetrics(t *testing.T) {
--- a/internal/operands/metrics/resources.go
+++ b/internal/operands/metrics/resources.go
@@ -1,6 +1,11 @@
 package metrics
 
 import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	rbac "k8s.io/api/rbac/v1"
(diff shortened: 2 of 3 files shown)


0xFelix:
Thanks! Can you add tests for the new feature please?

akrejcir:
Does the env variable have to be a format string, or will it always be a prefix?

In my opinion, accepting a format string as an env variable doesn't seem safe. It could probably cause a panic, if the formatting parameters are incorrect.

assafad:
Blocked by https://github.com/kubevirt/monitoring/pull/165

assafad:
> Thanks! Can you add tests for the new feature please?

@0xFelix Hi, I've added some tests on [internal/operands/metrics/reconcile_test.go](https://github.com/kubevirt/ssp-operator/pull/524/files#diff-deb2db8083195e83b2f83a274276cbff858b24afa4b624078885d191e0876144). Beware that we also have [this test](https://github.com/kubevirt/ssp-operator/blob/master/tests/metrics_test.go#L261), that checks if the runbooks are accessible. Please let me know if the new tests make sense.

assafad:
> Does the env variable have to be a format string, or will it always be a prefix?
> 
> In my opinion, accepting a format string as an env variable doesn't seem safe. It could probably cause a panic, if the formatting parameters are incorrect.

@akrejcir Not sure I fully understand your concern. We don't assume that `RUNBOOK_URL_TEMPLATE` is a string (`os.LookUpEnv()` returns a string), and don't assume it's a prefix. We only verify it contains exactly 1 `%s`, as different installations may use different runbooks versions, and we can't assume nothing other than that the runbook URLs will contain the alert name.

0xFelix:
@akrejcir wanted to point out that you can inject arbitrary format specifiers when reading `RUNBOOK_URL_TEMPLATE` from the envs and using it as a format string (e.g. `https://kubevirt.io/monitoring/runbooks/%d`, what will happen?). Can't you make a prefix instead, so it is safe to read from the envs?

assafad:
> @akrejcir wanted to point out that you can inject arbitrary format specifiers when reading `RUNBOOK_URL_TEMPLATE` from the envs and using it as a format string (e.g. `https://kubevirt.io/monitoring/runbooks/%d`, what will happen?). Can't you make it a prefix instead, so it is safe to read from the envs?

@0xFelix, @akrejcir  For now, U/S URL template is indeed a prefix, but different installations/versions have different documentation requirements. Thus, assuming the alerts names are suffix is not generic enough and might require future updates. I validated [here](https://github.com/kubevirt/ssp-operator/blob/81feb52532a4af34d2e478dedb2e4919ffae7e00/internal/operands/metrics/resources.go#L294) that `RUNBOOK_URL_TEMPLATE` value contains exactly 1 `%s`, and in addition we have [this test](https://github.com/kubevirt/ssp-operator/blob/master/tests/metrics_test.go#L267) to verify that the runbooks are accessible.

akrejcir:
> @akrejcir @assafad So it will not panic but instead will fail silently? IIUC the formatted string will contain the error message, but we don't check that?

Right, it's not a good solution.

@assafad, can you change the code, so that the env variable `RUNBOOK_URL_TEMPLATE` will contain our own simple template format with a placeholder, that will be replaced using `strings.Replace()`.
For example:
```
https://kubevirt.io/monitoring/runbooks/{}
```
Where `{}` will be replaced.

internal/operands/metrics/resources.go
```diff
@@ -268,9 +278,22 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
...
 }
+
+func getRunbookURLTemplate() string {
+	runbookURLTemplate, exists := os.LookupEnv(runbookURLTemplateEnv)
+	if !exists {
+		runbookURLTemplate = defaultRunbookURLTemplate
+	}
+
+	if strings.Count(runbookURLTemplate, "%s") != 1 || strings.Count(runbookURLTemplate, "%") != 1 {
+		panic(errors.New("runbook URL template must have exactly 1 %s substring"))
```
0xFelix, line 295: Can you return an error instead of panicking? The error is caused by user input and that could be handled.

assafad: Updated, please let me know if the change makes sense

internal/operands/metrics/reconcile_test.go
```diff
@@ -54,14 +56,51 @@ var _ = Describe("Metrics operand", func() {
...
+	It("should create metrics resources", func() {
 		_, err := operand.Reconcile(&request)
 		Expect(err).ToNot(HaveOccurred())
 		ExpectResourceExists(newPrometheusRule(namespace), request)
 		ExpectResourceExists(newServiceMonitorCR(namespace), request)
 		ExpectResourceExists(newMonitoringClusterRole(), request)
 		ExpectResourceExists(newMonitoringClusterRoleBinding(), request)
 	})
+
+	It("should use the default runbook URL template when no ENV Variable is set", func() {
```
0xFelix, line 74: You might be able to deduce both tests into one `DescribeTable`. The only difference between both tests is whether the env variable was set or not.

internal/operands/metrics/resources.go
```diff
@@ -268,9 +277,22 @@ func newPrometheusRule(namespace string) *promv1.PrometheusRule {
...
 }
+
+func getRunbookURLTemplate() (string, error) {
+	runbookURLTemplate, exists := os.LookupEnv(runbookURLTemplateEnv)
+	if !exists {
+		runbookURLTemplate = defaultRunbookURLTemplate
+	}
+
+	if strings.Count(runbookURLTemplate, "%s") != 1 || strings.Count(runbookURLTemplate, "%") != 1 {
+		return "", fmt.Errorf("runbook URL template must have exactly 1 %%s substring")
```
0xFelix, line 294: Could use `errors.New` instead of `fmt.Errorf`.

internal/operands/metrics/resources.go
```diff
@@ -73,100 +78,104 @@ var RecordRulesDescList = []RecordRulesDesc{
...
-			"summary":     "The ssp-operator pod is up but failing to reconcile",
-			"runbook_url": runbookURLBasePath + "SSPFailingToReconcile",
-		},
-		Labels: map[string]string{
-			severityAlertLabelKey:     "critical",
-			healthImpactAlertLabelKey: "critical",
-			partOfAlertLabelKey:       partOfAlertLabelValue,
-			componentAlertLabelKey:    componentAlertLabelValue,
+func getAlertRules() []promv1.Rule {
+	runbookURLTemplate, _ := getRunbookURLTemplate()
```
0xFelix, line 82: When returning an error of course you should also check it and possibly pass it up in the call stack. Right now the possible error from `getRunbookURLTemplate` is just ignored.

internal/operands/metrics/reconcile_test.go
assafad: Done. Also added a case where an invalid value is set for the env variable.

internal/operands/metrics/resources.go
assafad: Updated, please let me know if the change makes sense

internal/operands/metrics/reconcile_test.go
```diff
@@ -54,14 +57,54 @@ var _ = Describe("Metrics operand", func() {
...
+						if rule.Annotations["runbook_url"] != "" {
+							Expect(rule.Annotations["runbook_url"]).To(Equal(fmt.Sprintf(template, rule.Alert)))
+						}
+					}
+				}
+			}
+		},
+		Entry("should use the default template when no ENV variable is set", defaultRunbookURLTemplate),
+		Entry("should use the ENV variable when a valid value is set", "valid/runbookURL/template/%s"),
+		Entry("should throw an error when an invalid ENV variable value is set", "invalid/runbookURL/template/"),
```
0xFelix, line 106: Add negative test case for another format specifier, e.g. `%d`?

0xFelix (commented):
Just one more nit, otherwise looks good to me!

internal/operands/metrics/reconcile_test.go
assafad: Added

akrejcir:
Looks good, thanks.

/lgtm

PR #526, opened 2023-03-16
refactor + fix: Cleanup proxy bundle

akrejcir:
- Cleaned up code in `vm-console-proxy-bundle` module
- Added validation that the bundle contains needed all resources

**Release note**:
```release-note
None
```


internal/vm-console-proxy-bundle/bundle-test-incorrect.yaml
```diff
@@ -0,0 +1,130 @@
+apiVersion: v1
```
codingben, line 1: What about to add here a comment to say what it's missing (incorrect)? E.g. `Incorrect because ConfigMap is missing`.

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -27,117 +28,100 @@ var (
...
 	}
-
-	return bundle, nil
 }
 
-func getObject(obj map[string]interface{}, newObj interface{}) error {
-	o, err := yamlv2.Marshal(&obj)
-	if err != nil {
-		return err
+func validateBundle(bundle *Bundle) error {
```
codingben, line 103: I think `validateBundle` it's too general naming and `validateBundleObjects` would be better.

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -27,117 +28,100 @@ var (
...
-					if err != nil {
-						return nil, err
-					}
-					bundle.ConfigMap = configMap
-				default:
-					continue
-				}
-			}
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), destObj)
+		if err != nil {
```
codingben, line 97: What about make it shorter:
```
if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), destObj); err != nil {
	return nil, err
}
```

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -27,117 +28,100 @@ var (
...
-	if err != nil {
-		return nil, err
-	}
+func loadBundleFromBytes(data []byte) (*Bundle, error) {
+	bundle := &Bundle{}
+	decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(data), 1024)
+	for {
+		obj := &unstructured.Unstructured{}
+		err := decoder.Decode(&obj)
+		if err == io.EOF {
```
codingben, line 66: If you like it:
```
if err := decoder.Decode(&obj); err == io.EOF {
	return bundle, nil
}
```

internal/vm-console-proxy-bundle/bundle.go
akrejcir: I disagree. It validates that the `Bundle` struct is correct and can be used.

internal/vm-console-proxy-bundle/bundle-test-incorrect.yaml
akrejcir: done

internal/vm-console-proxy-bundle/bundle.go
akrejcir: The line is already quite long. I prefer it this way.

internal/vm-console-proxy-bundle/bundle.go
akrejcir: The `err` is checked multiple times, and this change would limit its scope only to the inside of the `if` statement.

internal/vm-console-proxy-bundle/bundle.go
codingben: Yes, but it didn't mention what kind of validation it does. Maybe `validateBundleParameters`.

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -36,109 +36,58 @@ type Bundle struct {
...
-					err = getObject(obj, &configMap)
-					if err != nil {
-						return nil, err
-					}
-					bundle.ConfigMap = configMap
-				default:
-					return nil, fmt.Errorf("unsupported Kind found in vm-console-proxy bundle: %s", kind)
-				}
-			}
+		kind := obj.GetKind()
```
0xFelix, line 73: Merge these lines into the switch?

0xFelix: Also, could an empty Kind be a decoding error?

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -42,7 +42,15 @@ func ReadBundle(path string) (*Bundle, error) {
 		return nil, err
 	}
 
-	return loadBundleFromBytes(data)
+	bundle, err := loadBundleFromBytes(data)
+	if err != nil {
+		return nil, err
+	}
+	err = validateBundle(bundle)
```
0xFelix, line 49: Inline that with error check below?

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -36,109 +36,58 @@ type Bundle struct {
...
+		kind := obj.GetKind()
+		if kind == "" {
+			continue
 		}
-	}
 
-	return bundle, nil
-}
+		var destObj any
+		switch kind {
```
0xFelix, line 79: How do you ensure that there are no duplicated resources in the bundle?

0xFelix (commented):
Is it really necessary to "half-hardcode" the expected bundle resources into the operand and then to load their values from the bundle YAML? Can't we either define the resources completely in code or load them just from the YAML?

0xFelix:
Also, the commit messages are a bit sparse. Could you give some more context, please?

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -91,3 +99,29 @@ func loadBundleFromBytes(data []byte) (*Bundle, error) {
 		}
 	}
 }
+
+func validateBundle(bundle *Bundle) error {
+	missingFields := make([]string, 0, 6)
```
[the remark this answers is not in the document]
akrejcir: We gain a little bit of performance for free.

internal/vm-console-proxy-bundle/bundle.go
akrejcir: done

internal/vm-console-proxy-bundle/bundle.go
akrejcir: Done

akrejcir:
> Is it really necessary to "half-hardcode" the expected bundle resources into the operand and then to load their values from the bundle YAML? Can't we either define the resources completely in code or load them just from the YAML?

We cannot define the resources in ssp operator code, because they can change when a new version of `vm-console-proxy` is released. Loading them only from yaml would be good, but the reconciliation code needs to know the types of resources, so it can use the correct update functions.

I think we can think of a more generic solution as part of the SSP cleanup epic.

codingben (approved):
I'd maybe move `bundle-test.yaml` and `bundle-test-incorrect.yaml` to `test-yamls` folder or something. It's up to you.

/lgtm

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -27,117 +28,103 @@ var (
...
+		if err != nil {
+			return nil, err
+		}
 
-	return [][]byte{file}, nil
-}
+		kind := obj.GetKind()
+
+		var destObj any
+		switch obj.GetKind() {
```
0xFelix, line 76: `switch kind := obj.GetKind(); kind`?

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -27,117 +28,103 @@ var (
...
-				break
-			}
-			if err != nil {
-				return nil, err
-			}
-			if kind, ok := obj["kind"].(string); ok {
-				if kind == "" {
-					continue
-				}
+		if *(destObj.(*any)) != nil {
```
0xFelix, line 95: Add test cases for duplicated objects?

0xFelix (commented):
Added some comments.

> We cannot define the resources in ssp operator code, because they can change when a new version of vm-console-proxy is released. Loading them only from yaml would be good, but the reconciliation code needs to know the types of resources, so it can use the correct update functions.

TTO can load resources from the pipeline / task manifests without knowing exactly which resources it will encounter. It only has a list of known kinds it can handle / reconcile. Can't you do the same?

internal/vm-console-proxy-bundle/bundle.go
0xFelix: Or if you need the `kind` var later, use it here too?

ksimon1: https://github.com/kubevirt/tekton-tasks-operator/blob/main/pkg/tekton-bundle/tekton-bundle.go#L150

internal/vm-console-proxy-bundle/bundle.go
akrejcir: Done

internal/vm-console-proxy-bundle/bundle.go
akrejcir: @0xFelix Done

@ksimon1 That code was used originally here, and I just refactored it to make it smaller.

0xFelix (approved):
/lgtm

PR #527, opened 2023-03-21
fix: console proxy and tests: Stabilize functional tests

akrejcir:
This PR contains 2 changes to make functional tests less flaky:

Functions `RevertToOriginalSspCr()` now wait until SSP is deployed before returning.
This is useful, because reverting SSP may change TLS policy, which causes the pod to crash, and it may be some time before it is running again. If in the meantime, another test tries to update the SSP, it will fail, because the webhook cannot be reached.

When the annotation `ssp.kubevirt.io/vm-console-proxy-enabled` is removed from SSP CR, the operator waits for resources to be removed before setting the `status.phase` to `Deployed`.

**Release note**:
```release-note
None
```


lyarwood:
Sounds valid at first glance, just waiting on CI.

akrejcir:
I need to fix another race.

/hold

PR #532, opened 2023-03-28
feat: add tekton-tasks-operator

codingben:
**What this PR does / why we need it**:

Add kubevirt/tekton-tasks-operator here to have fewer operators and reduce maintenance efforts. 

**Release note**:

```release-note
Add tekton-tasks-operator
```

PR #535, opened 2023-04-03
fix: stabilize functests of VM Console proxy

akrejcir:
This change fixes a race, where the tests did not wait until console proxy resources were deployed.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/vm_console_proxy_test.go
+++ b/tests/vm_console_proxy_test.go
@@ -37,7 +37,7 @@ var _ = Describe("VM Console Proxy Operand", func() {
 		routeResource              testResource
 	)
 
-	BeforeEach(func() {
+	BeforeEach(OncePerOrdered, func() {
 		strategy.SkipSspUpdateTestsIfNeeded()
 
 		updateSsp(func(foundSsp *ssp.SSP) {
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
To be honest I would much prefer if the vm console proxy would be enabled by a feature gate as part of the SSP CR instead of an annotation like right now. If it was part of the SSP CR we could properly wait for SSP to be ready instead of introducing workarounds like these.

Can you move the feature gate into the SSP CR?

lyarwood:
> This PR will be backported to `release-v0.17` branch to fix failing tests in 4.13. Here is the bug: https://issues.redhat.com/browse/CNV-27768
> 
> Then in another PR, we will add the feature gate to `.spec`, which makes more sense.

/approve
/hold

This sounds like a good approach to me, @0xFelix can you `/unhold` if you're also okay with this given you asked the question initially?

ksimon1:
@akrejcir @codingben will add the feature gate in this PR https://github.com/kubevirt/ssp-operator/pull/532/files#diff-4f0e94347ff5dab9597e34bebabc082fc269a3311ec24f5038ec7cdb7849737fR99. Please use the same naming (or agree with Ben on different one)

0xFelix:
Cool! Sounds like a plan to me. Thank you.

/unhold

PR #538, opened 2023-04-06
fix: correctly delete proxy resources if namespace is not "kubevirt"

akrejcir:
`Cleanup()` method of vm-console-proxy deletes resources from the correct namespace.

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/operands/vm-console-proxy/reconcile.go
+++ b/internal/operands/vm-console-proxy/reconcile.go
@@ -1,6 +1,7 @@
 package vm_console_proxy
 
 import (
+	"context"
 	"fmt"
 	"strconv"
 
@@ -23,6 +24,8 @@ const (
 
 	operandName      = "vm-console-proxy"
 	operandComponent = "vm-console-proxy"
+
+	routeName = "vm-console-proxy"
 )
 
 // Define RBAC rules needed by this operand:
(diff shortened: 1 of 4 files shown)


akrejcir:
The `e2e-single-node-functests` failed when gathering logs after tests. The actual tests were successful:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/538/pull-ci-kubevirt-ssp-operator-master-e2e-single-node-functests/1648326650770755584

codingben (approved):
/lgtm

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -201,6 +200,100 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+			ExpectResourceExists(serviceAccount, request)
+			ExpectResourceExists(bundle.ClusterRole, request)
+			ExpectResourceExists(bundle.ClusterRoleBinding, request)
+			ExpectResourceExists(configMap, request)
+			ExpectResourceExists(service, request)
+			ExpectResourceExists(deployment, request)
+			ExpectResourceExists(route, request)
+
+			delete(request.Instance.Annotations, EnableAnnotation)
+			delete(request.Instance.Annotations, VmConsoleProxyNamespaceAnnotation)
```
codingben, line 283: What is the reason of deleting `VmConsoleProxyNamespaceAnnotation` if it's not going to be used in `Reconcile`?

Then without deleting it, it makes sense to write it as: `should remove resources from the namespace once EnableAnnotation deleted`.

akrejcir: Both annotations are for the proxy functionality, so I just deleted both.
We can improve it later, this PR is to unblock failing d/s tests.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -271,10 +331,43 @@ func getVmConsoleProxyImage() string {
 	return common.EnvOrDefault("VM_CONSOLE_PROXY_IMAGE", "quay.io/kubevirt/vm-console-proxy:v0.1.0")
 }
 
+func findResourcesUsingLabels[PtrL interface {
+	*L
+	client.ObjectList
+}, PtrT interface {
+	*T
+	client.Object
+}, L any, T any](ctx context.Context, name string, cli client.Client, itemsFunc func(list PtrL) []T) ([]client.Object, error) {
```
lyarwood, line 340: Would love to see this simplified with better variable names if possible but that's something for a follow up as this works as expected at the moment.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -271,10 +331,43 @@ func getVmConsoleProxyImage() string {
...
+			common.AppKubernetesComponentLabel: operandComponent,
+			common.AppKubernetesManagedByLabel: common.AppKubernetesManagedByLabelValue,
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error listing objects: %w", err)
+	}
+
+	// Filtering in a loop instead of using a FieldSelector in the List() call.
+	// It is only slightly inefficient, because all objects are already cached locally, so there is no API call.
```
lyarwood, line 354: Again something for a follow up but doesn't this duplicate what the client would be doing when using FieldSelector anyway? I didn't think it would fire off additional API calls to filter by fields.

lyarwood (commented):
/approve

This looks okay to unblock testing but I'd like to see my two comments discussed and possibly resolved in follow ups if you have time.

internal/operands/vm-console-proxy/reconcile.go
akrejcir: Correct, but it was easier to implement here because of the internals of the client.

We use the cached client from the `controller-runtime` library. It watches and stores objects locally. 
If we want to use a field selector for a specific type, we first need to add an index for the field name into the cache. These indexes are separate for every type of object.

I think the code would look like this:
```go
const field = "metadata.name"
var indexerFunc client.IndexerFunc = func(object client.Object) []string {
	return []string{object.GetName()}
}

mgr.GetFieldIndexer().IndexField(ctx, &v1.ServiceAccount{}, field, indexerFunc)
mgr.GetFieldIndexer().IndexField(ctx, &v1.ConfigMap{}, field, indexerFunc)
mgr.GetFieldIndexer().IndexField(ctx, &v1.Service{}, field, indexerFunc)
mgr.GetFieldIndexer().IndexField(ctx, &appsv1.Deployment{}, field, indexerFunc)
mgr.GetFieldIndexer().IndexField(ctx, &routev1.Route{}, field, indexerFunc)
```
I haven't tested it, so maybe it's incorrect.
I didn't like that we need to register the index for each type separately.

PR #544, opened 2023-05-02
refactor: add default update and status functions

akrejcir:
Added default update and status functions for most resources that are reconciled by the operator.

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/common/resource.go
+++ b/internal/common/resource.go
@@ -5,10 +5,16 @@ import (
 	"reflect"
 
 	"github.com/go-logr/logr"
+	routev1 "github.com/openshift/api/route/v1"
 	libhandler "github.com/operator-framework/operator-lib/handler"
+	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/prometheus/client_golang/prometheus"
+	apps "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	instancetypev1alpha2 "kubevirt.io/api/instancetype/v1alpha2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


internal/common/resource.go
```diff
@@ -431,3 +432,108 @@ func ResourceDeletedResult(resource client.Object, res OperationResult) Reconcil
...
+		foundConfigMap.Immutable = newTyped.Immutable
+		foundConfigMap.Data = newTyped.Data
+		foundConfigMap.BinaryData = newTyped.BinaryData
+
+	case *core.Namespace:
+		// Intentionally empty
+
+	case *core.Service:
+		foundService := foundObj.(*core.Service)
+		// ClusterIP should not be updated
```
0xFelix, line 449: What does this mean?

0xFelix (commented):
/approve

internal/common/resource.go
akrejcir: The `ClusterIP` field is set be the API server and is immutable. The problem is that `newObj` does not have the `ClusterIP` set, so if we would simply set the sepc:
```go
foundObj.spec = newObj.spec
```
it would set `ClusterIP` to `nil`, and the update would fail.

0xFelix: ACK

ksimon1:
lgtm
@codingben you will have to probably adjust again the PR with TTO migration

internal/common/resource.go
```diff
@@ -431,3 +432,108 @@ func ResourceDeletedResult(resource client.Object, res OperationResult) Reconcil
...
+		foundObj.(*routev1.Route).Spec = newTyped.Spec
+
+	case *promv1.PrometheusRule:
+		foundObj.(*promv1.PrometheusRule).Spec = newTyped.Spec
+
+	case *promv1.ServiceMonitor:
+		foundObj.(*promv1.ServiceMonitor).Spec = newTyped.Spec
+
+	default:
+		panic(fmt.Sprintf("Default update is not supported for type: %T", newObj))
```
codingben, line 494: We'll need to add also `pipeline.Task` and `pipeline.Pipeline` after your PR is merged.

codingben (approved):
/lgtm

PR #546, opened 2023-05-09
fix: Disable functest for runbook reachability

akrejcir:
The test fails in downstream, because the URL does not exist yet.

**Release note**:
```release-note
None
```


akrejcir:
Closing. We will disable this test only in downstream.

/close

PR #554, opened 2023-05-23
fix: fix setting namespace for example pipelines role binding

ksimon1:
fix: fix setting namespace for example pipelines role binding

one role binding for example pipeline has to be deployed to the kubevirt-os-images namespace. SSP operator will not set namespace for roleBindings if they have namespace defined.

**Release note**:
```
NONE
```


0xFelix:
/approve

I would very much appreciate if we could investigate if it is possible to remove the additional roles / SAs in the pipelines again.

PR #555, opened 2023-05-23
fix: add validator image variable to pipeline test scripts

ksimon1:
fix: add validator image variable to pipeline test scripts

Add validator img link to example pipelines test script. Without this link, validator uses latest tag, which is 3 years old and ssp operator is in deploying phase, because validator is not working

**Release note**:
```
NONE
```

### DIFF
--- a/automation/e2e-tekton/example-pipelines-test.sh
+++ b/automation/e2e-tekton/example-pipelines-test.sh
@@ -2,6 +2,7 @@
 
 cp -L $KUBECONFIG /tmp/kubeconfig && export KUBECONFIG=/tmp/kubeconfig
 export IMG=${CI_OPERATOR_IMG}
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
IMO it would make more sense to ensure the `latest` tag is updated again. Otherwise running `make deploy` without setting `VALIDATOR_IMG` is broken for developers too.

ksimon1:
> IMO it would make more sense to ensure the `latest` tag is updated again. Otherwise running `make deploy` without setting `VALIDATOR_IMG` is broken for developers too.

We have a story for deciding what to do with latest tag and in the near future someone will take care of it

akrejcir:
@0xFelix , this change is needed regardless of the `latest` tag, because we want to run CI on containers build from the current PR, not on the one published as `latest`.

PR #557, opened 2023-05-23
fix: fix wrong version of tekton tasks containers

ksimon1:
fix: fix wrong version of tekton tasks containers

fix wrong version of tekton tasks containers and update release script to update version file.

**Release note**:
```
NONE
```


0xFelix (approved):
/approve
/lgtm

PR #559, opened 2023-05-24
 Fix total_restored_common_templates metric update

machadovilaca:

SSPCommonTemplatesModificationReverted is being triggered during upgrades.

Fixes #

**Release note**:

```release-note
 Fix total_restored_common_templates metric update
```

### DIFF
--- a/internal/common/resource.go
+++ b/internal/common/resource.go
@@ -39,6 +39,7 @@ type ResourceStatus struct {
 
 type ReconcileResult struct {
 	Status          ResourceStatus
+	InitialResource client.Object
 	Resource        client.Object
 	OperationResult OperationResult
 }
@@ -192,7 +193,7 @@ func (r *reconcileBuilder) Reconcile() (ReconcileResult, error) {
 		return nil
 	}
 
-	res, err := r.createOrUpdateWithImmutableSpec(found, mutateFn)
+	res, existing, err := r.createOrUpdateWithImmutableSpec(found, mutateFn)
 	if err != nil {
 		r.request.Logger.Info(fmt.Sprintf("Resource create/update failed: %v", err))
 		return ReconcileResult{}, err
@@ -206,7 +207,7 @@ func (r *reconcileBuilder) Reconcile() (ReconcileResult, error) {
 	logOperation(res, found, r.request.Logger)
 
 	status := r.statusFunc(found)
-	return ReconcileResult{status, r.resource, res}, nil
+	return ReconcileResult{status, existing, r.resource, res}, nil
 }
 
 func CreateOrUpdate(request *Request) ReconcileBuilder {
@@ -285,43 +286,43 @@ func DeleteAll(request *Request, resources ...client.Object) ([]CleanupResult, e
 }
 
 // This function was initially copied from controllerutil.CreateOrUpdate
-func (r *reconcileBuilder) createOrUpdateWithImmutableSpec(obj client.Object, f controllerutil.MutateFn) (OperationResult, error) {
+func (r *reconcileBuilder) createOrUpdateWithImmutableSpec(obj client.Object, f controllerutil.MutateFn) (OperationResult, client.Object, error) {
 	key := client.ObjectKeyFromObject(obj)
 	if err := r.request.Client.Get(r.request.Context, key, obj); err != nil {
 		if !errors.IsNotFound(err) {
-			return OperationResultNone, err
+			return OperationResultNone, nil, err
 		}
 		if err := mutate(f, key, obj); err != nil {
-			return OperationResultNone, err
+			return OperationResultNone, nil, err
 		}
 		if err := r.request.Client.Create(r.request.Context, obj); err != nil {
-			return OperationResultNone, err
+			return OperationResultNone, nil, err
 		}
-		return OperationResultCreated, nil
+		return OperationResultCreated, nil, nil
 	}
 
-	existing := obj.DeepCopyObject()
+	existing := obj.DeepCopyObject().(client.Object)
 	if err := mutate(f, key, obj); err != nil {
-		return OperationResultNone, err
+		return OperationResultNone, existing, err
 	}
 
 	if equality.Semantic.DeepEqual(existing, obj) {
-		return OperationResultNone, nil
+		return OperationResultNone, existing, nil
 	}
 
 	// If the resource is immutable and specs are not equal, delete it.
 	// It will be recreated in the next iteration.
-	if r.immutableSpec && !equality.Semantic.DeepEqual(r.specGetter(existing.(client.Object)), r.specGetter(obj)) {
+	if r.immutableSpec && !equality.Semantic.DeepEqual(r.specGetter(existing), r.specGetter(obj)) {
 		if err := r.request.Client.Delete(r.request.Context, obj); err != nil {
-			return OperationResultNone, err
+			return OperationResultNone, existing, err
 		}
-		return OperationResultDeleted, nil
+		return OperationResultDeleted, existing, nil
 	}
 
 	if err := r.request.Client.Update(r.request.Context, obj); err != nil {
-		return OperationResultNone, err
+		return OperationResultNone, existing, err
 	}
-	return OperationResultUpdated, nil
+	return OperationResultUpdated, existing, nil
 }
 
 // This function is a copy of controllerutil.mutate
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -5,9 +5,7 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/go-logr/logr"
 	templatev1 "github.com/openshift/api/template/v1"
-	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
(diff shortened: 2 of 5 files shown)


tests/metrics_test.go
```diff
@@ -246,8 +196,9 @@ var _ = Describe("Metrics", func() {
 			unpauseSsp()
 			waitUntilDeployed()
 
-			newRestoredCount := totalRestoredTemplatesCount()
-			Expect(newRestoredCount).To(BeZero())
+			Consistently(func() int {
+				return totalRestoredTemplatesCount()
+			}, 5*time.Minute, 10*time.Second).Should(Equal(restoredCount))
```
akrejcir, line 201: Are 5 minutes really necessary? Tests already take a long time to run, and we try to minimize the time.

akrejcir:
Both commits change the same lines in file `tests/metrics_test.go`. Can you move changes in that file into a single commit?

0xFelix (commented):
Can you please add some context into the commit descriptions?

internal/common/resource.go
```diff
@@ -392,6 +398,31 @@ func logOperation(result OperationResult, resource client.Object, logger logr.Lo
 	}
 }
 
+func createOperationMetrics(req *Request, result OperationResult, expected client.Object, found client.Object, logger logr.Logger) {
```
akrejcir, line 401: Can you rename this function? It does not create anything.

internal/common/resource.go
```diff
@@ -392,6 +398,31 @@ func logOperation(result OperationResult, resource client.Object, logger logr.Lo
...
+
+	logger.Info(message)
+	TemplatesRestored.Inc()
+}
+
+func isVersionUpgrade(req *Request, result OperationResult, expected client.Object, found client.Object) bool {
+	isOperatorUpgrade := req.Instance.Status.ObservedVersion != GetOperatorVersion()
+
+	expectedVersion := expected.GetLabels()[AppKubernetesVersionLabel]
+	foundVersion := found.GetLabels()[AppKubernetesVersionLabel]
```
akrejcir, line 421: At this point in the code the labels on `expected` and `found` are guaranteed to be the same, because the labels on `found` are updated here on line 190.

tests/metrics_test.go
```diff
@@ -167,87 +158,48 @@ var _ = Describe("Metrics", func() {
...
-							rs.Status.ReadyReplicas == *rs.Spec.Replicas {
-							return nil
-						}
-					}
-				}
-
-				return fmt.Errorf("operator version not yet updated")
-			}, env.ShortTimeout(), time.Second).Should(Succeed())
-		}
+		var templates *templatev1.TemplateList
```
akrejcir, line 161: Only the first template is used in the tests. Can you change the type of this variable to `*templatev1.Template`?

tests/metrics_test.go
```diff
@@ -167,87 +158,48 @@ var _ = Describe("Metrics", func() {
...
-
-				for i, env := range envs {
-					if env.Name == common.OperatorVersionKey {
-						envs[i].Value = version
-						break
-					}
-				}
+		getTemplates := func() *templatev1.TemplateList {
+			templates := &templatev1.TemplateList{}
+			err := apiClient.List(ctx, templates, &client.ListOptions{Namespace: getSsp().Namespace})
```
akrejcir, line 165: Templates are not deployed in the same namespace as SSP CR. Please use `strategy.GetTemplatesNamespace()` to get the namespace. Also that namespace can contain templates that are not managed by SSP, so you need to check the annotations: `operator-sdk/primary-resource` and `operator-sdk/primary-resource-type`.

It would be simpler to use the code that was previously here, that used `createTestTemplate()` function.

akrejcir (commented):
Can you squash the commits? The fist commit is not correct without the second one.

akrejcir (commented):
Can you squash the commits? The fist commit is not correct without the second one.

internal/operands/common-templates/reconcile.go
```diff
@@ -239,6 +245,10 @@ func reconcileTemplatesFuncs(templatesBundle []templatev1.Template) []common.Rec
 					newTemplate := newRes.(*templatev1.Template)
 					foundTemplate := foundRes.(*templatev1.Template)
 
+					if foundTemplate.GetLabels()[common.AppKubernetesVersionLabel] == newTemplate.GetLabels()[common.AppKubernetesVersionLabel] {
```
akrejcir, line 248: At this point in the code these labels are guaranteed to be the same, so this condition is always `true`.
Look here: https://github.com/kubevirt/ssp-operator/blob/6cb14a15b4d1169fddd412193da0aa4ff58d83d0/internal/common/resource.go#L185-L191

The `UpdateLabels()` function is called before `r.updateFunc()`.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -258,6 +277,63 @@ var _ = Describe("Common-Templates operand", func() {
...
+			template.Namespace = namespace
+
+			desc, value := getCommonTemplatesRestoredMetric()
+			Expect(desc).To(ContainSubstring("total_restored_common_templates"))
+			initialMetricValue = value
+		})
+
+		It("should increase by 1 when one template is restored", func() {
+			cloneTemplate := template.DeepCopy()
+			escapedTemplateTypeLabel := strings.ReplaceAll(TemplateTypeLabel, "/", "~1")
```
akrejcir, line 299: Using `Patch()` is complex for a unit test. Can you instead call `Get()` followed by `Update()`?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -258,6 +277,63 @@ var _ = Describe("Common-Templates operand", func() {
...
+			Expect(updatedTpl.Labels[TemplateTypeLabel]).To(Equal(template.Labels[TemplateTypeLabel]))
+
+			desc, value := getCommonTemplatesRestoredMetric()
+			Expect(desc).To(ContainSubstring("total_restored_common_templates"))
+			Expect(value).To(Equal(initialMetricValue + 1))
+		})
+
+		It("should not increase when template restored is from previous version", func() {
+			cloneTemplate := template.DeepCopy()
+			escapedTemplateTypeLabel := strings.ReplaceAll(TemplateTypeLabel, "/", "~1")
```
akrejcir, line 317: The same comment as above.

internal/operands/common-templates/reconcile_test.go
```diff
@@ -285,3 +361,24 @@ func getTestTemplates() []templatev1.Template {
 		},
 	}}
 }
+
+func getCommonTemplatesRestoredMetric() (string, float64) {
+	ch := make(chan prometheus.Metric, 1)
+	CommonTemplatesRestored.Collect(ch)
```
akrejcir, line 367: Is `.Collect()` guaranteed to only send 1 metric to the channel? Otherwise it would block forever.

The `prometheus.Counter` interface also implements `Metric` interface. Could you use that directly, instead of calling `Collect()`?
```go
CommonTemplatesRestored.Write(dto)
```

internal/operands/common-templates/reconcile_test.go
```diff
@@ -285,3 +361,24 @@ func getTestTemplates() []templatev1.Template {
...
+	close(ch)
+	m := <-ch
+	dto := &io_prometheus_client.Metric{}
```
akrejcir, line 370: What does `dto` mean? Can you use longer variable names?

tests/metrics_test.go
```diff
@@ -167,87 +159,36 @@ var _ = Describe("Metrics", func() {
...
+			template.Labels[common_templates.TemplateTypeLabel] = "test"
+			template.Labels[common_templates.TemplateVersionLabel] = "v" + rand.String(5)
 			Expect(apiClient.Update(ctx, template)).To(Succeed())
 
-			unpauseSsp()
-			waitUntilDeployed()
-
-			newRestoredCount := totalRestoredTemplatesCount()
-			Expect(newRestoredCount).To(BeZero())
+			Consistently(func() int {
```
akrejcir, line 189: Using `Consistently()` is not ideal. There may be a way to wait for the template update directly, and not just for some arbitrary duration.

Can you add a `// TODO:` comment here, so that we don't forget about it?

akrejcir (commented):
Returning the `existing` resource from `createOrUpdateWithImmutableSpec()` seems hacky, but it's ok for this PR. We can improve the code later.

internal/operands/common-templates/reconcile_test.go
machadovilaca: updated

internal/operands/common-templates/reconcile_test.go
machadovilaca: updated

internal/operands/common-templates/reconcile_test.go
machadovilaca: `Collect` implementation is just `ch <- c.self`, so it is only one.

`Write` works fine, but the Prometheus registry uses `Collect` and not `Write` (which just encapsulates the metric), so I usually prefer to go with `Collect`

internal/operands/common-templates/reconcile_test.go
machadovilaca: it is a commonly used name by Prometheus in these cases, but we can use `metric` for clarity

tests/metrics_test.go
machadovilaca: added

0xFelix (commented):
/approve

internal/operands/common-templates/reconcile_test.go
akrejcir: ok

PR #560, opened 2023-05-31
fix(CNV-29287): consider tekton crd as an optional

codingben:
Currently if the Tekton Feature Gate is set to false, ssp-controller will  consider
Tekton CRD as required. But, since it's not needed, then consider it as not required.

**Release note**:

```release-note
Consider Tekton CRD as an optional.
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -51,6 +51,9 @@ import (
 	"kubevirt.io/ssp-operator/internal/controller/predicates"
 	crd_watch "kubevirt.io/ssp-operator/internal/crd-watch"
 	"kubevirt.io/ssp-operator/internal/operands"
+	tekton_pipelines "kubevirt.io/ssp-operator/internal/operands/tekton-pipelines"
+	tekton_tasks "kubevirt.io/ssp-operator/internal/operands/tekton-tasks"
+	tekton_bundle "kubevirt.io/ssp-operator/internal/tekton-bundle"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


controllers/ssp_controller.go
```diff
@@ -205,7 +205,14 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 	}
 
 	if r.areCrdsMissing {
-		err := updateStatusMissingCrds(sspRequest, r.crdList.MissingCrds())
+		var missingCrds []string
+		if !sspRequest.Instance.Spec.FeatureGates.DeployTektonTaskResources {
+			missingCrds = excludeTektonCrd(r.crdList.MissingCrds())
+		} else {
+			missingCrds = r.crdList.MissingCrds()
+		}
```
0xFelix, line 213:
```
missingCrds := r.crdList.MissingCrds()
if !sspRequest.Instance.Spec.FeatureGates.DeployTektonTaskResources {
	missingCrds = excludeTektonCrd(missingCrds)
}
```

?

codingben: I removed this code, as I didn't pay attention to the `setup.go` where there is also error log of missing crds.

codingben:
@akrejcir @ksimon1 @0xFelix As a quick fix, I removed `Crd: tektonCrd, WatchFullObject: true` from the `WatchClusterTypes()` .
This change will not consider Tekton CRD as required and won't throw an error (and the operands reconciliation protected by Feature Gate).

akrejcir:
CI is not relevant for this PR, because it installs Tekton CRDs before running the tests.
Holding until it's verified locally on a cluster without Tekton.

/hold

akrejcir:
> @akrejcir @ksimon1 @0xFelix As a quick fix, I removed `Crd: tektonCrd, WatchFullObject: true` from the `WatchClusterTypes()` . This change will not consider Tekton CRD as required and won't throw an error (and the operands reconciliation protected by Feature Gate).

It will probably still panic here:
https://github.com/kubevirt/ssp-operator/blob/21811a6de3a627bb0d60995c936f2c419999cab4/controllers/ssp_controller.go#L123-L126
Because tekton operands will try to add tekton resources to watch, Which will panic, because CRD does not exist.

PR #561, opened 2023-06-01
fix(CNV-29287): do not watch tekton crd

codingben:

Currently, we're watching to Tekton CRD which
means it's required CRD. This is not always the
case if the Feature Gate is disabled.

This is a quick workaround to the required
Tekton CRD even when Feature Gate is disabled.

A solution to that issue is tracked in #560.

**Release note**:

```release-note
Do not watch Tekton CRD.
```

### DIFF
--- a/tests/tekton-pipelines_test.go
+++ b/tests/tekton-pipelines_test.go
@@ -5,7 +5,8 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	pipeline "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+
+	// pipeline "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
 	ssp "kubevirt.io/ssp-operator/api/v1beta1"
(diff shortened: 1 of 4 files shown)


codingben:
@dominikholler @ksimon1 @akrejcir This is a quick workaround to HCO CI, so we can meet the FF date. Until then, an effort to that solution is tracked in #560.

HCO CI currently is failing because ssp-operator throws an error about missing Tekton CRD, even when the Feature Gate is disabled. It blocks the PR in HCO, https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2349.

akrejcir:
Functional tests that check if tekton resources are reconciled need to be disabled, because they will fail after this PR.

codingben:
> Functional tests that check if tekton resources are reconciled need to be disabled, because they will fail after this PR.

@akrejcir Ah okay, I disabled them in this PR. Let's see if it works now.

tests/tekton-pipelines_test.go
```diff
@@ -33,25 +34,25 @@ var _ = Describe("Tekton Pipelines Operand", func() {
...
-				Expect(err).ToNot(HaveOccurred())
-				return len(pipelineList.Items) > 0
-			}, env.ShortTimeout(), time.Second).Should(BeTrue())
-
-			for _, pipeline := range pipelineList.Items {
-				Expect(pipeline.Labels[common.AppKubernetesManagedByLabel]).To(Equal(common.AppKubernetesManagedByValue), "managed by label should equal")
-				Expect(pipeline.Labels[common.AppKubernetesComponentLabel]).To(Equal(string(common.AppComponentTektonPipelines)), "component label should equal")
-			}
-		})
+		// It("[test_id:TODO] should create pipelines", func() {
```
akrejcir, line 37: In the future, you can disable tests by setting them to "pending", using `PIt()` instead of `It()`. Then we don't need to comment out code.

PR #562, opened 2023-06-02
Remove node-labeller

ksimon1:
feat: remove node labeller and bump api to v1beta2

Original credit goes to https://github.com/kubevirt/ssp-operator/pull/492 and @lyarwood 

**Release note**:
```
feat: remove node labeller and bump api to v1beta2
```


0xFelix (commented):
As this is right now it will break existing installations. Both API versions need to be served in parallel for existing installations to be able to upgrade. And you should ensure that the `None` conversion strategy is dropping unknown API fields. (see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning)

0xFelix:
Can you also swap the order of commits, i.e. first introduce `v1beta2`, then change its schema?

ksimon1:
@akrejcir, @0xFelix, @lyarwood can you please review this?

api/v1beta1/ssp_types.go
```diff
@@ -146,7 +146,7 @@ type SSPStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-
+// +kubebuilder:deprecatedversion:warning="ssp.kubevirt.io/v1beta1 ssp is deprecated"
```
0xFelix, line 149: Is it necesarry to add this explicitly? IIRC this is also the default?

config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -3012,6 +3014,2038 @@ spec:
             type: object
         type: object
     served: true
+    storage: false
```
0xFelix, line 3017: Keep the storage version at `v1beta1` until we have a proper migration/conversion strategy?

api/v1beta1/ssp_types.go
ksimon1: Yes, we have to add it, otherwise generator will not add `deprecate: true` attribute.

config/crd/bases/ssp.kubevirt.io_ssps.yaml
ksimon1: AFAIU we don't need the conversion strategy, because we don't need to migrate any values

webhooks/ssp_webhook.go
```diff
@@ -32,19 +32,19 @@ import (
...
 func Setup(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
-		For(&sspv1beta1.SSP{}).
+		For(&ssp.SSP{}).
 		WithValidator(newSspValidator(mgr.GetClient())).
 		Complete()
 }
 
-// +kubebuilder:webhook:verbs=create;update,path=/validate-ssp-kubevirt-io-v1beta1-ssp,mutating=false,failurePolicy=fail,groups=ssp.kubevirt.io,resources=ssps,versions=v1beta1,name=validation.ssp.kubevirt.io,admissionReviewVersions=v1,sideEffects=None
+// +kubebuilder:webhook:verbs=create;update,path=/validate-ssp-kubevirt-io-v1beta2-ssp,mutating=false,failurePolicy=fail,groups=ssp.kubevirt.io,resources=ssps,versions=v1beta2,name=validation.ssp.kubevirt.io,admissionReviewVersions=v1,sideEffects=None
```
akrejcir, line 47: The webhook should still be called for `v1beta1` version. Can you include `versions=v1beta1;v1beta2` in the bubebuilder comment, and regenerate manifests?

config/crd/bases/ssp.kubevirt.io_ssps.yaml
ksimon1: Taken from: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects 
```
if the default strategy value None is specified, the only modifications to the object are changing the apiVersion string and perhaps pruning unknown fields (depending on the configuration).
```
Which is exactly our use case, we just need to prune NL fields

lyarwood: Ah ha! I had forgotten the second part apologies, in that case `None` is fine.

api/v1beta1/ssp_types.go
0xFelix: ACK

lyarwood:
@ksimon1 Would it be worthwhile adding `NodeLabeller` to the initial upgrade func test CR?

https://github.com/kubevirt/ssp-operator/blob/b8798993fc8932d152c07bb2009dabb16272150b/automation/e2e-upgrade-functests/ssp-cr-template.yaml

lyarwood:
> @ksimon1 Would it be worthwhile adding `NodeLabeller` to the initial upgrade func test CR?
> 
> https://github.com/kubevirt/ssp-operator/blob/b8798993fc8932d152c07bb2009dabb16272150b/automation/e2e-upgrade-functests/ssp-cr-template.yaml

AFAICT we would also need to rework the test to upgrade/deploy the operator itself from the local source tree for such a test to be valid:

https://github.com/kubevirt/ssp-operator/blob/b8798993fc8932d152c07bb2009dabb16272150b/automation/e2e-upgrade-functests/run.sh

lyarwood:
/lgtm

Thanks!

lyarwood:
@ksimon1  Any reason to hold this anymore?

ksimon1:
/hold cancel
If you have approve permissions, you can approve any PR :)

PR #563, opened 2023-06-02
chore: remove old common-template manifests

ksimon1:
chore: remove old common-template manifests
this change will save few MB in container

**Release note**:
```
removed old common-templates manifests
```


0xFelix (approved):
/approve
/lgtm

PR #566, opened 2023-06-07
fix: add virtualMachine.kubevirt.io crd to required crds

ksimon1:
fix: add virtualMachine.kubevirt.io crd to required crds

vm-controller is not checking if virtualMachine.kubevirt.io crd exists. this is causing troubles, when kubevirt crds are uninstalled when ssp oeprator
is running.

/hold

**Release note**:
```
fix: add virtualMachine.kubevirt.io crd to required crds
```

### DIFF
--- a/controllers/setup.go
+++ b/controllers/setup.go
@@ -99,7 +99,8 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 		)
 	}
 
-	var requiredCrds []string
+	requiredCrds := []string{getVMControllerRequiredCRD()}
+
 	for i := range sspOperands {
 		requiredCrds = append(requiredCrds, getRequiredCrds(sspOperands[i])...)
 	}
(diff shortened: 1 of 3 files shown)


controllers/setup.go
```diff
@@ -99,7 +99,10 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 		)
 	}
 
-	var requiredCrds []string
+	requiredCrds := []string{
+		"virtualMachine.kubevirt.io",
```
0xFelix, line 103: Can you put this into the vm_controller instead of defining it just here? e.g. adding a `WatchTypes` func to vm_controller?

codingben: @0xFelix Isn't that function related only to operands? Is `vm_controller` considered as an operand?

0xFelix: My point is that this is a requirement of `vm_controller.go` but when defined like this in `setup.go` it is not obvious. The suggestion was to add something similar to a `WatchTypes()` func to `vm_controller.go`.

akrejcir: +1 for Felix's idea. It will make it clear that `virtualMachine.kubevirt.io` is needed by the `vm_controller`.

If the CRD is missing, the controller should not be added to the manager. You can add an `if` condition round lines  143-150

PR #581, opened 2023-06-19
fix: Keep API v1beta1 in scheme.

akrejcir:
This allows the operator to decode objects from the old API version.

Changed webhook to accept both versions.

**Release note**:
```release-note
None
```


config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -19,8 +19,6 @@ spec:
     - jsonPath: .status.phase
       name: Phase
       type: string
-    deprecated: true
-    deprecationWarning: ssp.kubevirt.io/v1beta1 ssp is deprecated
     name: v1beta1
```
lyarwood, line 22: Why are we dropping the deprecation?

lyarwood (changes requested):

lyarwood:
So are the upgrade tests just deploying `v0.18.0` and updating to HEAD?

Shouldn't an upgrade test deploy `v0.17.0` and upgrade to HEAD? 

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kubevirt_ssp-operator/581/pull-ci-kubevirt-ssp-operator-master-e2e-upgrade-functests/1672166756140978176

```
Warning: ssp.kubevirt.io/v1beta1 ssp is deprecated
Error from server: error when creating "STDIN": admission webhook "validation.ssp.kubevirt.io" denied the request: no kind "SSP" is registered for version "ssp.kubevirt.io/v1beta1" in scheme "/workspace/internal/common/scheme.go:19"
{"component":"entrypoint","error":"wrapped process failed: exit status 1","file":"k8s.io/test-infra/prow/entrypoint/run.go:79","func":"k8s.io/test-infra/prow/entrypoint.Options.Run","level":"error","msg":"Error executing test process","severity":"error","time":"2023-06-23T09:46:01Z"}
```

config/crd/bases/ssp.kubevirt.io_ssps.yaml
akrejcir: Because I kept the storage version `v1beta1`. We can deprecate it in a later PR.

akrejcir:
> So are the upgrade tests just deploying `v0.18.0` and updating to HEAD?

Correct. Upgrade tests check that it's possible to upgrade from the latest released version to the changes in the current PR.

> Shouldn't an upgrade test deploy `v0.17.0` and upgrade to HEAD?

Before `v0.18.0` was released the update tests were using `v0.17.x`, so the test for upgrade from `v0.17.x` to `v0.18.0` were done on the latest PR before `v0.18.0` release.

Do you think we should add another CI lane for update from the second last released version to HEAD?

akrejcir:
For this PR, I think we can ignore upgrade tests, because `v0.18.0` is broken.

config/crd/bases/ssp.kubevirt.io_ssps.yaml
lyarwood: What's the reasoning behind keeping the storage version `v1beta1`? Moving it to `v1beta2` shouldn't have any impact for us as https://github.com/openshift/cluster-kube-storage-version-migrator-operator should always be present in our environment?

lyarwood:
> > So are the upgrade tests just deploying `v0.18.0` and updating to HEAD?
> 
> Correct. Upgrade tests check that it's possible to upgrade from the latest released version to the changes in the current PR.

Right in my opinion that isn't an upgrade but an update as you're not moving between `minor` versions. It's a valid test but I wouldn't call it an upgrade.

> > Shouldn't an upgrade test deploy `v0.17.0` and upgrade to HEAD?
> 
> Before `v0.18.0` was released the update tests were using `v0.17.x`, so the tests for upgrade from `v0.17.x` to `v0.18.0` were done on the latest PR before `v0.18.0` release.

I guess we didn't catch the bug because the above update test of `v0.18.0` to HEAD didn't run causing us to submit a `v1beta1` CR to the webhooks?

> Do you think we should add another CI lane for update from the second last released version to HEAD?

Yes, the latest `release-0.17` release to HEAD would be an upgade IMHO and useful coverage to always have.

akrejcir:
> Right in my opinion that isn't an upgrade but an update as you're not moving between `minor` versions. It's a valid test but I wouldn't call it an upgrade.
>

Good to know. I didn't know there was a different meaning between the words in software context.

>> ...

> I guess we didn't catch the bug because the above update test of `v0.18.0` to HEAD didn't run causing us to submit a `v1beta1` CR to the webhooks?

The test did run, but there was no test case to catch this bug.


> > Do you think we should add another CI lane for update from the second last released version to HEAD?
> 
> Yes, the latest `release-0.17` release to HEAD would be an upgade IMHO and useful coverage to always have.

Ok, I will add an Issue to track it.

config/crd/bases/ssp.kubevirt.io_ssps.yaml
akrejcir: Ok, I didn't know about it. I will change it back.

lyarwood:
/approve
/lgtm

Assuming CI is happy this looks correct to me now.

lyarwood:
/cherry-pick release-0.18

@akrejcir btw I also want to land https://github.com/kubevirt/ssp-operator/pull/584 in `release-0.18` before you cut `v0.18.1`, not sure if you want to track that with a milestone?

lyarwood:
Can you bump the version of the CR used in the current upgrade test to `v1beta2` to work around this bug? Once this PR lands and is tagged in a release we can move this back to `v1beta1`:

https://github.com/kubevirt/ssp-operator/blob/b75c5bf101791ae50aaea91408e2e7558a8fadd1/config/samples/ssp_v1beta1_ssp.yaml#L1

akrejcir:
Is it ok, if I just override the test?

lyarwood:
> Is it ok, if I just override the test?

We can but it's going to fail everywhere until this gets released. I don't mind but I'd appreciate it if you could post a PR after that just showing the job passing again?

PR #591, opened 2023-06-26
[release-v0.18] Fix env variables

ksimon1:
fix: add VIRTIO_IMG env variable to manager manifest
remove VIRT_LAUNCHER env variable, which is not used anywhere.

VIRTIO_IMG is needed for tekton pipelines

PR to main will follow after this PR
/cc @akrejcir 
**Release note**:
```
NONE

```


akrejcir:
I see the the `VIRTIO_IMG` is used by the newly added tekton-pipelines operand.

/lgtm

PR #592, opened 2023-06-26
fix(CNV-30284): rename tekton images env vars

codingben:

Rename Tekton environment variables to be the
same as declared in CSV. Currently, they are not
the same as in CSV and no value is set.

**Release note**:

```release-note
Fix Tekton images enviroment variables.
```


0xFelix (approved):
/approve

PR #593, opened 2023-06-26
common-instancetypes: Move to `instancetype.kubevirt.io/v1beta1`

lyarwood:
**What this PR does / why we need it**:

`kubevirt.io/api` `v1.0.0-rc.0` is also pulled in but the underlying dependency on `k8s.io/api` is pinned to `v0.26.2` for now as moving to `v0.27.3` requires a large amount of work in SSP operator thanks to changes to [`controller-runtime` `v0.15.0`](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0), see https://github.com/kubevirt/ssp-operator/pull/598 for more on that work.

**Which issue(s) this PR fixes**: 

Fixes #590

**Special notes for your reviewer**:

**Release note**:

```release-note
 The `instancetype.kubevirt.io/v1beta1` API is now used by the `common-instancetypes` operand
```

PR #595, opened 2023-06-26
Don't override autoupdated datasources that point to a snapshot instead of PVC

akalenyu:
With CDI 1.57.0 we have a new [feature](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/os-image-poll-and-update.md#dataimportcron-source-formats) that introduces datasources that point to snapshot objects instead of PVCs.
We want to not override those, since this is the correct source to be pointing at.

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=2215285

**Release note**:

```release-note
BugFix: SSP resets datasource reference to initial state
```

### DIFF
--- a/internal/operands/data-sources/reconcile.go
+++ b/internal/operands/data-sources/reconcile.go
@@ -396,7 +396,7 @@ func reconcileDataSource(dsInfo dataSourceInfo, request *common.Request) (common
 
 			foundDs := foundRes.(*cdiv1beta1.DataSource)
 			newDs := newRes.(*cdiv1beta1.DataSource)
-			if !dsInfo.autoUpdateEnabled || foundDs.Spec.Source.PVC == nil {
+			if !dsInfo.autoUpdateEnabled || (foundDs.Spec.Source.PVC == nil && foundDs.Spec.Source.Snapshot == nil) {
 				foundDs.Spec.Source.PVC = newDs.Spec.Source.PVC
 			}
 		}).
--- a/Makefile
+++ b/Makefile
@@ -167,8 +167,8 @@ vet:
 # Update vendor modules
 .PHONY: vendor
 vendor:
-	go mod vendor
 	go mod tidy
+	go mod vendor
 
 # Validate that this repository does not contain offensive language
 .PHONY: validate-no-offensive-lang
--- a/hack/kubevirtci.sh
+++ b/hack/kubevirtci.sh
@@ -38,7 +38,8 @@ function kubevirtci::up() {
   KUBECONFIG=$(kubevirtci::kubeconfig)
   export KUBECONFIG
   echo "adding kubevirtci registry to cdi-insecure-registries"
-  ${_kubectl} patch configmap cdi-insecure-registries -n cdi --type merge -p '{"data":{"kubevirtci": "registry:5000"}}'
+  ${_kubectl} get cdis --output='name' --ignore-not-found \
+  | xargs -r ${_kubectl} patch --type merge -p '{"spec": {"config": {"insecureRegistries": ["registry:5000"]}}}'
   echo "installing kubevirt..."
   LATEST=$(curl -L https://storage.googleapis.com/kubevirt-prow/devel/release/kubevirt/kubevirt/stable.txt)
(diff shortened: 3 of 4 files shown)


config/manager/kustomization.yaml
```diff
@@ -18,5 +18,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/kubevirt/ssp-operator
-  newTag: latest
+  newName: registry:5000/kubevirt/ssp-operator
```
ksimon1, line 21: I would prefer to not commit these changes

akalenyu: this got changed with `make vendor generate manifests`

go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	k8s.io/kube-aggregator v0.26.2
 	k8s.io/utils v0.0.0-20230202215443-34013725500c
 	kubevirt.io/api v0.59.0
-	kubevirt.io/containerized-data-importer-api v1.55.2
+	kubevirt.io/containerized-data-importer-api v1.57.0-alpha1
```
ksimon1, line 31: Is possible to use stable release instead of alpha?

akalenyu: we should have one pretty soon, and then we can switch

config/manager/kustomization.yaml
lyarwood: This is my fault, the `kubevirt` and `kubevirtci` makefile targets leave artefacts like this around in `config`. I'll raise this as an issue tomororw and will fix this up but for now please drop them from this PR.

lyarwood:
FWIW I'm also pulling in `kubevirt.io/containerized-data-importer-api v1.57.0-alpha1` in https://github.com/kubevirt/ssp-operator/pull/593

go.mod
codingben: @akalenyu Will you change it in another PR? If yes, is there an issue or ticket to track it?

codingben:
Hi @akalenyu, we're using in this project commit convention [1]. Can you please follow it?

Please write your commits as:
```
feat: some feature
fix: some bug
refactor: some code
docs: update readme
chore: some makefile changes
```

[1] https://www.conventionalcommits.org/en/v1.0.0

config/manager/kustomization.yaml
akalenyu: done

go.mod
akalenyu: usually prefer to pull in releases instead of stray commits which we are doing, but I can open an issue here to track this
@lyarwood wdyt?

internal/operands/data-sources/reconcile.go
```diff
@@ -396,7 +396,7 @@ func reconcileDataSource(dsInfo dataSourceInfo, request *common.Request) (common
 
 			foundDs := foundRes.(*cdiv1beta1.DataSource)
 			newDs := newRes.(*cdiv1beta1.DataSource)
-			if !dsInfo.autoUpdateEnabled || foundDs.Spec.Source.PVC == nil {
+			if !dsInfo.autoUpdateEnabled || (foundDs.Spec.Source.PVC == nil && foundDs.Spec.Source.Snapshot == nil) {
```
lyarwood, line 399: I forget why SSP even needs to update the `PVC` here, shouldn't CDI keep the DS updated using the DIC?

lyarwood (commented):
/approve

This makes sense for the new Snapshot source, I've got a somewhat related question about the original PVC logic if you could take a look as well. That's likely something for a follow up if we can drop it.

internal/operands/data-sources/reconcile.go
akalenyu: I guess the DataSources can be also managed by something else (dsInfo.autoUpdateEnabled=false),
in that case, if there's nothing there we put some defaults on?

lyarwood: Yeah apologies now I remember, these are just empty `DataSources` without an associated `DataImportCron` that SSP creates for each template. Something we can hopefully drop when/if templates go away in the future and instead just rely on CDI and `DataImportCrons` to create and manage `DataSources` for us.

PR #608, opened 2023-06-29
fix(CNV-28185): add required labels

codingben:

Add required labels:
- "app.kubernetes.io/managed-by"
- "app.kubernetes.io/version"
- "app.kubernetes.io/component"
- "app.kubernetes.io/part-of"

That are required by a specific resource:
- ssp-operator-metrics

Jira-Url: https://issues.redhat.com/browse/CNV-28185

**Release note**:

```release-note
Add required labels to SSP operator metrics
```

### DIFF
--- a/controllers/services_controller.go
+++ b/controllers/services_controller.go
@@ -21,21 +21,29 @@ import (
 )
 
 const (
-	MetricsServiceName    = "ssp-operator-metrics"
-	OperatorName          = "ssp-operator"
-	serviceControllerName = "service-controller"
+	ServiceManagedByLabelValue = "ssp-operator-services"
+	MetricsServiceName         = "ssp-operator-metrics"
+	OperatorName               = "ssp-operator"
+	ServiceControllerName      = "service-controller"
 )
 
-func ServiceObject(namespace string) *v1.Service {
+func ServiceObject(namespace string, appKubernetesPartOfValue string) *v1.Service {
 	policyCluster := v1.ServiceInternalTrafficPolicyCluster
 	familyPolicy := v1.IPFamilyPolicySingleStack
+	labels := map[string]string{
+		common.AppKubernetesManagedByLabel: ServiceManagedByLabelValue,
+		common.AppKubernetesVersionLabel:   common.GetOperatorVersion(),
+		common.AppKubernetesComponentLabel: ServiceControllerName,
+		metrics.PrometheusLabelKey:         metrics.PrometheusLabelValue,
+	}
+	if appKubernetesPartOfValue != "" {
+		labels[common.AppKubernetesPartOfLabel] = appKubernetesPartOfValue
+	}
 	return &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      MetricsServiceName,
 			Namespace: namespace,
-			Labels: map[string]string{
-				metrics.PrometheusLabelKey: metrics.PrometheusLabelValue,
-			},
+			Labels:    labels,
 		},
 		Spec: v1.ServiceSpec{
 			InternalTrafficPolicy: &policyCluster,
@@ -67,7 +75,7 @@ func CreateServiceController(ctx context.Context, mgr ctrl.Manager) (*serviceRec
 }
 
 func (r *serviceReconciler) Name() string {
-	return serviceControllerName
+	return ServiceControllerName
 }
 
 func (r *serviceReconciler) Start(ctx context.Context, mgr ctrl.Manager) error {
@@ -84,7 +92,8 @@ func (r *serviceReconciler) setServiceOwnerReference(service *v1.Service) error 
 }
 
 func (r *serviceReconciler) createMetricsService(ctx context.Context) error {
-	service := ServiceObject(r.serviceNamespace)
+	appKubernetesPartOfValue := r.deployment.GetLabels()[common.AppKubernetesPartOfLabel]
+	service := ServiceObject(r.serviceNamespace, appKubernetesPartOfValue)
 	err := r.setServiceOwnerReference(service)
 	if err != nil {
 		return fmt.Errorf("error setting owner reference: %w", err)
@@ -144,7 +153,8 @@ func newServiceReconciler(ctx context.Context, mgr ctrl.Manager) (*serviceReconc
 
 func (r *serviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
 	r.log.Info("Starting service reconciliation...", "request", req.String())
-	service := ServiceObject(req.Namespace)
+	appKubernetesPartOfValue := r.deployment.GetLabels()[common.AppKubernetesPartOfLabel]
+	service := ServiceObject(req.Namespace, appKubernetesPartOfValue)
 	var foundService v1.Service
 	foundService.Name = service.Name
 	foundService.Namespace = service.Namespace
--- a/tests/service_controller_test.go
+++ b/tests/service_controller_test.go
@@ -14,12 +14,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"kubevirt.io/ssp-operator/controllers"
+	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/operands/metrics"
 	"kubevirt.io/ssp-operator/tests/env"
 )
 
 func getSspMetricsService() (*v1.Service, error) {
-	service := controllers.ServiceObject(strategy.GetSSPDeploymentNameSpace())
+	service := controllers.ServiceObject(strategy.GetSSPDeploymentNameSpace(), "")
 	err := apiClient.Get(ctx, client.ObjectKeyFromObject(service), service)
 	return service, err
 }
@@ -36,6 +37,16 @@ var _ = Describe("Service Controller", func() {
 	It("[test_id: 8807] Should create ssp-operator-metrics service", func() {
 		_, serviceErr := getSspMetricsService()
 		Expect(serviceErr).ToNot(HaveOccurred(), "Failed to get ssp-operator-metrics service")
+	})
+
+	It("[test_id: TODO] Service ssp-operator-metrics should contain required labels", func() {
+		service, serviceErr := getSspMetricsService()
+		Expect(serviceErr).ToNot(HaveOccurred(), "Failed to get ssp-operator-metrics service")
+
+		Expect(service.GetLabels()).To(HaveKeyWithValue(common.AppKubernetesManagedByLabel, controllers.ServiceManagedByLabelValue))
+		Expect(service.GetLabels()).To(HaveKeyWithValue(common.AppKubernetesVersionLabel, common.GetOperatorVersion()))
+		Expect(service.GetLabels()).To(HaveKeyWithValue(common.AppKubernetesComponentLabel, controllers.ServiceControllerName))
+		Expect(service.GetLabels()[common.AppKubernetesPartOfLabel]).To(BeEmpty())
 	})
 
 	It("[test_id: 8808] Should re-create ssp-operator-metrics service if deleted", func() {


internal/common/labels.go
```diff
@@ -18,6 +18,7 @@ const (
 	AppComponentTektonTasks           AppComponent = "tektonTasks"
 	AppKubernetesManagedByValue       string       = "ssp-operator"
 	TektonAppKubernetesManagedByValue string       = "tekton-tasks-operator"
+	AppKubernetesPartOfValue          string       = "hyperconverged-cluster"
```
0xFelix, line 21: Is that always the case?

config/rbac/service_account.yaml
```diff
@@ -4,4 +4,8 @@ metadata:
   name: operator
   namespace: system
   labels:
+    app.kubernetes.io/managed-by: "ssp-operator"
```
0xFelix, line 7: Is this change useful? AFAIK this only applies to deployments of SSP withouth CSV.

akrejcir: If i understand correctly, the value of this label is the component that manages this `ServiceAccount`. But `ssp-operator` does not manage it, it does not create it.
It's created and managed either by HCO or OLM.

config/rbac/service_account.yaml
```diff
@@ -4,4 +4,8 @@ metadata:
...
+    app.kubernetes.io/version: "0.0.0"
+    app.kubernetes.io/component: "ssp-operator"
+    app.kubernetes.io/part-of: "hyperconverged-cluster"
```
akrejcir, line 10: This value `hyperconverged-cluster` should probably be set by HCO or OLM.

internal/common/labels.go
akrejcir: This value should not be hardcoded. It is copied from the SSP CR:
https://github.com/kubevirt/ssp-operator/blob/e92bc396eca2195d7d43b30d8f9fb363fb418f47/internal/common/labels.go#L36-L49

Because it is set by HCO or OLM.

internal/operands/template-validator/resources.go
```diff
@@ -42,6 +42,13 @@ const (
 
 func CommonLabels() map[string]string {
 	return map[string]string{
+		// Required Labels
+		common.AppKubernetesManagedByLabel: common.AppKubernetesManagedByValue,
+		common.AppKubernetesVersionLabel:   common.GetOperatorVersion(),
+		common.AppKubernetesComponentLabel: VirtTemplateValidator,
+		common.AppKubernetesPartOfLabel:    common.AppKubernetesPartOfValue,
```
akrejcir, line 49: These labels are already added, see:
https://github.com/kubevirt/ssp-operator/blob/e92bc396eca2195d7d43b30d8f9fb363fb418f47/internal/common/labels.go#L36-L49

codingben: Yes, but this is no used by `virt-template-validator`, specifically `WithAppLabels()`. There is just `CommonLabels()`. Should we use `WithAppLabels()`` instead of `CommonLabels()`?

config/rbac/service_account.yaml
codingben: Okay, how we can achieve it? by changing some HCO code to set it? This `service_account.yaml` is not part of some code, where we can use something like `WithAppLabels()`.

config/rbac/service_account.yaml
codingben: Thanks for the info, it means HCO should set those labels?

akrejcir: I'm not sure. I think the SA is created by OLM. Can you look into if OLM can create these labels?

config/rbac/service_account.yaml
codingben: Hi @orenc1 @tiraboschi @nunnatsa, do you know, by chance, if these labels can be set by OLM? Is it required by [1].

[1] https://issues.redhat.com/browse/CNV-9097

orenc1: yes, HCO is setting the labels on all underlying components when constructing the CSV:
see here:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/901f0dd11ca46c928033dd8987edecb0c601b2ea/tools/csv-merger/csv-merger.go#L541C43-L541C64

config/rbac/service_account.yaml
codingben: I see that HCO is deploying SSP's service-account [1] so I guess it's on HCO side now.

[1] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/deploy/service_account.yaml#L42-L48

config/rbac/service_account.yaml
codingben: Okay, I see. Question is, why the service-account here [1] doesn't contain any labels? also the same question about other resources that HCO deploys, e.g. [2].

[1] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/deploy/service_account.yaml#L42-L48
[2] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/deploy/operator.yaml#L498-L500

akrejcir:
Please update the commit message and PR's description to reflect the current change.

codingben:
> Please update the commit message and PR's description to reflect the current change.

Done.

controllers/services_controller.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 const (
-	MetricsServiceName    = "ssp-operator-metrics"
-	OperatorName          = "ssp-operator"
-	serviceControllerName = "service-controller"
+	MetricsServiceName       = "ssp-operator-metrics"
+	OperatorName             = "ssp-operator"
+	serviceControllerName    = "service-controller"
+	appKubernetesPartOfValue = "hyperconverged-cluster"
```
0xFelix, line 27: I am not sure about this value. What if SSP is not deployed by HCO? Shouldn't this value come from HCO?

codingben: I removed it.

controllers/services_controller.go
```diff
@@ -26,14 +26,21 @@ const (
...
 	return &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      MetricsServiceName,
 			Namespace: namespace,
 			Labels: map[string]string{
+				// Required Labels
+				common.AppKubernetesManagedByLabel: common.AppKubernetesManagedByValue,
+				common.AppKubernetesVersionLabel:   common.GetOperatorVersion(),
+				common.AppKubernetesComponentLabel: serviceControllerName,
+				common.AppKubernetesPartOfLabel:    appKubernetesPartOfValue,
```
0xFelix, line 41: Omit this label if `appKubernetesPartOfValue` is empty?

0xFelix (commented):
There is still a CI failure.

0xFelix (commented):
Can you please add functional tests too?

codingben:
/hold

There is a failure in the new functional test that I've added, I'll try to run it locally again.

tests/service_controller_test.go
```diff
@@ -38,6 +39,21 @@ var _ = Describe("Service Controller", func() {
...
+
+		managedByValue := service.GetLabels()[common.AppKubernetesManagedByLabel]
+		versionLabelValue := service.GetLabels()[common.AppKubernetesVersionLabel]
+		componentLabelValue := service.GetLabels()[common.AppKubernetesComponentLabel]
+		partOfValue := service.GetLabels()[common.AppKubernetesPartOfLabel]
+
+		Expect(managedByValue).To(Equal(common.AppKubernetesManagedByValue))
+		Expect(versionLabelValue).To(Equal(common.GetOperatorVersion()))
+		Expect(componentLabelValue).To(Equal(controllers.ServiceControllerName))
+		Expect(partOfValue).To(BeEmpty())
```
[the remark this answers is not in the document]
codingben: I mean I'd expect it to by set when HCO will set SSP's deployment object, so then:
```
func (r *serviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
	r.log.Info("Starting service reconciliation...", "request", req.String())
	appKubernetesPartOfValue := r.deployment.GetLabels()[common.AppKubernetesPartOfLabel]
	service := ServiceObject(req.Namespace, appKubernetesPartOfValue)
...
```
`appKubernetesPartOfValue` will be not empty.

codingben:
I don't see that the failures are related to my changes.

/retest

akrejcir:
Probably flakes.
/retest

codingben:
Yes, I see again infra failures such as:
```
{  openshift cluster install failed with cluster creation}
```

/retest

tests/service_controller_test.go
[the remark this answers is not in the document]
codingben: > One where the label is set to the value you pass to SSP

Isn't something that HCO should test? because HCO will set the part of label, and not SSP.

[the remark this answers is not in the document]
0xFelix: No, this is a feature of SSP so SSP needs to test it.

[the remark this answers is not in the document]
codingben: Sorry, I don't understand then.

1. HCO will deploy SSP [1] with the `part-of` label. This is something we can't test here.
2. If we want to set `part-of` label to SSP deployment, then we should add it to [2] (?) and then functional test can test it. This is what you mean?

[1] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/deploy/operator.yaml#L483
[2] https://github.com/kubevirt/ssp-operator/blob/main/config/manager/manager.template.yaml#L18-L21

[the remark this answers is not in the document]
akrejcir: To test this, you would need to add the label to the SSP deployment, and then restart the operator, by scaling it down to 0 and back up. 
It seems like too complex test just for label propagation.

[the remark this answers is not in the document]
codingben: Yes, that's why I don't think it should be tested here, but maybe in HCO (if they have tests for labels at all).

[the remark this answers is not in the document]
codingben: IMO, functional test should test the existing behavior of the SSP deployment, and not try to modify it and then test it, especially if that is only label - it doesn't have some critical priority in SSP deployment operation. SSP won't crash if that label will be missed.

akrejcir:
The failing test checks if all resources with label `app.kubernetes.io/managed-by: "ssp-operator"` have been removed when the SSP CR is removed. But the `ssp-operator-metrics` service is created when the operator starts, and is not affected by the SSP CR.

@codingben Can you change the test, so that it ignores this one service?

https://github.com/kubevirt/ssp-operator/blob/eb60b158ff1d5914f6ec3cfdd32d51b5e884f4e3/tests/cleanup_test.go#L33

tests/service_controller_test.go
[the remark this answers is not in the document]
0xFelix: Agreed, the test is expensive. But this is literally a feature of SSP, so why would somebody else test this for us? A feature can be non-critical and yet it is a feature somebody could rely on, it doesn't matter if we crash or not if the feature is not working. So in order to avoid regressions we should test it.

codingben:
> Can you add a link to the jira ticket to the PR's description and the commit message? When a ticket number is in the name, I would expect a link.

Maybe then we would add that to the PR template, similar to HCO's template [1]?

[1] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L24-L30

akrejcir:
There is already a place of the issue in the template:
https://github.com/kubevirt/ssp-operator/blob/d4cfe2cdd499504fb62a3a8c7ac79b2a390a9aa9/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L7-L10

But you can change the template if you want.

tests/service_controller_test.go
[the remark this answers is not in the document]
codingben: > I missed that the metrics service is also reconciled. It is enough to delete the service, and test that it was recreated with changed labels.

But who will add the part-of label to the SSP deployment? the functional test itself?

codingben:
> The failing test checks if all resources with label `app.kubernetes.io/managed-by: "ssp-operator"` have been removed when the SSP CR is removed. But the `ssp-operator-metrics` service is created when the operator starts, and is not affected by the SSP CR.
> 
> @codingben Can you change the test, so that it ignores this one service?
> 
> https://github.com/kubevirt/ssp-operator/blob/eb60b158ff1d5914f6ec3cfdd32d51b5e884f4e3/tests/cleanup_test.go#L33

@akrejcir Thanks for spotting it! :) Can we just change the `managed-by` label value of service object to `ssp-operator-metrics` instead of `ssp-operator`? It would also not need to modify the functional test.

akrejcir:
> @akrejcir Thanks for spotting it! :) Can we just change the `managed-by` label value of service object to `ssp-operator-metrics` instead of `ssp-operator`? It would also not need to modify the functional test.

I'm not sure. The meaning of the label is: `The tool being used to manage the operation of an application` https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

There is no tool called `ssp-operator-metrics`. But the example in the documentation is `helm`, so maybe we are using the label incorrectly.

codingben:
> > @akrejcir Thanks for spotting it! :) Can we just change the `managed-by` label value of service object to `ssp-operator-metrics` instead of `ssp-operator`? It would also not need to modify the functional test.
> 
> I'm not sure. The meaning of the label is: `The tool being used to manage the operation of an application` https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
> 
> There is no tool called `ssp-operator-metrics`. But the example in the documentation is `helm`, so maybe we are using the label incorrectly.

I think there is a design issue from the beginning. Why metrics is part of ssp-operator? why ssp-operator deploys Metrics Rules and many other resources?

From what I see, the service object `ssp-operator-metrics` is controlled by `services_controller`, and all other resources are controlled by `ssp_controller`. Maybe it would be okay to have the `managed-by` label as `ssp-operator-services` or `ssp-operator-metrics`.

What others think, @lyarwood and @0xFelix? your opinion is very much appreciated. :)

0xFelix:
@codingben Depends on what we want to achieve.

If we want absolutely not resources to be left when the SSP CR is removed then we have another bug. If we want just objects from `ssp_controller` to be gone then we could change the label value.

IMO we should not leave any resources when the SSP CR is removed, so IMO we have a bug?

0xFelix:
How do we ensure the resources are cleaned up when removing the operator then?

0xFelix:
@akrejcir @codingben Then it sounds to me like we should have different `managed-by` labels for the operands.

0xFelix:
AFAIU they are used by HCO.

codingben:
In the failed tests I see some infra failures, e.g. network/timeout errors.

/retest

codingben:
@akrejcir @0xFelix @ksimon1 @lyarwood Hi, finally CI passed. Can you please merge it?

0xFelix (approved):
/approve

Thanks!

PR #611, opened 2023-06-30
fix: fix default instanceTypeName parameter in tekton pipelines

ksimon1:
fix: fix default instanceTypeName parameter in tekton pipelines

n1.large instance type was renamed to o1.large

**Release note**:
```
fix default instanceTypeName parameter in tekton pipelines
```

### DIFF
--- a/data/tekton-pipelines/windows-bios-installer-pipeline.yaml
+++ b/data/tekton-pipelines/windows-bios-installer-pipeline.yaml
@@ -11,7 +11,7 @@ spec:
     - name: instanceTypeName
       type: string
       description: Name of VirtualMachineClusterInstancetype object
-      default: n1.large
+      default: u1.large
     - name: instanceTypeKind
       type: string
       description: Kind of VirtualMachineInstancetype object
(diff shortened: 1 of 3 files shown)


data/tekton-pipelines/windows-bios-installer-pipeline.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     - name: instanceTypeName
       type: string
       description: Name of VirtualMachineClusterInstancetype object
-      default: n1.large
+      default: o1.large
```
lyarwood, line 14: `u1` replaces `n1` in `v0.3.0` common-instancetypes.

lyarwood (changes requested):
Apologies I didn't realise we were directly referencing these types from within SSP. I'll make a note in the repo to update this and HCO together in the future.

lyarwood:
We should also follow up and ensure the associated pipelines are ran when the common-instancetypes bundle is updated in the future to catch things like this.

PR #615, opened 2023-07-02
chore: rename tekton operands

codingben:

Rename Tekton operands from tekton-*.go to
reconcile.go to be similar as in other operands.

Not necessarily needed to be backported. As you wish.

**Release note**:

```release-note
None
```


0xFelix (commented):
/approve

I'm fine with it, but please no backport.

ksimon1:
/lgtm
please bare in your mind, that midstream dockerfile will have to be updated after this change.
And I agree with Felix, this PR should not be backported.

codingben:
@ksimon1 Yes, that dockerfile was updated. Thanks!

PR #617, opened 2023-07-04
git rm -rf 

opokornyy:

This PR adds a notification to the README that the development has been moved from the master branch to the main branch.

**Release note**:
```
None
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+# Branch name change:
+
+## Development has been moved from the master branch to the main branch.
+
 # SSP Operator
 
 Operator that deploying and controlling additional [KubeVirt](https://kubevirt.io) resources:


0xFelix (commented):
Can't we just drop the whole branch instead?

opokornyy:
> Can't we just drop the whole branch instead?

@lyarwood what do you think about dropping the branch?

lyarwood:
The only time I've done this in the past we kept `master` around to point people towards `main` but that might not be required here so I don't really mind either way. I'd like to hear input from @akrejcir (when you're back next week) , @ksimon1 and @codingben if possible before we decide either way.

codingben:
I'd vote to rename `master` to `master-deprecated` and keep it, then after a few weeks remove it (or just keep it).

ksimon1:
I agree to keep the branch for some time and then removes it

opokornyy:
Can we merge this PR so there is a notification in the README, and then later rename/delete the master branch?

lyarwood:
> Can we merge this PR so there is a notification in the README, and then later rename/delete the master branch?

/lgtm

Okay this works for me.

0xFelix (approved):
/approve

ACK, when do we want to remove the master branch?

opokornyy:
> /approve
> 
> ACK, when do we want to remove the master branch?

Lets wait for @akrejcir when he is back and then decide.

akrejcir:
In `kubevirt` they removed the `master` branch. I think we can keep it temporarily, maybe until the next release, and then remove it.

PR #616, opened 2023-07-03
Audit RBAC rules

jcanocan:
**What this PR does / why we need it**:
As it is now, some of the RBAC rules and/or verbs are not required for the operator to be able to perform all intended operations.
This PR audits all RBAC rules. The auditory has been carried out following a black box approach, i.e., remove all rules, deploy the operator, run functest, collect forbidden operations, add the missing rules/verbs and repeat. The auditing process has been split in the next action items:

- Deploying the operator in KubevirtCI.
- Deploying the operator in a CRC cluster and running the functest.
- Deploying the operator in a 3 masters and 3 workers nodes and running the functest.
- Deploying a custom downstream build and running the Tier 2 test.

Just in case changes are not clear enough, this is a simplified and normalized diff of the `ClusterRole` file before and after: [https://www.textcompare.org/?id=64a2c9d599377eb6104d8ae3](https://www.textcompare.org/?id=64a2c9d599377eb6104d8ae3) 

The simplified version has been obtained using this [script](https://github.com/jcanocan/rbacSimplifier).

jira-ticket: https://issues.redhat.com/browse/CNV-24031

**Which issue(s) this PR fixes**: 

Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
Reduce the `ClusterRole` rules used
```

PR #622, opened 2023-07-10
chore: Audit template validator RBAC rules.

jcanocan:
This PR is a follow-up of https://github.com/kubevirt/ssp-operator/pull/616. For this auditory, we have used the same methodology described in #616. However, due to the fact that these rules are implemented in code, we have not been able to try on Tier 2 test suite. 

jira-ticket: https://issues.redhat.com/browse/CNV-30721

Fixes #

**Release note**:

```release-note
Audit template validator RBAC rules
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -268,7 +268,6 @@ rules:
   resources:
   - virtualmachines
   verbs:
-  - get
   - list
   - watch
 - apiGroups:
--- a/data/olm-catalog/ssp-operator.clusterserviceversion.yaml
+++ b/data/olm-catalog/ssp-operator.clusterserviceversion.yaml
@@ -329,7 +329,6 @@ spec:
           resources:
           - virtualmachines
           verbs:
-          - get
           - list
           - watch
         - apiGroups:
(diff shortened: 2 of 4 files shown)


0xFelix (approved):
/approve

Thanks!

akrejcir:
@jcanocan @0xFelix , I'd like to know your opinion about RBAC permissions. 
I always thought about them as 2 groups:
- Read permissions: `get`, `list`, `watch`
- Write permissions: `create`, `update`, `delete`, `patch`

When I created rules for the operator, I always added a whole group. I find it strange that we give the operator permission to `list` and `watch` but not `get`.

What do you think?

jcanocan:
> @jcanocan @0xFelix , I'd like to know your opinion about RBAC permissions. I always thought about them as 2 groups:
> 
>     * Read permissions: `get`, `list`, `watch`
> 
>     * Write permissions: `create`, `update`, `delete`, `patch`
> 
> 
> When I created rules for the operator, I always added a whole group. I find it strange that we give the operator permission to `list` and `watch` but not `get`.
> 
> What do you think?

Yes, it may be a bit strange. However, I've followed an approach where if something is not required, drop it. My idea was to get a minimal list of verbs. Nevertheless, it does not mean that make sense. So, we can revert all `get` verbs if that makes more sense.

akrejcir:
No need to revert the change. I just wanted to know your opinion.

0xFelix:
I agree with @jcanocan, we should follow the need to know principle. Even if those verbs are pretty similar.

akrejcir:
Ok, thanks.

jcanocan:
> @jcanocan The jira epic is targeted to `4.14`, so this should be backported to `release-v0.18`.

Thanks for taking care of this on my behalf :)

PR #628, opened 2023-07-19
fix: GH actions for bumping virtio version

ksimon1:
fix: GH actions for bumping virtio version
script contained wrong paths for files, which need update

**Release note**:
```
NONE

```


0xFelix (approved):
/lgtm

PR #629, opened 2023-07-19
feat: Add github action that will add reviewer

opokornyy:

This PR should automate adding given users to monitoring PRs.

**Release note**:
```
None
```

### DIFF
--- a/.github/workflows/add-reviewer-monitoring.yaml
+++ b/.github/workflows/add-reviewer-monitoring.yaml
@@ -1,0 +1,25 @@
+name: Add reviewer from sig-monitoring to PR
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+      - internal/operands/metrics/*
+      - internal/common/resource.go
+      - docs/metrics.md
+      - tests/metrics_test_utils.go
+      - tools/metricsdocs/*
+
+jobs:
+  add_reviewer:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Add Reviewer
+        id: add_reviewer
+        run: gh pr edit ${{ github.event.pull_request.number }} --add-reviewer sradco
+        env:
+          GITHUB_TOKEN: ${{ github.token }}


akrejcir:
Have you tested this in your own fork?

opokornyy:
> Have you tested this in your own fork?

Yes, I did.

.github/workflows/add-reviewer-monitoring.yaml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Add Reviewer
+        id: add_reviewer
+        run: gh pr edit $PULL_REQUEST_NUMBER --add-reviewer sradco
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          PULL_REQUEST_NUMBER: ${{  github.event.pull_request.number }}
```
0xFelix, line 26: Can you inline this?

.github/workflows/add-reviewer-monitoring.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - docs/metrics.md
+      - tests/metrics_test_utils.go
+      - tools/metricsdocs/*
+
+jobs:
+  add_reviewer:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
```
0xFelix, line 18: Do we really need to checkout the code first?

.github/workflows/add-reviewer-monitoring.yaml
```diff
@@ -0,0 +1,26 @@
+name: Add reviewer on monitoring PRs
```
0xFelix, line 1: Can you name it `Add reviewer from sig-monitoring to PR`?

.github/workflows/add-reviewer-monitoring.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+on:
+  pull_request:
+    types: [opened]
```
0xFelix, line 5: Is there a type for when an existing PR is updated to change a file from this list?

.github/workflows/add-reviewer-monitoring.yaml
opokornyy: Yes, without checkout the `gh` command does not work properly.

.github/workflows/add-reviewer-monitoring.yaml
opokornyy: Yes, I have missed this case, not it should work also on updating existing PR

lyarwood:
/approve

It's a hackaround, still would prefer to see this done with OWNERS but as we saw it's awkward and we can't enforce reviewers so this will have to do for now.

PR #609, opened 2023-06-29
feat: add deploy VM console proxy feature gate

codingben:
**What this PR does / why we need it**:

Add a new Feature Gate to deploy VM console proxy
and set namespace.

Jira-Url: https://issues.redhat.com/browse/CNV-30283

**Release note**:

```release-note
Add VM console proxy feature gate.
```

PR #634, opened 2023-07-23
refactor(CNV-31248): remove namespace annotation

codingben:

Remove vm-console-proxy-namespace annotation that
is no longer relevant since a new released VM
console proxy is no longer require namespace
to operate, by default it was kubevirt as it
required an access to virt-handler (and also
virt-handler certificates).

Jira-Url: https://issues.redhat.com/browse/CNV-31248

**Release note**:

```release-note
Remove vm-console-proxy-namespace annotation
```

### DIFF
--- a/internal/operands/vm-console-proxy/reconcile.go
+++ b/internal/operands/vm-console-proxy/reconcile.go
@@ -18,8 +18,6 @@ import (
 )
 
 const (
-	VmConsoleProxyNamespaceAnnotation = "ssp.kubevirt.io/vm-console-proxy-namespace"
-
 	operandName      = "vm-console-proxy"
 	operandComponent = "vm-console-proxy"
 
@@ -55,11 +53,16 @@ func WatchClusterTypes() []operands.WatchType {
 	return []operands.WatchType{
 		{Object: &rbac.ClusterRole{}},
 		{Object: &rbac.ClusterRoleBinding{}},
+		{Object: &apiregv1.APIService{}},
+	}
+}
+
+func WatchTypes() []operands.WatchType {
+	return []operands.WatchType{
 		{Object: &core.ServiceAccount{}},
 		{Object: &core.Service{}},
 		{Object: &apps.Deployment{}, WatchFullObject: true},
 		{Object: &core.ConfigMap{}},
-		{Object: &apiregv1.APIService{}},
 		{Object: &routev1.Route{}},
 	}
 }
@@ -95,7 +98,7 @@ func (v *vmConsoleProxy) Name() string {
 }
 
 func (v *vmConsoleProxy) WatchTypes() []operands.WatchType {
-	return nil
+	return WatchTypes()
 }
 
 func (v *vmConsoleProxy) WatchClusterTypes() []operands.WatchType {
@@ -229,9 +232,9 @@ func (v *vmConsoleProxy) deleteRoute(request *common.Request) ([]common.CleanupR
 
 func reconcileServiceAccount(serviceAccount core.ServiceAccount) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		serviceAccount.Namespace = getVmConsoleProxyNamespace(request)
+		serviceAccount.Namespace = request.Instance.Namespace
 		return common.CreateOrUpdate(request).
-			ClusterResource(&serviceAccount).
+			NamespacedResource(&serviceAccount).
 			WithAppLabels(operandName, operandComponent).
 			Reconcile()
 	}
@@ -248,6 +251,7 @@ func reconcileClusterRole(clusterRole rbac.ClusterRole) common.ReconcileFunc {
 
 func reconcileClusterRoleBinding(clusterRoleBinding rbac.ClusterRoleBinding) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
+		clusterRoleBinding.Subjects[0].Namespace = request.Instance.Namespace
 		return common.CreateOrUpdate(request).
 			ClusterResource(&clusterRoleBinding).
 			WithAppLabels(operandName, operandComponent).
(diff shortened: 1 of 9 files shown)


internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -250,14 +248,7 @@ func reconcileRoute(serviceName string) common.ReconcileFunc {
...
 func getVmConsoleProxyNamespace(request *common.Request) string {
-	const defaultNamespace = "kubevirt"
-	if request.Instance.GetAnnotations() == nil {
-		return defaultNamespace
-	}
-	if namespace, isFound := request.Instance.GetAnnotations()[VmConsoleProxyNamespaceAnnotation]; isFound {
-		return namespace
-	}
-	return defaultNamespace
+	return "kubevirt"
```
codingben, line 251: @akrejcir As you've mentioned, `kubevirt` namespace is no longer needed to be default one (and IIUC, no namespace is needed at all to be provided by the user). That being said, `getVmConsoleProxyNamespace` can be refactored to use SSP's CR namespace.

So just to confirm, in https://github.com/kubevirt/ssp-operator/pull/609#discussion_r1246259536 you've mentioned that we can deploy vm-console-proxy resources in SSP's CR namespace, right?

/cc @lyarwood @0xFelix

0xFelix: How is the deployment of the template validator handled? Can we do the same for vm-console-proxy?

codingben:
I'm trying to debug the failed unit tests, they are failing because there is no annotation(s). That's weird, because in previous PR it worked, now it's failing because probably we try to set different namespace in the unit tests.

internal/operands/vm-console-proxy/reconcile.go
codingben: Okay, but I see the implementation of `NamespacedResource()` that is:
```
func (r *reconcileBuilder) NamespacedResource(resource client.Object) ReconcileBuilder {
	r.resource = resource
	r.isClusterResource = false
	return r
}
```
I don't see in the code how it would set the object's namespace, that was done by `someObject.namespace = someNamespace`. Do you know by chance? maybe we need to anyway keep setting the namespace? e.g. `serviceAccount.namespace = request.Instance.Namespace`?

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -183,9 +181,9 @@ func (v *vmConsoleProxy) Cleanup(request *common.Request) ([]common.CleanupResul
 
 func reconcileServiceAccount(serviceAccount core.ServiceAccount) common.ReconcileFunc {
 	return func(request *common.Request) (common.ReconcileResult, error) {
-		serviceAccount.Namespace = getVmConsoleProxyNamespace(request)
+		serviceAccount.Namespace = request.Instance.Namespace
```
0xFelix, line 184: Deployment of template validator is using `request.Namespace`. What's the difference?

See e.g. https://github.com/kubevirt/ssp-operator/blob/fc1d67c1b3ee193d02cab94149217525975c1f0c/internal/operands/template-validator/reconcile.go#L96

codingben: IMO, both should return the same value.

1. `request.Instance.Namespace` as a reference to the SSP object, where is deployed so then it makes sense to me to use it.
2. `request.Namespace` seems like a requested namespace where the SSP object should be reconciled, thus I don't think `request.Namespace` should be used in operands but in `ssp_controller` [1].

Tekton operand is using `request.Instance.Namespace` as well [2]:
```
func getTektonTasksNamespace(request *common.Request) string {
	if request.Instance.Spec.TektonTasks != nil && request.Instance.Spec.TektonTasks.Namespace != "" {
		return request.Instance.Spec.TektonTasks.Namespace
	}
	return request.Instance.Namespace
}
```

@akrejcir @ksimon1 WDYT?

[1] https://github.com/kubevirt/ssp-operator/blob/main/controllers/ssp_controller.go#L129-L164
[2] https://github.com/kubevirt/ssp-operator/blob/main/internal/operands/tekton-tasks/reconcile.go#L304-L309

codingben:
> IIUC `GetVmConsoleProxyNamespace` should be dropped from the test suite and instead `GetNamespace` should be used, because vm-console-proxy is deployed in the same namespace as SSP?

You're right, thanks! I forgot about it.

codingben:
There is an issue with functional tests, I'm trying to debug it.

internal/operands/vm-console-proxy/reconcile.go
akrejcir: Both are the same value, so it does not matter.

0xFelix:
Actually there is a `/vnc` endpoint now, which `/token` was renamed to. It should still be accessible, see https://github.com/kubevirt/vm-console-proxy/blob/main/docs/api.md#example.

akrejcir:
PR https://github.com/kubevirt/ssp-operator/pull/645 has been merged.

codingben:
> PR #645 has been merged.

I rebased and resolved conflicts. `make container-build` passed locally.

akrejcir:
The `VmConsoleProxyNamespace()` function in `/tests/env/env.go` is not used anymore and can be removed. The same applies for any mention of environment variable `VM_CONSOLE_PROXY_NAMESPACE`

codingben:
I see CI job are failing and locally as well, functional tests:
```
Summarizing 5 Failures:
  [FAIL] VM Console Proxy Operand Resource creation created cluster resource [It] [test_id:9848] service account
  /home/*/GitHub/codingben/ssp-operator/tests/vm_console_proxy_test.go:156
  [FAIL] VM Console Proxy Operand Resource deletion recreate after delete [It] [test_id:9866] config map
  /home/*/GitHub/codingben/ssp-operator/tests/tests_common_test.go:105
  [FAIL] VM Console Proxy Operand Resource change should restore modified resource [It] [test_id:9871] config map
  /home/*/GitHub/codingben/ssp-operator/tests/tests_common_test.go:182
  [FAIL] VM Console Proxy Operand Resource change should restore modified app labels [It] [test_id:9884] config map
  /home/*/GitHub/codingben/ssp-operator/tests/labels_test.go:78
  [FAIL] VM Console Proxy Operand Accessing proxy [It] [test_id:TODO] should be able to access /vnc endpoint
  /home/*/GitHub/codingben/ssp-operator/tests/vm_console_proxy_test.go:336

Ran 35 of 289 Specs in 423.144 seconds
FAIL! -- 30 Passed | 5 Failed | 1 Pending | 253 Skipped
```

akrejcir:
Please rebase on top of latest `main`. But good catch, that the CI does not do rebase.

PR #640, opened 2023-08-01
fix: CrashLoopBackOff once tlsProfile changed

opokornyy:

This PR adds the usage of callbacks to Prometheus and the webhook server, enabling the dynamic reloading of the TLS configuration. This eliminates the necessity of restarting the SSP pod, which had previously led to the pod entering the CrashLoopBackOff state.

Fixes #

https://bugzilla.redhat.com/show_bug.cgi?id=2151248

**Release note**:
```
None
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"os"
 	"reflect"
 	"strconv"
 	"strings"
@@ -148,7 +147,6 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
-	restartNeeded := r.isRestartNeeded(instance)
 	r.clearCacheIfNeeded(instance)
 
 	sspRequest := &common.Request{
@@ -161,10 +159,6 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 		VersionCache:   r.subresourceCache,
 		TopologyMode:   r.topologyMode,
 		CrdList:        r.crdList,
-	}
-
-	if restartNeeded {
-		r.restart(sspRequest)
 	}
 
 	if !isInitialized(sspRequest.Instance) {
@@ -233,25 +227,6 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 	}
 
 	return ctrl.Result{}, nil
-}
-
-func (r *sspReconciler) isRestartNeeded(sspObj *ssp.SSP) bool {
-	if reflect.DeepEqual(r.lastSspSpec, ssp.SSPSpec{}) {
-		return false
-	}
-	if !reflect.DeepEqual(r.lastSspSpec.TLSSecurityProfile, sspObj.Spec.TLSSecurityProfile) {
-		return true
-	}
-	return false
-}
-
-func (r *sspReconciler) restart(request *common.Request) {
-	r.log.Info("TLSSecurityProfile changed, restarting")
-	err := setSspResourceDeploying(request)
-	if err != nil {
-		r.log.Info("Error at setSspResourceDeploying", "Error: ", err)
-	}
-	os.Exit(0)
 }
 
 func (r *sspReconciler) clearCacheIfNeeded(sspObj *ssp.SSP) {
--- a/internal/common/crypto_policy.go
+++ b/internal/common/crypto_policy.go
@@ -1,16 +1,12 @@
 package common
 
 import (
-	"context"
 	"crypto/tls"
 	"fmt"
 
 	"github.com/go-logr/logr"
 	ocpv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/crypto"
-	ssp "kubevirt.io/ssp-operator/api/v1beta2"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type SSPTLSOptions struct {
@@ -65,8 +61,8 @@ func GetKnownCipherId(IANACipherName string) (uint16, bool) {
 	return 0, false
 }
 
-func (s *SSPTLSOptions) CipherIDs(logger *logr.Logger) (cipherSuites []uint16) {
-	for _, cipherName := range crypto.OpenSSLToIANACipherSuites(s.OpenSSLCipherNames) {
+func CipherIDs(names []string, logger *logr.Logger) (cipherSuites []uint16) {
+	for _, cipherName := range crypto.OpenSSLToIANACipherSuites(names) {
 		if id, ok := GetKnownCipherId(cipherName); ok {
 			cipherSuites = append(cipherSuites, id)
 		} else {
@@ -76,33 +72,6 @@ func (s *SSPTLSOptions) CipherIDs(logger *logr.Logger) (cipherSuites []uint16) {
 		}
 	}
 	return
-}
-
-func GetSspTlsOptions(ctx context.Context) (*SSPTLSOptions, error) {
-	setupLog := ctrl.Log.WithName("setup tls options")
-	restConfig := ctrl.GetConfigOrDie()
-	apiReader, err := client.New(restConfig, client.Options{Scheme: Scheme})
-	if err != nil {
-		return nil, err
-	}
-
-	var sspList ssp.SSPList
-	if err := apiReader.List(ctx, &sspList, &client.ListOptions{}); err != nil {
-		return nil, err
-	}
-
-	if len(sspList.Items) == 0 {
-		setupLog.Info("SSP CR not found, will use default tlsProfile")
-		return &SSPTLSOptions{}, nil
-	}
-
-	ssp := sspList.Items[0]
-
-	sspTLSOptions, err := NewSSPTLSOptions(ssp.Spec.TLSSecurityProfile, &setupLog)
-	if err != nil {
-		return nil, err
-	}
-	return sspTLSOptions, nil
 }
 
 func selectCipherSuitesAndMinTLSVersion(profile *ocpv1.TLSSecurityProfile) ([]string, ocpv1.TLSProtocolVersion) {
--- a/internal/template-validator/tlsinfo/tlsinfo.go
+++ b/internal/template-validator/tlsinfo/tlsinfo.go
@@ -192,7 +192,7 @@ func (ti *TLSInfo) CreateTlsConfig() *tls.Config {
 	}
 
 	if !ti.sspTLSOptions.IsEmpty() {
-		tlsConfig.CipherSuites = ti.sspTLSOptions.CipherIDs(nil)
+		tlsConfig.CipherSuites = common.CipherIDs(ti.sspTLSOptions.OpenSSLCipherNames, nil)
 		minVersion, err := ti.sspTLSOptions.MinTLSVersionId()
 		if err != nil {
 			panic(fmt.Sprintf("TLS Configuration broken, min version misconfigured %v", err))
--- a/main.go
+++ b/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"flag"
 	"fmt"
@@ -28,13 +29,18 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	ocpconfigv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	ssp "kubevirt.io/ssp-operator/api/v1beta2"
 	"kubevirt.io/ssp-operator/controllers"
 	"kubevirt.io/ssp-operator/internal/common"
 	common_templates "kubevirt.io/ssp-operator/internal/operands/common-templates"
@@ -66,53 +72,121 @@ const (
 	webhookPort = 9443
 )
 
-func runPrometheusServer(metricsAddr string, tlsOptions common.SSPTLSOptions) error {
+// This callback executes on each client call returning a new config to be used
+// please be aware that the APIServer is using http keepalive so this is going to
+// be executed only after a while for fresh connections and not on existing ones
+func getConfigForClient(ctx context.Context, cfg *tls.Config, cache cache.Cache) (*tls.Config, error) {
+	var sspList ssp.SSPList
+	err := cache.List(ctx, &sspList)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(sspList.Items) == 0 || sspList.Items[0].Spec.TLSSecurityProfile == nil {
+		cfg.MinVersion = crypto.DefaultTLSVersion()
+		cfg.CipherSuites = nil
+		return cfg, nil
+	}
+
+	tlsProfile := sspList.Items[0].Spec.TLSSecurityProfile
+	if tlsProfile.Type == ocpconfigv1.TLSProfileCustomType {
+		minVersion, err := crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
+		if err != nil {
+			return nil, err
+		}
+		cfg.MinVersion = minVersion
+		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &ctrl.Log)
+		return cfg, nil
+	}
+
+	minVersion, err := crypto.TLSVersion(string(ocpconfigv1.TLSProfiles[tlsProfile.Type].MinTLSVersion))
+	if err != nil {
+		return nil, err
+	}
+	cfg.MinVersion = minVersion
+	cfg.CipherSuites = common.CipherIDs(ocpconfigv1.TLSProfiles[tlsProfile.Type].Ciphers, &ctrl.Log)
+
+	return cfg, nil
+}
+
+type prometheusServer struct {
+	cache         cache.Cache
+	certPath      string
+	keyPath       string
+	serverAddress string
+}
+
+// NeedLeaderElection implements the LeaderElectionRunnable interface, which indicates
+// the prometheus server doesn't need leader election.
+func (s *prometheusServer) NeedLeaderElection() bool {
+	return false
+}
+
+func (s *prometheusServer) Start(ctx context.Context) error {
 	setupLog.Info("Starting Prometheus metrics endpoint server with TLS")
-	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
-	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
 	handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{})
 	mux := http.NewServeMux()
 	mux.Handle("/metrics", handler)
 
-	minTlsVersion, err := tlsOptions.MinTLSVersionId()
+	server := &http.Server{
+		Addr:    s.serverAddress,
+		Handler: mux,
+	}
+
+	certWatcher, err := certwatcher.New(s.certPath, s.keyPath)
 	if err != nil {
 		return err
 	}
 
-	tlsConfig := tls.Config{
-		CipherSuites: tlsOptions.CipherIDs(&setupLog),
-		MinVersion:   minTlsVersion,
-	}
-
-	server := http.Server{
-		Addr:      metricsAddr,
-		Handler:   mux,
-		TLSConfig: &tlsConfig,
-	}
-
 	go func() {
-		err := server.ListenAndServeTLS(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
-		if err != nil {
-			setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
+		// TODO: change context, so it can be closed when
+		// this function returns an error
+		if err := certWatcher.Start(ctx); err != nil {
+			setupLog.Error(err, "certificate watcher error")
 		}
 	}()
+
+	idleConnsClosed := make(chan struct{})
+	go func() {
+		// TODO: make sure that the goroutine finishes when
+		// this function returns an error
+		<-ctx.Done()
+		setupLog.Info("shutting down Prometheus metrics server")
+
+		if err := server.Shutdown(context.Background()); err != nil {
+			setupLog.Error(err, "error shutting down the HTTP server")
+		}
+		close(idleConnsClosed)
+	}()
+
+	server.TLSConfig = s.getPrometheusTLSConfig(ctx, certWatcher)
+
+	if err := server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil && err != http.ErrServerClosed {
+		setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
+		return err
+	}
+
+	<-idleConnsClosed
 	return nil
 }
 
-func getWebhookServer(sspTLSOptions common.SSPTLSOptions) *webhook.Server {
-	// If TLSSecurityProfile is empty, we want to return nil so that the default
-	// webhook server configuration is used.
-	if sspTLSOptions.IsEmpty() {
-		return nil
+func (s *prometheusServer) getPrometheusTLSConfig(ctx context.Context, certWatcher *certwatcher.CertWatcher) *tls.Config {
+	return &tls.Config{
+		GetConfigForClient: func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
+			cfg := &tls.Config{}
+			cfg.GetCertificate = certWatcher.GetCertificate
+			return getConfigForClient(ctx, cfg, s.cache)
+		},
 	}
+}
 
-	tlsCfgFunc := func(cfg *tls.Config) {
-		cfg.CipherSuites = sspTLSOptions.CipherIDs(&setupLog)
-		setupLog.Info("Configured ciphers", "ciphers", cfg.CipherSuites)
+func newPrometheusServer(metricsAddr string, cache cache.Cache) *prometheusServer {
+	return &prometheusServer{
+		certPath:      path.Join(sdkTLSDir, sdkTLSCrt),
+		keyPath:       path.Join(sdkTLSDir, sdkTLSKey),
+		cache:         cache,
+		serverAddress: metricsAddr,
 	}
-
-	funcs := []func(*tls.Config){tlsCfgFunc}
-	return &webhook.Server{Port: webhookPort, TLSMinVersion: sspTLSOptions.MinTLSVersion, TLSOpts: funcs}
 }
 
 func main() {
@@ -127,6 +201,8 @@ func main() {
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
+	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
+	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
@@ -138,26 +214,24 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 
-	tlsOptions, err := common.GetSspTlsOptions(ctx)
-	if err != nil {
-		setupLog.Error(err, "Error while getting tls profile")
-		os.Exit(1)
+	var mgr ctrl.Manager
+
+	getTLSOptsFunc := func(cfg *tls.Config) {
+		cfg.GetConfigForClient = func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
+			return getConfigForClient(ctx, cfg, mgr.GetCache())
+		}
 	}
 
-	err = runPrometheusServer(metricsAddr, *tlsOptions)
-	if err != nil {
-		setupLog.Error(err, "unable to start prometheus server")
-		os.Exit(1)
-	}
-
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                 common.Scheme,
 		MetricsBindAddress:     "0",
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       leaderElectionID,
-		// If WebhookServer is set to nil, a default one will be created.
-		WebhookServer: getWebhookServer(*tlsOptions),
+		WebhookServer: &webhook.Server{
+			Port:    webhookPort,
+			TLSOpts: []func(*tls.Config){getTLSOptsFunc},
+		},
 	})
 
 	if err != nil {
@@ -171,10 +245,23 @@ func main() {
 			os.Exit(1)
 		}
 	}
+
+	metricsServer := newPrometheusServer(metricsAddr, mgr.GetCache())
+	if err != nil {
+		setupLog.Error(err, "unable create Prometheus server")
+		os.Exit(1)
+	}
+
+	if err := mgr.Add(metricsServer); err != nil {
+		setupLog.Error(err, "unable to set up metrics")
+		os.Exit(1)
+	}
+
 	if err := mgr.AddReadyzCheck("check", healthz.Ping); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
+
 	if err := mgr.AddHealthzCheck("health", healthz.Ping); err != nil {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
--- a/tests/crypto_policy_test.go
+++ b/tests/crypto_policy_test.go
@@ -7,14 +7,12 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	ocpv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/crypto"
-	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	ssp "kubevirt.io/ssp-operator/api/v1beta2"
@@ -106,19 +104,6 @@ var _ = Describe("Crypto Policy", func() {
 
 	AfterEach(func() {
 		strategy.RevertToOriginalSspCr()
-
-		// Because of bug[1], the SSP operator will move to CrashLoopBackOff state,
-		// so we need to wait until it is running.
-		// [1] - https://bugzilla.redhat.com/show_bug.cgi?id=2151248
-		Eventually(func(g Gomega) {
-			deployment := &apps.Deployment{}
-			err := apiClient.Get(ctx, client.ObjectKey{
-				Name:      strategy.GetSSPDeploymentName(),
-				Namespace: strategy.GetSSPDeploymentNameSpace(),
-			}, deployment)
-			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(deployment.Status.ReadyReplicas).To(BeNumerically(">=", 1))
-		}, env.Timeout(), time.Second).Should(Succeed())
 	})
 
 	Context("setting Crypto Policy", func() {


controllers/ssp_controller.go
```diff
@@ -239,9 +239,6 @@ func (r *sspReconciler) isRestartNeeded(sspObj *ssp.SSP) bool {
 	if reflect.DeepEqual(r.lastSspSpec, ssp.SSPSpec{}) {
```
0xFelix, line 239: This func is never returning `true`, so it can likely be dropped.

main.go
```diff
@@ -138,26 +180,63 @@ func main() {
...
-	err = runPrometheusServer(metricsAddr, *tlsOptions)
+	tlsOptions, err := common.GetSspTlsOptions(ctx)
 	if err != nil {
-		setupLog.Error(err, "unable to start prometheus server")
+		setupLog.Error(err, "Error while getting tls Profile")
 		os.Exit(1)
 	}
 
+	var webhookServer *webhook.Server
+	if tlsOptions.IsEmpty() {
```
0xFelix, line 223: Not sure we can do this. I don't think this allows to change the config later without restart?

tests/crypto_policy_test.go
```diff
@@ -65,7 +64,7 @@ var _ = Describe("Crypto Policy", func() {
 				},
 			},
 		}
-
+		// is not supported anymore
```
0xFelix, line 67: What does it mean?

tests/crypto_policy_test.go
```diff
@@ -106,19 +105,13 @@ var _ = Describe("Crypto Policy", func() {
...
-			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(deployment.Status.ReadyReplicas).To(BeNumerically(">=", 1))
-		}, env.Timeout(), time.Second).Should(Succeed())
+		deployment := &apps.Deployment{}
+		err := apiClient.Get(ctx, client.ObjectKey{
+			Name:      strategy.GetSSPDeploymentName(),
+			Namespace: strategy.GetSSPDeploymentNameSpace(),
+		}, deployment)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(deployment.Status.ReadyReplicas).To(BeNumerically(">=", 1))
```
0xFelix, line 114: Ready replicas is dangerous, might better use `UpdatedReplicas`. @jcanocan

main.go
```diff
@@ -66,28 +68,43 @@ const (
...
-		TLSConfig: &tlsConfig,
+		Addr:    metricsAddr,
+		Handler: mux,
+		TLSConfig: &tls.Config{
+			GetConfigForClient: func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
+				cfg := &tls.Config{}
+				tlsOptions, err := common.GetSspTlsOptions(ctx)
+				if err != nil {
+					setupLog.Error(err, "Error while getting tls profile")
+					os.Exit(1)
```
0xFelix, line 88: Do not exit here, better return an error.

codingben:
Hi, if this PR is still under work, can you please convert it to Draft? so CI won't run.

main.go
```diff
@@ -66,28 +68,43 @@ const (
...
+				}
+
+				if tlsOptions.IsEmpty() {
+					// if ciphersSuites is nil then default ciphers will be used
+					cfg.CipherSuites = nil
+					return cfg, nil
+				}
+				cfg.CipherSuites = tlsOptions.CipherIDs(&setupLog)
+				cfg.MinVersion, _ = minTLSVersionId(tlsOptions.MinTLSVersion)
+				certificate, err := loadCertificates(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
```
0xFelix, line 98: Not sure how expensive it is to load certs on every request. You might cache some of this.

0xFelix (commented):
In general this goes into the right direction. The `GetConfigForClient` functions still need to be deduplicated, improved and should not be lambda functions.

main.go
```diff
@@ -99,20 +116,45 @@ func runPrometheusServer(metricsAddr string, tlsOptions common.SSPTLSOptions) er
 	return nil
 }
 
-func getWebhookServer(sspTLSOptions common.SSPTLSOptions) *webhook.Server {
-	// If TLSSecurityProfile is empty, we want to return nil so that the default
-	// webhook server configuration is used.
-	if sspTLSOptions.IsEmpty() {
-		return nil
+// vm-console-proxy function
```
codingben, line 119: Can you please add a link to that function? e.g. so I can compare the changes here and there, if there any changes.

main.go
```diff
@@ -99,20 +116,45 @@ func runPrometheusServer(metricsAddr string, tlsOptions common.SSPTLSOptions) er
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load leaf certificate: %v\n", err)
+	}
+	crt.Leaf = leaf[0]
+	return &crt, nil
+}
 
-	funcs := []func(*tls.Config){tlsCfgFunc}
-	return &webhook.Server{Port: webhookPort, TLSMinVersion: sspTLSOptions.MinTLSVersion, TLSOpts: funcs}
+// crypto_policy.go function
```
codingben, line 142: Can you please add a link to that function, in that file?

main.go
```diff
@@ -66,28 +68,43 @@ const (
...
-	}
-
 	server := http.Server{
-		Addr:      metricsAddr,
-		Handler:   mux,
-		TLSConfig: &tlsConfig,
+		Addr:    metricsAddr,
+		Handler: mux,
+		TLSConfig: &tls.Config{
+			GetConfigForClient: func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
```
codingben, line 83: Can you please extract it to a new function, to keep `runPrometheusServer()` short?

main.go
```diff
@@ -138,26 +180,63 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 
-	tlsOptions, err := common.GetSspTlsOptions(ctx)
+	tlsCfgFunc := func(cfg *tls.Config) {
```
codingben, line 183: Is it possible to keep that function outside of `main()`? I think it's a good practice to keep `main()` short function, it's easier to read the code that way.

codingben:
> Hi, if this PR is still under work, can you please convert it to Draft? so CI won't run.

Also I'd suggest to have a better commit title, and update the PR description to be the same as commit description (so you won't forget to sync them later).

Now you have:
```
fix: SSP operator moving to CrashLoopBackOff after tlsProfile change

This commit adds the usage of callbacks to Prometheus and
Webhook server to fetch TLS config on every request.

```
I'd suggest to give it a name according to "This commit will fix...". For example:
```
fix: CrashLoopBackOff once tlsProfile changed

Add usage of callbacks to Prometheus and webhook
server to fetch TLS config on each client request.

```

controllers/ssp_controller.go
akrejcir: When the operator starts, `r.lastSspSpec` is a zero object, which in this case is `ssp.SSPSpec{}`. It is set in method `r.clearCacheIfNeeded()`, which is called after this `r.isRestartNeeded()`:

https://github.com/kubevirt/ssp-operator/blob/9759ddfe94f066df6cb0d97e727bf06342b6d7bd/controllers/ssp_controller.go#L151-L152

main.go
akrejcir: Agreed.
There is already a certificate watcher in the `controller-runtime` library. Maybe we can use that, I will check...

akrejcir: Here it is: https://github.com/kubevirt/ssp-operator/blob/9d3869accf1c183cdf66b3fdf572a3d8e9ec5a2d/vendor/sigs.k8s.io/controller-runtime/pkg/certwatcher/certwatcher.go#L45-L44

You can simply create a new `CertWatcher` and then add it to the `manager` to start it. Or call `Start()` manually in a separate goroutine.

tests/crypto_policy_test.go
jcanocan: You can check both just to be sure.

controllers/ssp_controller.go
akrejcir: _I've removed my comment here, because it was irrelevant._

main.go
```diff
@@ -66,53 +68,115 @@ const (
...
+			if err != nil {
+				return nil, err
+			}
+
+			if tlsOptions.IsEmpty() {
+				cfg.CipherSuites = nil
+				return cfg, nil
+			}
+			cfg.CipherSuites = tlsOptions.CipherIDs(&setupLog)
+			setupLog.Info("Configured ciphers", "ciphers", cfg.CipherSuites)
```
0xFelix, line 109: IMO we don't need to log this on every request.

main.go
```diff
@@ -66,53 +70,112 @@ const (
...
+			}
+
+			cipherNames, minTypedTLSVersion := ocpconfigv1.TLSProfiles[controllers.TLSProfile.Type].Ciphers, ocpconfigv1.TLSProfiles[controllers.TLSProfile.Type].MinTLSVersion
+			cfg.CipherSuites = common.CipherIDs(cipherNames)
+			cfg.MinVersion, err = crypto.TLSVersion(string(minTypedTLSVersion))
+			if err != nil {
+				return nil, err
+			}
+			return cfg, nil
+		},
```
jcanocan, line 105: Can we extract this to an external function? It is duplicated on `getTLSConfigFunc()`.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -192,8 +193,8 @@ func (ti *TLSInfo) CreateTlsConfig() *tls.Config {
 	}
 
 	if !ti.sspTLSOptions.IsEmpty() {
-		tlsConfig.CipherSuites = ti.sspTLSOptions.CipherIDs(nil)
-		minVersion, err := ti.sspTLSOptions.MinTLSVersionId()
+		tlsConfig.CipherSuites = common.CipherIDs(ti.sspTLSOptions.OpenSSLCipherNames)
```
0xFelix, line 196: Is there a lib for this too?

main.go
```diff
@@ -66,53 +70,112 @@ const (
...
+			cfg.GetCertificate = certWatcher.GetCertificate
+
+			if controllers.TLSProfile == nil {
+				cfg.MinVersion = crypto.DefaultTLSVersion()
+				cfg.CipherSuites = nil
+				return cfg, nil
+			}
+
+			if controllers.TLSProfile.Type == ocpconfigv1.TLSProfileCustomType {
+				cfg.CipherSuites = common.CipherIDs(controllers.TLSProfile.Custom.Ciphers)
```
0xFelix, line 90: Can you dedup the calls to `common.CipherIDs` and `crypto.TLSVersion`? This block and the block below look very similar.

main.go
0xFelix: Indeed

internal/template-validator/tlsinfo/tlsinfo.go
opokornyy: There is this function, but the problem is that it causes panic for some cipher names and we just want  to log that the cipher is not supported  https://github.com/kubevirt/ssp-operator/blob/fc1d67c1b3ee193d02cab94149217525975c1f0c/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go#L233-L245

main.go
```diff
@@ -66,53 +70,93 @@ const (
 	webhookPort = 9443
 )
 
-func runPrometheusServer(metricsAddr string, tlsOptions common.SSPTLSOptions) error {
+func callbackFunc(cfg *tls.Config) (*tls.Config, error) {
```
codingben, line 73: `getConfigForClientCallback`

internal/common/crypto_policy.go
```diff
@@ -65,46 +44,17 @@ func GetKnownCipherId(IANACipherName string) (uint16, bool) {
...
 }
 
-func (s *SSPTLSOptions) CipherIDs(logger *logr.Logger) (cipherSuites []uint16) {
-	for _, cipherName := range crypto.OpenSSLToIANACipherSuites(s.OpenSSLCipherNames) {
+func CipherIDs(names []string) (cipherSuites []uint16) {
+	for _, cipherName := range crypto.OpenSSLToIANACipherSuites(names) {
 		if id, ok := GetKnownCipherId(cipherName); ok {
 			cipherSuites = append(cipherSuites, id)
 		} else {
-			if logger != nil {
```
0xFelix, line 73: Can we keep the logger?

main.go
```diff
@@ -66,53 +72,114 @@ const (
...
+// please be aware that the APIServer is using http keepalive so this is going to
+// be executed only after a while for fresh connections and not on existing ones
+func getConfigForClientCallback(cfg *tls.Config, apiClient client.Client, ctx context.Context) (*tls.Config, error) {
+	var err error
+	var tlsProfile *ocpconfigv1.TLSSecurityProfile
 
-	minTlsVersion, err := tlsOptions.MinTLSVersionId()
-	if err != nil {
-		return err
+	var sspList ssp.SSPList
```
0xFelix, line 82: Why did you revert back to an explicit API call per callback? Please keep a cached version / a shared informer of the SSP CR.

main.go
```diff
@@ -66,53 +72,114 @@ const (
...
+}
+
+type prometheusServer struct {
+	server   *http.Server
+	client   client.Client
+	certPath string
+	keyPath  string
+}
+
+func (s *prometheusServer) InjectClient(client client.Client) error {
```
0xFelix, line 122: Make the `client` field public instead?

main.go
```diff
@@ -66,53 +72,114 @@ const (
...
+	s.client = client
+	return nil
+}
+
+func (s *prometheusServer) NeedLeaderElection() bool {
+	return false
```
0xFelix, line 128: What's the point in always returning false?

main.go
```diff
@@ -66,53 +72,114 @@ const (
...
+	}()
+
+	metricsServer.server = &http.Server{
+		Addr:    metricsAddr,
+		Handler: mux,
 	}
 
-	funcs := []func(*tls.Config){tlsCfgFunc}
-	return &webhook.Server{Port: webhookPort, TLSMinVersion: sspTLSOptions.MinTLSVersion, TLSOpts: funcs}
+	metricsServer.server.TLSConfig = metricsServer.getPrometheusTLSConfig(ctx, certWatcher)
```
0xFelix, line 180: Put this in struct initialization above?

main.go
```diff
@@ -138,26 +205,26 @@ func main() {
...
-	}
+	var mgr ctrl.Manager
 
-	err = runPrometheusServer(metricsAddr, *tlsOptions)
-	if err != nil {
-		setupLog.Error(err, "unable to start prometheus server")
-		os.Exit(1)
+	getTLSOptsFunc := func(cfg *tls.Config) {
+		cfg.GetConfigForClient = func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
+			// This is a hack, because webhook.Server does not implement InjectClient, so
```
0xFelix, line 212: Please avoid this

main.go
opokornyy: This is the client that is passed from manager and it should be using cache.

akrejcir: It was my idea. The `apiClient` uses a cache internally. It creates a shared informer for any requested resource. So calling `Get` and `List` methods on the client does not make an API call.

main.go
akrejcir: The `controller-runtime` `Manager` uses this interface to inject client to any `Runnable`, when the client is created.

main.go
akrejcir: This interface is used by the manager to check if the `Runnable` should run only when leader election succedes: 
https://github.com/kubevirt/ssp-operator/blob/862a7efa8fb1bb3e1f20494996b45c7bb1479773/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go#L339-L344

In case of metrics server, we want it to start for all the pods, even ones that fail leader election.

opokornyy: It indicates that this server does not need leader election. As far as I know, it is required by the manager. Maybe @akrejcir can elaborate more.

main.go
akrejcir: What do you recommend instead?
This is the easiest solution.

main.go
0xFelix: ACK

main.go
0xFelix: Might add a comment about this interface?

main.go
0xFelix: Might add a comment about this interface?

main.go
0xFelix: ACK

main.go
0xFelix: As discussed today, can you move this based on deprecation of the `InjectClient` interface to use `mgr.GetClient()` directly instead?

main.go
0xFelix: Can you check if this interface is also deprecated?

main.go
0xFelix: The term hack drove me away initially, but after looking at this again this seems to be an elegant solution to me, so let's do it like this.

akrejcir: Ok, @opokornyy can you remove this comment, or rewrite it?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -192,8 +193,8 @@ func (ti *TLSInfo) CreateTlsConfig() *tls.Config {
 	}
 
 	if !ti.sspTLSOptions.IsEmpty() {
-		tlsConfig.CipherSuites = ti.sspTLSOptions.CipherIDs(nil)
-		minVersion, err := ti.sspTLSOptions.MinTLSVersionId()
+		tlsConfig.CipherSuites = common.CipherIDs(ti.sspTLSOptions.OpenSSLCipherNames, &logger.Log)
```
akrejcir, line 196: Please keep the `nil` log here, because the change is not part of this PR. If you want, you can open a new PR to add the log.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -192,8 +193,8 @@ func (ti *TLSInfo) CreateTlsConfig() *tls.Config {
...
+		minVersion, err := crypto.TLSVersion(ti.sspTLSOptions.MinTLSVersion)
```
akrejcir, line 197: This is not correct, because `crypto.TLSVersion()` function expects one of these names:
https://github.com/kubevirt/ssp-operator/blob/9e99e78cceda599b72055610d9ddb9575af95094/vendor/github.com/openshift/library-go/pkg/crypto/crypto.go#L37-L42

So a string like:  `VersionTLS12`, but `ti.sspTLSOptions.MinTLSVersion` is a string returned by this function:
https://github.com/kubevirt/ssp-operator/blob/11c2e890e986955b25ce7fd43790c4416e9c4bbb/internal/common/crypto_policy.go#L119-L134

For example: `1.2`.

main.go
```diff
@@ -28,8 +29,13 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	ocpconfigv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	ssp "kubevirt.io/ssp-operator/api/v1beta2"
```
akrejcir, line 35: Please move local imports to the block below.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
-	setupLog.Info("Starting Prometheus metrics endpoint server with TLS")
-	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
-	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
-	handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{})
-	mux := http.NewServeMux()
-	mux.Handle("/metrics", handler)
+// This callback executes on each client call returning a new config to be used
+// please be aware that the APIServer is using http keepalive so this is going to
+// be executed only after a while for fresh connections and not on existing ones
+func getConfigForClientCallback(cfg *tls.Config, apiClient client.Client, ctx context.Context) (*tls.Config, error) {
```
akrejcir, line 78: Three comments for this function:
- After thinking about this more, I think it is better to have a parameter type `cache.Cache` instead of `client.Client`. It has the same reading methods as `Client`, but cannot be used for writing.
- Usually the `context.Context` parameter is the first.
- Please don't include the word `Callback` in a function name. In my opinion, it's better to name a function, by what it does, then by how it is used.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+	var err error
```
akrejcir, line 79: This line can be removed, because `err` is redefined below.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+	var tlsProfile *ocpconfigv1.TLSSecurityProfile
```
akrejcir, line 80: Please move this varaible definition close to it's usage. It makes the code easier to read.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
 
-	minTlsVersion, err := tlsOptions.MinTLSVersionId()
-	if err != nil {
-		return err
+	var sspList ssp.SSPList
+	if err := apiClient.List(ctx, &sspList, &client.ListOptions{}); err != nil {
```
akrejcir, line 83: The last argument is not needed, and it can be omitted:
```go
apiClient.List(ctx, &sspList)
```

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
-		Handler:   mux,
-		TLSConfig: &tlsConfig,
+	if tlsProfile == nil {
+		cfg.MinVersion = crypto.DefaultTLSVersion()
+		cfg.CipherSuites = nil
+		return cfg, nil
 	}
 
+	if tlsProfile.Type == ocpconfigv1.TLSProfileCustomType {
+		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &setupLog)
```
akrejcir, line 98: Passing `setupLog` here can cause confusing log messages, because this function is not called during setup.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+		cfg.MinVersion, err = crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
+		if err != nil {
+			return nil, err
+		}
+		return cfg, nil
+	}
+
+	cipherNames, minTypedTLSVersion := ocpconfigv1.TLSProfiles[tlsProfile.Type].Ciphers, ocpconfigv1.TLSProfiles[tlsProfile.Type].MinTLSVersion
```
akrejcir, line 106: Can you split this assignment into 2 lines?

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+		cfg.MinVersion = crypto.DefaultTLSVersion()
+		cfg.CipherSuites = nil
+		return cfg, nil
 	}
 
+	if tlsProfile.Type == ocpconfigv1.TLSProfileCustomType {
+		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &setupLog)
+		cfg.MinVersion, err = crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
+		if err != nil {
+			return nil, err
```
akrejcir, line 101: When this error is returned, the `cfg` parameter has already been modified. Can you change the code, so that on error, the `cfg` parameter is kept unchanged?

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+		}
+		return cfg, nil
+	}
+
+	cipherNames, minTypedTLSVersion := ocpconfigv1.TLSProfiles[tlsProfile.Type].Ciphers, ocpconfigv1.TLSProfiles[tlsProfile.Type].MinTLSVersion
+	cfg.CipherSuites = common.CipherIDs(cipherNames, &setupLog)
+	cfg.MinVersion, err = crypto.TLSVersion(string(minTypedTLSVersion))
+	if err != nil {
+		return nil, err
```
akrejcir, line 110: Similarly here, please keep the `cfg` unchanged on error.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+	return &tls.Config{
+		GetConfigForClient: func(_ *tls.ClientHelloInfo) (*tls.Config, error) {
+			cfg := &tls.Config{}
+			cfg.GetCertificate = certWatcher.GetCertificate
+			return getConfigForClientCallback(cfg, s.client, ctx)
+		},
+	}
+}
+
+func getPrometheusServer(metricsAddr string, ctx context.Context) (*prometheusServer, error) {
```
akrejcir, line 155: Can you rename this function to `newPrometheusServer()`? Because it is a constructor.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+	setupLog.Info("Starting Prometheus metrics endpoint server with TLS")
+	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
+	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
+	handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{})
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", handler)
```
akrejcir, line 161: Can you move this logic to `prometheusServer.Start()` method? It makes sense to call it when the server is started, not when it's created.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
 	}
 
-	tlsCfgFunc := func(cfg *tls.Config) {
-		cfg.CipherSuites = sspTLSOptions.CipherIDs(&setupLog)
-		setupLog.Info("Configured ciphers", "ciphers", cfg.CipherSuites)
+	go func() {
+		if err := certWatcher.Start(ctx); err != nil {
+			setupLog.Error(err, "certificate watcher error")
+		}
+	}()
```
akrejcir, line 177: Similarly, can you move the `certWatcher` creation and start to the `prometheusServer.Start()` method?

akrejcir (commented):
A partial review. I will look at it more later.

main.go
opokornyy: This interface should not be deprecated.

main.go
```diff
@@ -66,53 +72,115 @@ const (
...
+
+	if len(sspList.Items) != 0 {
+		tlsProfile = sspList.Items[0].Spec.TLSSecurityProfile
+	}
+
+	if tlsProfile == nil {
+		cfg.MinVersion = crypto.DefaultTLSVersion()
+		cfg.CipherSuites = nil
+		return cfg, nil
+	}
```
0xFelix, line 95:
```suggestion
	if len(sspList.Items) == 0 {
		cfg.MinVersion = crypto.DefaultTLSVersion()
		cfg.CipherSuites = nil
		return cfg, nil
	}
	
	tlsProfile := sspList.Items[0].Spec.TLSSecurityProfile
```

0xFelix: Do it like this?

main.go
```diff
@@ -66,53 +72,115 @@ const (
...
+		minVersion, err := crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
+		if err != nil {
+			return nil, err
+		}
+		cfg.MinVersion = minVersion
+		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &ctrl.Log)
+		return cfg, nil
+	}
+
+	cipherNames := ocpconfigv1.TLSProfiles[tlsProfile.Type].Ciphers
```
0xFelix, line 107: Maybe inline this?

main.go
0xFelix: This too?

main.go
```diff
@@ -66,53 +72,115 @@ const (
...
+	cfg.CipherSuites = common.CipherIDs(cipherNames, &ctrl.Log)
+
+	minTypedTLSVersion := ocpconfigv1.TLSProfiles[tlsProfile.Type].MinTLSVersion
```
0xFelix, line 110: This too?

main.go
```diff
@@ -66,53 +72,115 @@ const (
...
+func (s *prometheusServer) Start(ctx context.Context) error {
 	setupLog.Info("Starting Prometheus metrics endpoint server with TLS")
 	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
 	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
 	handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{})
 	mux := http.NewServeMux()
 	mux.Handle("/metrics", handler)
 
-	minTlsVersion, err := tlsOptions.MinTLSVersionId()
+	s.server.Handler = mux
```
0xFelix, line 141: Merge it with L138?

tests/crypto_policy_test.go
```diff
@@ -106,19 +105,14 @@ var _ = Describe("Crypto Policy", func() {
...
```
0xFelix, line 114: This should assert that we have the amount of ready and updated replicaes specified in `spec.Replicas`. Not just `>= 1`.

0xFelix (commented):
Added some comments

jcanocan:
@0xFelix @akrejcir @opokornyy I have two questions:

1. As it is now, as far as I was able to test, the crypto suite is not reverting TLS profiles to its original state. Should we fix this as part of this PR or as a follow-up? Could @opokornyy confirm if it's still happening with your code?
2. IIUC, when this PR lands, TLS configuration changes do not require restarting pods or even to update the pods. Am I right? I'm raising this because in case pods do not need to be updated, it's very likely that line `Expect(deployment.Status.UpdatedReplicas).To(BeNumerically(">=", 1))` will never be satisfied.

Thanks! :)

main.go
opokornyy: We still need to check if `spec.TLSSecurityProfile` is nil or not, so we would need to do it like this

```
if len(sspList.Items) == 0 || sspList.Items[0].Spec.TLSSecurityProfile == nil {
  cfg.MinVersion = crypto.DefaultTLSVersion()
  cfg.CipherSuites = nil
  return cfg, nil
}

tlsProfile := sspList.Items[0].Spec.TLSSecurityProfile
```

main.go
opokornyy: It is not possible to call `s.server.Handler.Handle()`  function, so we probably can not do it like that

0xFelix: ACK

main.go
```diff
@@ -66,53 +72,108 @@ const (
...
+	}
+
+	if len(sspList.Items) == 0 || sspList.Items[0].Spec.TLSSecurityProfile == nil {
+		cfg.MinVersion = crypto.DefaultTLSVersion()
+		cfg.CipherSuites = nil
+		return cfg, nil
+	}
+
+	tlsProfile := sspList.Items[0].Spec.TLSSecurityProfile
+
```
0xFelix, line 92: nit: newline

tests/crypto_policy_test.go
```diff
@@ -106,19 +105,13 @@ var _ = Describe("Crypto Policy", func() {
...
-		Eventually(func(g Gomega) {
-			deployment := &apps.Deployment{}
-			err := apiClient.Get(ctx, client.ObjectKey{
-				Name:      strategy.GetSSPDeploymentName(),
-				Namespace: strategy.GetSSPDeploymentNameSpace(),
-			}, deployment)
-			g.Expect(err).ToNot(HaveOccurred())
-			g.Expect(deployment.Status.ReadyReplicas).To(BeNumerically(">=", 1))
-		}, env.Timeout(), time.Second).Should(Succeed())
+		deployment := &apps.Deployment{}
```
0xFelix, line 108: Do we still need `Eventually` here, or does `strategy.RevertToOriginalSspCr` wait?

0xFelix (commented):
One more question and one nit.

@jcanocan 
1. IMO this should be fixed in a follow up PR. It has nothing to do with SSP restarting.
2. Good point!

tests/crypto_policy_test.go
```diff
@@ -106,19 +105,13 @@ var _ = Describe("Crypto Policy", func() {
...
+		err := apiClient.Get(ctx, client.ObjectKey{
+			Name:      strategy.GetSSPDeploymentName(),
+			Namespace: strategy.GetSSPDeploymentNameSpace(),
+		}, deployment)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(deployment.Status.ReadyReplicas).To(Equal(*deployment.Spec.Replicas))
```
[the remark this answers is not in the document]
opokornyy: IMO we can drop it

main.go
```diff
@@ -66,53 +72,107 @@ const (
...
+	keyPath  string
+}
+
+// NeedLeaderElection implements the LeaderElectionRunnable interface, which indicates
+// the prometheus server doesn't need leader election.
+func (s *prometheusServer) NeedLeaderElection() bool {
+	return false
+}
+
+func (s *prometheusServer) Start(ctx context.Context) error {
```
akrejcir, line 125: The `Start()` method needs to block until the context is closed or an error occurs:

https://github.com/kubevirt/ssp-operator/blob/862a7efa8fb1bb3e1f20494996b45c7bb1479773/vendor/sigs.k8s.io/controller-runtime/pkg/manager/manager.go#L319-L327

main.go
```diff
@@ -66,53 +72,107 @@ const (
...
 	setupLog.Info("Starting Prometheus metrics endpoint server with TLS")
 	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
 	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
```
akrejcir, line 128: Can you move these 2 global operations outside of this function? Maybe to `main()` or `init()`?

main.go
```diff
@@ -66,53 +72,107 @@ const (
...
 	handler := promhttp.HandlerFor(metrics.Registry, promhttp.HandlerOpts{})
 	mux := http.NewServeMux()
 	mux.Handle("/metrics", handler)
 
-	minTlsVersion, err := tlsOptions.MinTLSVersionId()
+	s.server.Handler = mux
```
akrejcir, line 133: The `s.server` could be a local variable now, because it is only used in this method.

main.go
```diff
@@ -66,53 +72,107 @@ const (
...
+	}()
 
-	server := http.Server{
-		Addr:      metricsAddr,
-		Handler:   mux,
-		TLSConfig: &tlsConfig,
-	}
+	s.server.TLSConfig = s.getPrometheusTLSConfig(ctx, certWatcher)
 
 	go func() {
```
akrejcir, line 148: No need to call this in a goroutine, because `Start()` method needs to block.

main.go
```diff
@@ -66,53 +72,107 @@ const (
...
-		err := server.ListenAndServeTLS(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
+		err := s.server.ListenAndServeTLS(s.certPath, s.keyPath)
```
akrejcir, line 149: The server needs to be stopped when `ctx` is closed.
Can you add a new gorotine that would wait until the context is closed and then calls `server.Shutdown()` , or maybe also `server.Close()` if `Shutdown()` does not finish in a certain time.

opokornyy: Done

main.go
```diff
@@ -28,16 +29,22 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
+	ssp "kubevirt.io/ssp-operator/api/v1beta2"
 	"kubevirt.io/ssp-operator/controllers"
 	"kubevirt.io/ssp-operator/internal/common"
 	common_templates "kubevirt.io/ssp-operator/internal/operands/common-templates"
+	"kubevirt.io/ssp-operator/internal/template-validator/logger"
```
akrejcir, line 47: This is not the logger that we want to use.

main.go
```diff
@@ -66,53 +73,116 @@ const (
...
+		return err
 	}
 
-	server := http.Server{
-		Addr:      metricsAddr,
-		Handler:   mux,
-		TLSConfig: &tlsConfig,
-	}
+	go func() {
+		if err := certWatcher.Start(ctx); err != nil {
```
akrejcir, line 143: Ideally we should use a different context for the `certWatcher` than `ctx`, so that it can be closed when this function returns on error.

We can do it in a follow-up PR.

main.go
```diff
@@ -66,53 +73,116 @@ const (
...
+		}
+	}()
 
+	idleConnsClosed := make(chan struct{})
 	go func() {
-		err := server.ListenAndServeTLS(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
-		if err != nil {
-			setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
+		<-ctx.Done()
+		logger.Log.Info("shutting down Prometheus metrics server")
```
akrejcir, line 151: You can use `setupLog` for now, but later we should probably change it, because this is not setup anymore.

main.go
```diff
@@ -66,53 +73,116 @@ const (
...
+
+		if err := server.Shutdown(context.Background()); err != nil {
+			logger.Log.Error(err, "error shutting down the HTTP server")
```
akrejcir, line 154: Please use `setupLog`.

main.go
```diff
@@ -66,53 +73,116 @@ const (
...
 		}
+		close(idleConnsClosed)
 	}()
+
+	server.TLSConfig = s.getPrometheusTLSConfig(ctx, certWatcher)
+
+	if err := server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil {
+		setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
```
akrejcir, line 162: The error should be returned from this function.

main.go
```diff
@@ -124,6 +194,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	metrics.Registry.MustRegister(common_templates.CommonTemplatesRestored)
+	metrics.Registry.MustRegister(common.SSPOperatorReconcileSucceeded)
```
akrejcir, line 198: Please move these lines lower. Now they are in the middle of flag parsing logic.

main.go
akrejcir: The documentation for `ListenAndServeTLS()` says that it always returns a non-nil error:
https://github.com/golang/go/blob/0ae54ddd37302bdd2a8c775135bf5f076a18eeb3/src/net/http/server.go#L3281-L3283

https://github.com/golang/go/blob/0ae54ddd37302bdd2a8c775135bf5f076a18eeb3/src/net/http/server.go#L3026-L3028

These links are for go `1.19` but the same comments are in `master`.

main.go
```diff
@@ -66,53 +72,117 @@ const (
...
+			setupLog.Error(err, "certificate watcher error")
+		}
+	}()
 
+	idleConnsClosed := make(chan struct{})
 	go func() {
-		err := server.ListenAndServeTLS(path.Join(sdkTLSDir, sdkTLSCrt), path.Join(sdkTLSDir, sdkTLSKey))
-		if err != nil {
-			setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
+		<-ctx.Done()
```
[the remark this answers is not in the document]
akrejcir: I've reconsidered this.
In this PR, we are already letting the above goroutine leak, from the `Start()` function, so this one can be left as is.

Can you add a `// TODO` comment to the code next to both goroutines, and open a new github issue, so that we don't forget to open a followup PR?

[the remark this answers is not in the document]
opokornyy: Done

PR #641, opened 2023-08-01
refactor: cleanup tekton operands

### LINKED ISSUE #548: fix: Issues left for next PR from #532
This issue is to track fixing multiple comments from #532, that were left for a future PR.

/cc @codingben

codingben:

Cleanup Tekton operands to improve code quality
and readability.

Fixes #548

**Release note**:

```release-note
None
```


internal/operands/tekton-tasks/reconcile.go
```diff
@@ -307,3 +276,21 @@ func getTektonTasksNamespace(request *common.Request) string {
 	}
 	return request.Instance.Namespace
 }
+
+func filterResources[PT interface {
```
0xFelix, line 280: Appreciate use of generics! Can this be simplified to just use the `metav1.Object` interface? It is working right now but looks very complicated.

codingben: Yes, I agree that's very complicated, also there is no Go docs for it, but it works. I tried to do it with just `metav1.Object`, I couldn't make it work. @akrejcir suggested to do that way yesterday https://github.com/kubevirt/ssp-operator/pull/532#discussion_r1191019175.

codingben: @akrejcir's yesterday's explanation for it:
```
If you would define simply [T metav1.Object] it means that parameter T is a type that implements interface metav1.Object . Which means it is a pointer to a struct, for example T = *v1.Pod.
When you have : []T it is in fact []*v1.Pod. So when a method accepts []T parameter, it does not accept []v1.Pod to it.
The PT interface {*T ; meta1.Object } is a trick. It means that type PT satisfies an anonymous typeset defined directly in the function signature. This typeset interface { *T; metav1.Object } means that it is a pointer to T and implements interface metav1.Object.  Which means that for example PT = *v1.Pod .
It would be nice if golang compiler was smarter and we didn't need PT. :)
```

akrejcir: Better explanation: https://stackoverflow.com/a/71444968

internal/operands/tekton-tasks/reconcile.go
```diff
@@ -307,3 +276,21 @@ func getTektonTasksNamespace(request *common.Request) string {
...
+	*T
+	metav1.Object
+}, T any](items []T, trimSuffix bool) []T {
+	var results []T
+	for i := range items {
+		object := PT(&items[i])
```
0xFelix, line 286: Isn't this an unchecked cast by the way?

internal/operands/tekton-tasks/reconcile.go
0xFelix: IMO we should not merge this just because it works.

I see the pain in changing the slices to hold pointers instead. Is there another maybe less 'generic' but still prettier solution?

lyarwood: What about something casting to keep this simple?

```go
func filterResources[T any](items []T, trimSuffix bool) []T {
	var results []T
	for _, obj := range items {
		metaObj, _ := any(obj).(metav1.ObjectMeta)
		name := metaObj.GetName()
		if trimSuffix {
			name = strings.TrimSuffix(name, "-task")
		}
		if _, ok := AllowedTasks[name]; ok {
			results = append(results, obj)
		}
	}
	return results
}
```

codingben: Ah sweet, thanks @lyarwood. Let's use it then.

internal/operands/tekton-tasks/reconcile.go
codingben: I've used @lyarwood's piece of code that is simpler.

internal/operands/tekton-tasks/reconcile.go
```diff
@@ -307,3 +276,18 @@ func getTektonTasksNamespace(request *common.Request) string {
 	}
 	return request.Instance.Namespace
 }
+
+func filterResources[T any](items []T, trimSuffix bool) []T {
+	var results []T
+	for _, item := range items {
+		object, _ := any(item).(metav1.ObjectMeta)
```
0xFelix, line 283: This is still an unchecked cast. Please return early with empty list at least if that happens.

internal/operands/tekton-tasks/reconcile.go
akrejcir: Just to reply. This is not an unchecked cast. The yype system guarantees that it will always be correct.
`PT` is a `*T` , `items[i]` has type `T`, and `&items[i]` has type `*T`.

codingben:
> IMO we should not merge this just because it works.
> I see the pain in changing the slices to hold pointers instead. Is there another maybe less 'generic' but still prettier solution?

@0xFelix I didn't see another way, I spent too much time on it. It seems like this generics feature is kind of new, and there is not enough documentation IMO, all examples based on simple things like "int/float/string" whatever.

Simpler solution would be not to use generics at all:
```
func New(bundle *tektonbundle.Bundle) operands.Operand {
	bundle.Tasks = filterTasks(bundle.Tasks)
	bundle.ServiceAccounts = filterServiceAccounts(bundle.ServiceAccounts)
	bundle.RoleBindings = filterRoleBindings(bundle.RoleBindings)
	bundle.ClusterRoles = filterClusterRoles(bundle.ClusterRoles)

	return &tektonTasks{
		tasks:           bundle.Tasks,
		serviceAccounts: bundle.ServiceAccounts,
		roleBindings:    bundle.RoleBindings,
		clusterRoles:    bundle.ClusterRoles,
	}
}
```
And just use separated functions for it.

0xFelix:
I still appreciate the use of generics, but I'm afraid this is making it more complicated than it needs to be. Golang can be repetitive, but at least it is easy to grok what happens?

jcanocan:
I'm not particularly against the use of generics, thanks for experimenting generics and pushing forward the use of them. However, I agree with @0xFelix that it may make the code a bit more hard to read. Nevertheless, I guess that's something that a good function documentation can solve. My only concern is if we can do the same with the `DeepCopy` functions.

internal/operands/tekton-pipelines/reconcile.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 // +kubebuilder:rbac:groups=tekton.dev,resources=pipelines,verbs=list;watch;create;update;delete
-// +kubebuilder:rbac:groups=*,resources=configmaps,verbs=list;watch;create;delete
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=list;watch;create;delete
```
jcanocan, line 20: Just out of curiosity, is there any difference between using `*` or `core`? AFAIK, it will refer to the same `apiGroup`, isn't it?

codingben: Yes, `*` would grant permission to any API group in your cluster. Imagine that some custom API group, e.g. `kubevirt.io` would contain a custom ConfigMap resource that may be accessible (e.g. some ConfigMap that contains secrets).

I see that it's also possible to just specify an empty API group, as `""` (`groups=""`) and it would also refer to the core objects. Anyways, by using `core` we embrace least privilege principle and readability.

From the beginning I think that It was specified as `*` to speed up the development process (also some verbs) - because to have a specific value requires some effort.

codingben: I love the explanation of that topic here [1]. Very nice guidance.

[1] https://learnk8s.io/rbac-kubernetes#modelling-access-to-resources

codingben:
@akrejcir All comments in issue #548 were resolved. Please review if you have some free time.

internal/tekton-bundle/tekton-bundle.go
```diff
@@ -38,74 +36,34 @@ type Bundle struct {
...
-	}
-
-	return filesBytes, nil
-}
-
-func decodeObjectsFromFiles(files [][]byte) (*Bundle, error) {
-	bundle := &Bundle{}
-	for _, file := range files {
-		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(file), 1024)
+		defer file.Close()
```
akrejcir, line 65: An unintuitive behavior of golang is that `defer` statements are executed when returning from a function, and not when a scope ends with `}`, like for example a destructor in C++ would be executed. When you call `defer` in a loop, all the defers will be remembered and only called on return. In this case, all the files will be closed only on return.

codingben: Ack, thanks for the explanation. Good to know.

codingben: Done. I'll leave that discussion open so others can learn as well.

internal/tekton-bundle/tekton-bundle.go
```diff
@@ -38,138 +36,107 @@ type Bundle struct {
...
-		return nil, err
-	}
-
-	return tektonObjs, nil
-}
-
-func getPipelineBundlePath() string {
-	return tektonPipelinesBundleDir
+func GetTektonPipelineBundlePaths() []string {
+	var paths []string
```
0xFelix, line 40: Can you assemble this slice one time as a global? Usually I dislike globals, but for storing this kind of information it should be fine. It saves recreating the slice on every func call.

0xFelix (commented):
As @jcanocan, can you add some explanation for the generic function's type parameters?

internal/tekton-bundle/tekton-bundle.go
codingben: Okay done.

codingben:
> As @jcanocan, can you add some explanation for the generic function's type parameters?

Done, can you please review?

codingben:
Unit tests and functional tests passed, except the Tekton example pipelines jobs that are failing. @ksimon1 said he is aware of it and working on a solution. This issue also happens in common-templates according to him.

codingben:
> Unit tests and functional tests passed, except the Tekton example pipelines jobs that are failing. @ksimon1 said he is aware of it and working on a solution. This issue also happens in common-templates according to him.

Karel told me that [1] affects all CI jobs in all projects, including this PR. Let's see when we'll get a solution for it, until then a workaround probably would be to retry the failed jobs.

[1] https://github.com/kubevirt/containerized-data-importer/issues/2838

0xFelix (approved):
/approve

Thanks!

jcanocan:
> > As @jcanocan, can you add some explanation for the generic function's type parameters?
> 
> Done, can you please review?

IMHO the explanation is very detailed and precise. Many thanks.
/lgtm

ksimon1:
@codingben can you please rebase this PR?

codingben:
> @codingben can you please rebase this PR?

@ksimon1 Ack, doing it.

codingben:
> /cc sradco

I think it's wrong comment, I don't think we need more reviews in this PR.

codingben:
Hi @lyarwood @0xFelix, I see a few tests failed because of:
```
[0mCommon Instance Types [38;5;243moperand [38;5;9m[1m[It] should reconcile from URL when provided[0m
[38;5;243m/go/src/github.com/kubevirt/ssp-operator/tests/common_instancetypes_test.go:39[0m

  [38;5;9m[FAILED] Expected success, but got an error:
      <*errors.StatusError | 0xc00012e640>: 
      virtualmachineclusterpreferences.instancetype.kubevirt.io "centos.9.stream.desktop" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "virtualmachineclusterpreferences.instancetype.kubevirt.io \"centos.9.stream.desktop\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "centos.9.stream.desktop",
                  Group: "instancetype.kubevirt.io",
                  Kind: "virtualmachineclusterpreferences",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }[0m
  [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/go/src/github.com/kubevirt/ssp-operator/tests/common_instancetypes_test.go:58[0m [38;5;243m@ 10/02/23 11:01:11.283[0m
```
Can you please check? it seems like it wasn't found, I'm not familiar enough with InstanceTypes resources.

codingben:
It seems like there is no CentOS 9 folder in `v0.1.0`: https://github.com/kubevirt/common-instancetypes/tree/v0.1.0/preferences/centos/9_stream

0xFelix:
`centos.9.stream` was renamed to `centos.stream9` recently, but that should not be the issue. SSP should reconcile the preferences from the URL but it looks like the cluster was not in a clean state before running this test?

0xFelix:
The URL looks to be incorrect in general, because it is only referring to Preferences but not Instancetypes?

codingben:
Thanks for looking on it, the fact that this happened only in this PR, let's re-run the tests then?

lyarwood:
https://github.com/kubevirt/ssp-operator/pull/693 for the common-instancetypes failure, smells like a race to me.

internal/tekton-bundle/tekton-bundle.go
```diff
@@ -38,138 +36,109 @@ type Bundle struct {
...
+			case configMapKind:
+				cm := v1.ConfigMap{}
+				err = getObject(obj, &cm)
+				if err != nil {
+					return err
+				}
+				bundle.ConfigMaps = append(bundle.ConfigMaps, cm)
+			default:
+				continue
 			}
```
jcanocan, line 138: Can we move the error checking to the end of the switch statement?  Something like:

```go
switch kind {
   case tasksString:
     task := pipeline.Task{}
     err = getObject(obj, &task)
     bundle.Tasks = append(bundle.Tasks, task)
[...]
}
if err != nil {
  return err
}
```

codingben:
@ksimon1 Seems like there is an issue with example Tekton pipelines again?

internal/tekton-bundle/tekton-bundle.go
codingben: Thanks, I'll look on it. I'd love to merge this PR firstly because it's open too much time already and too many CI issues there.

codingben:
@akrejcir @ksimon1 @0xFelix Can you please approve and merge it?

PR #645, opened 2023-08-08
fix: Use the latest image of vm-console-proxy

akrejcir:
- Updated release script to change `vm-console-proxy` image tag.
- csv-generator uses `VM_CONSOLE_PROXY_IMAGE` environment variable to set the image.

**Release note**:
```release-note
csv-generator can be used to set the image of vm-console-proxy.
```

### DIFF
--- a/.github/workflows/release-vm-console-proxy.yaml
+++ b/.github/workflows/release-vm-console-proxy.yaml
@@ -17,6 +17,7 @@ jobs:
           OUTPUT_FILE=./data/vm-console-proxy-bundle/vm-console-proxy.yaml
           mkdir -p ./data/vm-console-proxy-bundle
           curl -L https://github.com/kubevirt/vm-console-proxy/releases/download/${RELEASE_VERSION}/vm-console-proxy.yaml > ${OUTPUT_FILE}
+          sed -i "s/defaultVmConsoleProxyImageTag = .*$/defaultVmConsoleProxyImageTag = \"${RELEASE_VERSION}\"/" ./internal/operands/vm-console-proxy/defaults.go 
 
       - name: Create pull request
         if: ${{ github.event.client_payload.release_version }} != ''
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -70,6 +70,17 @@ rules:
   - list
   - watch
 - apiGroups:
+  - apiregistration.k8s.io
+  resources:
+  - apiservices
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
   - apps
   resources:
   - deployments
@@ -187,13 +198,29 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - serviceaccounts
   verbs:
   - create
   - delete
+  - get
   - list
+  - patch
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
+  verbs:
+  - create
 - apiGroups:
   - ""
   resources:
@@ -224,14 +251,6 @@ rules:
   - delete
   - list
   - update
-  - watch
-- apiGroups:
-  - kubevirt.io
-  resources:
-  - virtualmachineinstances
-  verbs:
-  - get
-  - list
   - watch
 - apiGroups:
   - kubevirt.io
@@ -282,6 +301,18 @@ rules:
 - apiGroups:
   - rbac.authorization.k8s.io
   resources:
+  - clusterrolebindings
+  - clusterroles
+  - rolebindings
+  verbs:
+  - create
+  - delete
+  - list
+  - update
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
   - clusterroles
   - rolebindings
   verbs:
@@ -320,7 +351,9 @@ rules:
   verbs:
   - create
   - delete
+  - get
   - list
+  - patch
   - update
   - watch
 - apiGroups:
@@ -328,10 +361,8 @@ rules:
   resources:
   - routes
   verbs:
-  - create
   - delete
   - list
-  - update
   - watch
 - apiGroups:
   - ssp.kubevirt.io
@@ -353,6 +384,12 @@ rules:
   - ssps/status
   verbs:
   - update
+- apiGroups:
+  - subresources.kubevirt.io
+  resources:
+  - virtualmachineinstances/vnc
+  verbs:
+  - get
 - apiGroups:
   - subresources.kubevirt.io
   resources:
--- a/internal/operands/vm-console-proxy/reconcile.go
+++ b/internal/operands/vm-console-proxy/reconcile.go
@@ -8,8 +8,8 @@ import (
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	apiregv1 "k8s.io/kube-aggregator/pkg/apis/apiregistration/v1"
 	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/operands"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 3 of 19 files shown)


codingben (approved):
/lgtm

EDIT: Seems like this PR needs more work since a new vm-console-proxy v0.3.0 contains breaking changes.

akrejcir:
This needs more work.

akrejcir:
This is nearly complete, it's only missing one functional test.
Can you take a look?

/cc @0xFelix @lyarwood

config/rbac/role.yaml
0xFelix, line 1: Have you checked that these are the minimal required RBAC changes?

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -30,16 +30,25 @@ const (
...
+// +kubebuilder:rbac:groups=route.openshift.io,resources=routes,verbs=list;watch;delete
 
 // RBAC for created roles
-// +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachineinstances,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;create;update;delete;patch
+// +kubebuilder:rbac:groups=core,resources=serviceaccounts/token,verbs=create
+// +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachineinstances;virtualmachines,verbs=get;list;watch
+// +kubebuilder:rbac:groups=subresources.kubevirt.io,resources=virtualmachineinstances/vnc,verbs=get
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=get;list;watch;create;update;delete;patch
```
0xFelix, line 45: Same as above.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -289,9 +388,62 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
 			ExpectResourceNotExists(service, request)
 			ExpectResourceNotExists(deployment, request)
+			ExpectResourceNotExists(bundle.ApiService, request)
+		})
+
+		It("should delete Route leftover from a previous version on reconcile", func() {
+			route := &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      routeName,
+					Namespace: namespace,
```
0xFelix, line 402: Use `otherNamespace`?

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -289,9 +388,62 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			ExpectResourceNotExists(route, request)
+		})
+
+		It("should delete Route leftover from a previous version on cleanup", func() {
+			route := &routev1.Route{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      routeName,
+					Namespace: namespace,
```
0xFelix, line 428: Same as above.

Do these test cases add any value compared to the same ones above?

internal/vm-console-proxy-bundle/bundle.go
```diff
@@ -105,7 +114,7 @@ func loadBundleFromBytes(data []byte) (*Bundle, error) {
 }
 
 func validateBundle(bundle *Bundle) error {
-	missingFields := make([]string, 0, 6)
+	missingFields := make([]string, 0, 7)
```
0xFelix, line 117: If you want to allocate space for all checks below this should be `8`?

0xFelix (commented):
Good start, added some comments!

config/rbac/role.yaml
akrejcir: Yes, they are minimal in the sense of groups:
- read permissions: `get`, `list`, `watch`
- write permissions: `create`, `update`, `patch`, `delete`.

In my opinion it does not make sense to grant `list` permission and not `get` permission. Because `list` can be used to `get` a single element anyway. Similarly for `update` and `patch`.

internal/operands/vm-console-proxy/reconcile.go
akrejcir: The same.

internal/operands/vm-console-proxy/reconcile_test.go
akrejcir: Yes, this case tests the `Cleanup()` method, and the above one tests `Reconcile()` method.

internal/vm-console-proxy-bundle/bundle.go
akrejcir: Thanks, I missed it.

config/rbac/role.yaml
lyarwood: @jcanocan what was the logic you applied with your previous RBAC audits?

I tend to agree with @0xFelix here that we shouldn't be granting everything even if a `list` basically provides `get` etc but lets stay consistant with @jcanocan 's previous work here and maybe write down the resulting best practice somewhere.

lyarwood: Ah @jcanocan is out this week, I'm okay leaving this as is for now if we write up a follow issue for next week to maybe reduce this down and document the best practice.

akrejcir: If we want to be more precise, It's harder to know what permissions libraries require, without running and failing, or reading the library code.

lyarwood: Yeah appreciate that, definitely a follow up task now given @jcanocan isn't around anyway.

tests/vm_console_proxy_test.go
```diff
@@ -195,25 +213,28 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
 			Entry("[test_id:9883] deployment", &deploymentResource),
 			Entry("[test_id:9884] config map", &configMapResource),
-			Entry("[test_id:9885] route", &routeResource),
+			Entry("[test_id:TODO] API service", &apiServiceResource),
 		)
 	})
 
-	Context("Route to access proxy", func() {
+	// TODO -- rewrite this test vv
+	PContext("Route to access proxy", func() {
```
codingben, line 237: Will you rewrite this test in this PR?

.github/workflows/release-vm-console-proxy.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
           OUTPUT_FILE=./data/vm-console-proxy-bundle/vm-console-proxy.yaml
           mkdir -p ./data/vm-console-proxy-bundle
           curl -L https://github.com/kubevirt/vm-console-proxy/releases/download/${RELEASE_VERSION}/vm-console-proxy.yaml > ${OUTPUT_FILE}
+          sed -i "s/defaultVmConsoleProxyImageTag = .*$/defaultVmConsoleProxyImageTag = \"${RELEASE_VERSION}\"/" ./internal/operands/vm-console-proxy/defaults.go
```
codingben, line 20: IMHO, I don't think we should do that bump here.

1. This workflow is used to update vm-console-proxy bundle only
2. We try to change code during that workflow, what if `defaults.go` will be removed in the future?

Why not just use `latest` tag? if anyways this intended to be updated on every release to latest one, just use `latest` then: `quay.io/kubevirt/vm-console-proxy:latest`.

tests/vm_console_proxy_test.go
akrejcir: Yes, I will post it soon.

.github/workflows/release-vm-console-proxy.yaml
akrejcir: > 1. This workflow is used to update vm-console-proxy bundle only

This workflow is used to update SSP operator code, so that it uses the latest release of `vm-cosole-proxy`. It can update the image tag and bundle.

> 2. We try to change code during that workflow, what if `defaults.go` will be removed in the future?

I will add a comment to `defaults.go` explaining how the file is used.

> Why not just use `latest` tag? if anyways this intended to be updated on every release to latest one, just use `latest` then: `quay.io/kubevirt/vm-console-proxy:latest`.

There is no `latest` tag. We would need to add automation to create and update it.
https://quay.io/repository/kubevirt/vm-console-proxy?tab=tags&tag=latest

Using the latest tag would probably not save us work, because we would need to fix the version when creating a release branch.

internal/operands/vm-console-proxy/reconcile_test.go
akrejcir: done

internal/operands/vm-console-proxy/reconcile_test.go
akrejcir: done

tests/vm_console_proxy_test.go
akrejcir: posted

akrejcir:
I've added the missing test. It's ready for review.

codingben:
/lgtm

I don't have more comments.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -289,9 +388,62 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
 			ExpectResourceNotExists(bundle.ClusterRole, request)
 			ExpectResourceNotExists(bundle.ClusterRoleBinding, request)
+			ExpectResourceNotExists(bundle.RoleBinding, request)
 			ExpectResourceNotExists(configMap, request)
 			ExpectResourceNotExists(service, request)
 			ExpectResourceNotExists(deployment, request)
+			ExpectResourceNotExists(bundle.ApiService, request)
+		})
+
+		It("should delete Route leftover from a previous version on reconcile", func() {
```
0xFelix, line 398: Same, make this and test below a Table?

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -87,21 +118,75 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
 		ExpectResourceNotExists(bundle.ClusterRoleBinding, request)
+		ExpectResourceNotExists(bundle.RoleBinding, request)
 		ExpectResourceNotExists(bundle.ConfigMap, request)
 		ExpectResourceNotExists(bundle.Service, request)
 		ExpectResourceNotExists(bundle.Deployment, request)
-		ExpectResourceNotExists(newRoute(namespace, serviceName), request)
+		ExpectResourceNotExists(bundle.ApiService, request)
+	})
+
+	It("should delete Route leftover from a previous version on reconcile", func() {
```
0xFelix, line 140: This and the test below could be a table.

tests/vm_console_proxy_test.go
```diff
@@ -226,27 +246,87 @@ var _ = Describe("VM Console Proxy Operand", func() {
...
-				g.Expect(response.StatusCode).To(Equal(http.StatusUnauthorized))
+				request.Header.Set("Authorization", "Bearer "+saToken)
 
-				body, err := io.ReadAll(response.Body)
+				response, err := httpClient.Do(request)
 				g.Expect(err).ToNot(HaveOccurred())
+				defer func() { _ = response.Body.Close() }()
 
-				g.Expect(body).To(ContainSubstring("authenticating token cannot be empty"))
+				g.Expect(response.StatusCode).To(Equal(http.StatusUnauthorized))
```
0xFelix, line 329: 401 on successful access?

akrejcir: Yes, the test only checks if the `vm-console-proxy` can be accessed. But maybe if the token is incorrect the API server also returns 401. I will check.

internal/operands/vm-console-proxy/reconcile_test.go
akrejcir: done

tests/vm_console_proxy_test.go
akrejcir: I've changed the test.

0xFelix (commented):
Looks good to me, thanks!

Can you please create an issue to track the RBAC follow up?

/approve

akrejcir:
@ksimon1 , are the CI failures related to the CDI issue?

config/rbac/role.yaml
jcanocan: Hi all! The methodology used for auditing the rules is described here: https://github.com/kubevirt/ssp-operator/pull/616. Basically, I've removed all verbs from all apiGroups, and I've added requested verbs. @akrejcir is right that it's hard to know in advance which permissions are required in advance. We are limited to trial and error in this sense.

akrejcir:
We cannot merge this yet.
Incorrect implementation of `APIService` extension in `vm-cosole-proxy` is blocking the deletion of **any** namespace in the cluster.

Holding until a fixed version of `vm-console-proxy` is released.

/hold

akrejcir:
I've fixed the bug in `vm-cosole-proxy`: https://github.com/kubevirt/vm-console-proxy/pull/27
And released a new version.

/unhold

PR #604, opened 2023-06-28
chore: Update controller-runtime to v0.15.0

lyarwood:
**What this PR does / why we need it**:

https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

* Make `*http.Client` configurable and use/share the same client by
  default https://github.com/kubernetes-sigs/controller-runtime/pull/2122

* Remove dependency injection functions
  https://github.com/kubernetes-sigs/controller-runtime/pull/2134,
  https://github.com/kubernetes-sigs/controller-runtime/pull/2120

* Add context to EventHandler(s)
  https://github.com/kubernetes-sigs/controller-runtime/pull/2139

* `Validator` and `CustomValidator` interfaces have been modified to
  allow returning warnings
  https://github.com/kubernetes-sigs/controller-runtime/pull/2014

* operator-framework is also pinned to ecb9be48837 until a new release
  is cut supporting controller-runtime v0.15.0

**Which issue(s) this PR fixes**: 

Fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #653, opened 2023-08-11
crd-watch: Add test suite

lyarwood:
/hold
/cc @akrejcir 
/cc @0xFelix 

FWIW I noticed this working on https://github.com/kubevirt/ssp-operator/pull/604, looks like we haven't been running unit tests on this and possibly other areas of the code for a while. I'll likely end up fixing this in the other PR but wanted to highlight for now, any help checking for other missed tests would be appreciated! 

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/crd-watch/crd_watch_suite_test.go
+++ b/internal/crd-watch/crd_watch_suite_test.go
@@ -1,0 +1,13 @@
+package crd_watch_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCrdWatch(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "CrdWatch Suite")
+}


internal/crd-watch/crd_watch_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package crd_watch_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestCrdWatch(t *testing.T) {
```
akrejcir, line 10: Thanks for noticing!
Can you move this function to `crd-watch_test.go`, because there is only one test file in the folder?

lyarwood: /close

ACK will do, I'll likely fix this within https://github.com/kubevirt/ssp-operator/pull/604 now, was just worried at the time that there might be other examples in the codebase but I think we are okay. Closing this out.

PR #661, opened 2023-08-16
common-instancetypes: Ignore attempts to reconcile virt-operator owned objects from bundle

lyarwood:

This PR introduces `IgnoreObjectOwnedByVirtOperator` to `reconcileBuilder` that as the name suggests ignores attempts to reconcile resources that are labelled as being owned by KubeVirt's `virt-operator`.

This is then used by the `common-instancetypes` operand ahead of deployment of these resources being enabled through `virt-operator` by https://github.com/kubevirt/kubevirt/pull/10309.

Future changes will also introduce a configurable to disable `common-instancetypes` deployment within SSP and ultimately deprecate the functionality ahead of removal in a future release of SSP.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/operands/common-instancetypes/reconcile.go
+++ b/internal/operands/common-instancetypes/reconcile.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/kustomize/api/krusty"
@@ -15,6 +16,7 @@ import (
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 
+	virtv1 "kubevirt.io/api/core/v1"
 	instancetypeapi "kubevirt.io/api/instancetype"
 	instancetypev1beta1 "kubevirt.io/api/instancetype/v1beta1"
 	"kubevirt.io/ssp-operator/internal/common"
@@ -260,7 +262,11 @@ func (c *CommonInstancetypes) reconcileFromURL(request *common.Request) ([]commo
 	// Generate the normal set of reconcile funcs to create or update the provided resources
 	c.virtualMachineClusterInstancetypes = clusterInstancetypesFromURL
 	c.virtualMachineClusterPreferences = clusterPreferencesFromURL
-	return common.CollectResourceStatus(request, c.reconcileFuncs()...)
+	reconcileFuncs, err := c.reconcileFuncs(request)
+	if err != nil {
+		return nil, err
+	}
+	return common.CollectResourceStatus(request, reconcileFuncs...)
 }
 
 func (c *CommonInstancetypes) reconcileFromBundle(request *common.Request) ([]common.ReconcileResult, error) {
@@ -277,7 +283,11 @@ func (c *CommonInstancetypes) reconcileFromBundle(request *common.Request) ([]co
 
 	c.virtualMachineClusterInstancetypes = clusterInstancetypesFromBundle
 	c.virtualMachineClusterPreferences = clusterPreferencesFromBundle
-	return common.CollectResourceStatus(request, c.reconcileFuncs()...)
+	reconcileFuncs, err := c.reconcileFuncs(request)
+	if err != nil {
+		return nil, err
+	}
+	return common.CollectResourceStatus(request, reconcileFuncs...)
 }
 
 func (c *CommonInstancetypes) Reconcile(request *common.Request) ([]common.ReconcileResult, error) {
@@ -309,17 +319,31 @@ func (c *CommonInstancetypes) Cleanup(request *common.Request) ([]common.Cleanup
 	return nil, nil
 }
 
-func (c *CommonInstancetypes) reconcileFuncs() []common.ReconcileFunc {
+func (c *CommonInstancetypes) reconcileFuncs(request *common.Request) ([]common.ReconcileFunc, error) {
+	instancetypeFuncs, err := c.reconcileVirtualMachineClusterInstancetypesFuncs(request)
+	if err != nil {
+		return nil, err
+	}
+	preferenceFuncs, err := c.reconcileVirtualMachineClusterPreferencesFuncs(request)
+	if err != nil {
+		return nil, err
+	}
+
 	funcs := []common.ReconcileFunc{}
-	funcs = append(funcs, c.reconcileVirtualMachineClusterInstancetypesFuncs()...)
-	funcs = append(funcs, c.reconcileVirtualMachineClusterPreferencesFuncs()...)
-	return funcs
+	funcs = append(funcs, instancetypeFuncs...)
+	funcs = append(funcs, preferenceFuncs...)
+	return funcs, nil
 }
 
-func (c *CommonInstancetypes) reconcileVirtualMachineClusterInstancetypesFuncs() []common.ReconcileFunc {
+func (c *CommonInstancetypes) reconcileVirtualMachineClusterInstancetypesFuncs(request *common.Request) ([]common.ReconcileFunc, error) {
 	funcs := make([]common.ReconcileFunc, 0, len(c.virtualMachineClusterInstancetypes))
 	for i := range c.virtualMachineClusterInstancetypes {
 		clusterInstancetype := &c.virtualMachineClusterInstancetypes[i]
+		if ignore, err := ignoreObjectOwnedByVirtOperator(request, clusterInstancetype); err != nil {
+			return nil, err
+		} else if ignore {
+			continue
+		}
 		funcs = append(funcs, func(request *common.Request) (common.ReconcileResult, error) {
 			return common.CreateOrUpdate(request).
 				ClusterResource(clusterInstancetype).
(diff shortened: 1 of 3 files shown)


internal/common/resource.go
```diff
@@ -161,6 +167,11 @@ func (r *reconcileBuilder) Options(options ReconcileOptions) ReconcileBuilder {
 	return r
 }
 
+func (r *reconcileBuilder) IgnoreObjectOwnedByVirtOperator() ReconcileBuilder {
```
lyarwood, line 170: FWIW if folks think we might need to ignore other types of resource in the future I can make this more generic passing a function as with the `UpdateFunc` and `StatusFunc` methods but at the moment I think it's better to be explicit.

lyarwood:
I'll respin with more tests later today but comments welcome in the meantime.

internal/common/resource.go
```diff
@@ -303,6 +314,14 @@ func (r *reconcileBuilder) createOrUpdateWithImmutableSpec(obj client.Object, f
 	}
 
 	existing := obj.DeepCopyObject().(client.Object)
+
+	// During an upgrade to v0.19.0 we might encounter virt-operator attempting
+	// to reconcile the same set of common-instancetype resources. If configured
+	// ignore these requests to reconcile such objects once owned by virt-operator
+	if r.options.IgnoreObjectOwnedByVirtOperator && isOwnedByVirtOperator(existing) {
```
0xFelix, line 321: Put this at the start of `Reconcile` instead to short-circuit the whole reconciliation?

0xFelix:
Can / should we test that SSP still reaches phase `Deployed` when `virt-operator` takes over?

internal/common/resource.go
```diff
@@ -15,6 +15,7 @@ import (
 	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	kubevirtCorev1 "kubevirt.io/api/core/v1"
```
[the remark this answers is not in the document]
lyarwood: what about `kubevirtv1`?
```
$ git grep -h "kubevirt.io/api/core/v1" | awk '{ print $1 }' | sort | uniq
corev1
k6tv1
kubevirt
kubevirtCorev1
kubevirt.io/api/core/v1
kubevirtv1
v1
```

internal/common/resource.go
lyarwood: Moving this earlier would cause another client lookup for every object for every reconcile. We can move this but I'm not sure it's worth it.

In terms of the `VersionCache` yeah we can avoid this but AFAICT there's no impact of this being registered?

lyarwood:
> Can / should we test that SSP still reaches phase `Deployed` when `virt-operator` takes over?

Yup that's easy enough to mock in an e2e test and would give us coverage in SSP without needing HCO or an updated `virt-operator`.

internal/common/resource.go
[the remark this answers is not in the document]
0xFelix: Or just `virtv1`, that's what's used in kubevirt/kubevirt a lot.

internal/common/resource.go
0xFelix: Alright, then let's keep it here.

Just wanted to make sure we don't create any issues by caching resources we do not own.

internal/common/resource.go
```diff
@@ -391,6 +418,10 @@ func logOperation(result OperationResult, resource client.Object, logger logr.Lo
 		logger.Info(fmt.Sprintf("Deleted %s resource: %s",
 			resource.GetObjectKind().GroupVersionKind().Kind,
 			resource.GetName()))
+	case OperationResultIgnored:
+		logger.Info(fmt.Sprintf("Ignored %s resource: %s",
```
0xFelix, line 422:
```suggestion
		logger.Info(fmt.Sprintf("Ignored %s resource owned by virt-operator: %s",
```

WDYT?

0xFelix (commented):
Are you still going to add an e2e for the upgrade scenario?

internal/common/resource.go
akrejcir: IMHO this is too specific to be in the `reconcileBuilder`. Can you move the logic to the functions that reconcile `InstanceTypes` and `Preferences`? `reconcileVirtualMachineClusterInstancetypesFuncs()` and `reconcileVirtualMachineClusterPreferencesFuncs()`

There you can check if the resource is owned by virt-operator or not.

internal/common/resource.go
akrejcir: > Moving this earlier would cause another client lookup for every object for every reconcile. We can move this but I'm not sure it's worth it.

Client lookups are cheap, because everything is cached in memory. The API client watches the cluster, and any `Get()` or `List()` function calls just read the cahce.

internal/common/resource.go
```diff
@@ -337,6 +356,14 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj client.Object)
 	return nil
 }
 
+func isOwnedByVirtOperator(obj client.Object) bool {
+	if obj.GetLabels() == nil {
+		return false
+	}
+	managedValue, managedOk := obj.GetLabels()[kubevirtCorev1.ManagedByLabel]
```
akrejcir, line 363: You can use constant `AppKubernetesManagedByLabel`, instead of the imported one `kubevirtCorev1.ManagedByLabel`.

0xFelix:
> Have you considered removing the SSP owner annotations from the InstanceTypes and Preferences that are owned by virt-operator? Then they would no longer trigger SSP reconciliation.

virt-operator reconstructs the metadata of objects during reconciliation, so this should already be the case.

internal/common/resource.go
```diff
@@ -28,6 +28,7 @@ const (
 	OperationResultCreated OperationResult = "created"
 	OperationResultUpdated OperationResult = "updated"
 	OperationResultDeleted OperationResult = "deleted"
+	OperationResultIgnored OperationResult = "ignored"
```
akrejcir, line 31: This constant is not needed.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -321,6 +323,12 @@ func (c *CommonInstancetypes) reconcileVirtualMachineClusterInstancetypesFuncs()
 	for i := range c.virtualMachineClusterInstancetypes {
 		clusterInstancetype := &c.virtualMachineClusterInstancetypes[i]
 		funcs = append(funcs, func(request *common.Request) (common.ReconcileResult, error) {
+			if ignore, err := ignoreObjectOwnedByVirtOperator(request, clusterInstancetype.DeepCopy()); err != nil {
```
akrejcir, line 326: Can you move this check outside of this closure? If `ignoreObjectOwnedByVirtOperator()` returns true, we don't need to add this closure to the `funcs` slice.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -335,6 +343,12 @@ func (c *CommonInstancetypes) reconcileVirtualMachineClusterPreferencesFuncs() [
 	for i := range c.virtualMachineClusterPreferences {
 		clusterPreference := &c.virtualMachineClusterPreferences[i]
 		funcs = append(funcs, func(request *common.Request) (common.ReconcileResult, error) {
+			if ignore, err := ignoreObjectOwnedByVirtOperator(request, clusterPreference.DeepCopy()); err != nil {
```
akrejcir, line 346: The same as above.

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -343,3 +357,25 @@ func (c *CommonInstancetypes) reconcileVirtualMachineClusterPreferencesFuncs() [
 	}
 	return funcs
 }
+
+func ignoreObjectOwnedByVirtOperator(request *common.Request, obj client.Object) (bool, error) {
+	// During an upgrade to v0.19.0 we might encounter virt-operator attempting
+	// to reconcile the same set of common-instancetype resources. Ignore these
+	// requests to reconcile such objects once owned by virt-operator.
+	if err := request.Client.Get(request.Context, client.ObjectKeyFromObject(obj), obj); err != nil {
```
akrejcir, line 365: nit: This call modifies the `obj` parameter. I would not expect this function to modify its parameters.
But we can change it in a future PR.
Can you please write a "TODO" comment here, so we don't forget about it?

internal/operands/common-instancetypes/reconcile_test.go
```diff
@@ -251,6 +252,56 @@ var _ = Describe("Common-Instancetypes operand", func() {
...
+
+		results, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Assert that we have reported ignoring the attempt to reconcile the objects owned by virt-operator
+		for _, res := range results {
+			if res.Resource.GetName() == instancetypeToUpdate.Name || res.Resource.GetName() == preferenceToUpdate.Name {
+				Expect(res.OperationResult).To(Equal(common.OperationResultIgnored))
+			}
+		}
```
akrejcir, line 296: Can you instead check if the `InstanceType` and `Preferece` with the label were not reverted?

internal/common/resource.go
0xFelix: ACK

internal/operands/common-instancetypes/reconcile.go
0xFelix: A client is needed to retrieve the object from the API. I don't think this is possible outside the closure?

internal/operands/common-instancetypes/reconcile_test.go
0xFelix: That is already done below.

internal/operands/common-instancetypes/reconcile.go
akrejcir: You can pass the `*common.Request` also to the enclosing function. We do something similar with DataSources:
https://github.com/kubevirt/ssp-operator/blob/9759ddfe94f066df6cb0d97e727bf06342b6d7bd/internal/operands/data-sources/reconcile.go#L335

Issue #682, opened 2023-09-06
Operator crashes when kubevirt is not installed


akrejcir:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
If `ssp-operator` deployment is created on a cluster without kubevirt CRDs, it will crash and cause `CrashLoopBackoff` state.

This can happen when `ssp-operator` is installed using OLM.

codingben:
We'll need to check which operand is using KubeVirt CRD and stop reconciliation of it? like we do in Tekton operands, if Tekton CRD doesn't exist.

akrejcir:
It is probably caused by the new controller that watches VMs, introduced in: https://github.com/kubevirt/ssp-operator/pull/663

0xFelix:
Pinging @machadovilaca, can you PTAL?

akrejcir:
I did not realize it during the review. But this is not an urgent bug. I've opened this issue so that we don't forget about it.

machadovilaca:
Looking

PR #685, opened 2023-09-07
Do not register VM controller if KubeVirt CRDs are not installed

machadovilaca:

If `ssp-operator` deployment is created on a cluster without KubeVirt CRDs, it will crash and cause `CrashLoopBackoff` state. This PR prevents the creation on the VM controller if the KubeVirt CRDs are not installed.

Fixes https://github.com/kubevirt/ssp-operator/issues/682

**Release note**:

```release-note
Do not register VM controller if KubeVirt CRDs are not installed
```

### DIFF
--- a/controllers/setup.go
+++ b/controllers/setup.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/go-logr/logr"
 	v1 "github.com/openshift/api/config/v1"
+	kubevirtv1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/ssp-operator/internal/common"
 	crd_watch "kubevirt.io/ssp-operator/internal/crd-watch"
 	"kubevirt.io/ssp-operator/internal/operands"
@@ -103,6 +105,11 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 		requiredCrds = append(requiredCrds, getRequiredCrds(sspOperands[i])...)
 	}
 
+	// Add VMController necessary VirtualMachine CRD
+	vmKind := strings.ToLower(kubevirtv1.VirtualMachineGroupVersionKind.Kind) + "s"
+	vmCRD := vmKind + "." + kubevirtv1.VirtualMachineGroupVersionKind.Group
+	requiredCrds = append(requiredCrds, vmCRD)
+
 	crdWatch := crd_watch.New(requiredCrds...)
 	// Cleanly stops the manager and exit. The pod will be restarted.
 	crdWatch.AllCrdsAddedHandler = cancel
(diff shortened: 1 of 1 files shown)


controllers/setup.go
```diff
@@ -139,13 +139,22 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 		return fmt.Errorf("error adding service controller: %w", err)
 	}
 
-	vmController, err := CreateVmController(mgr)
+	hasKubeVirtCRDs, err := common.CheckIfClusterHasKubeVirtCRDs(ctx, mgr.GetAPIReader())
 	if err != nil {
-		return fmt.Errorf("failed to create vm controller: %w", err)
-	}
+		return fmt.Errorf("[vm controller] failed to check if cluster has KubeVirt CRDs: %w", err)
+	} else if hasKubeVirtCRDs {
```
0xFelix, line 145: This can be a plain `if`, no need for `else`.

0xFelix (commented):
@akrejcir I wonder if if makes sense to add the KubeVirt CRD to the crdWatch in L103 instead? Or in a more general way instead, as this operator has no use if KubeVirt is not present?

controllers/setup.go
machadovilaca: updated

akrejcir:
> @akrejcir I wonder if if makes sense to add the KubeVirt CRD to the crdWatch in L103 instead? Or in a more general way instead, as this operator has no use if KubeVirt is not present?

Yes, It would be better. Then when the KubeVirt CRDs are added, the `crdWatch` will trigger operator restart.

@machadovilaca, can you please add the CRDs to the `crdWatch`, and only create the vmController if `crdWatch.CrdExists()` returns `true`?

controllers/setup.go
```diff
@@ -139,13 +146,17 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
 		return fmt.Errorf("error adding service controller: %w", err)
 	}
 
-	vmController, err := CreateVmController(mgr)
-	if err != nil {
-		return fmt.Errorf("failed to create vm controller: %w", err)
-	}
+	if crdWatch.CrdExists(vmCRD) {
+		vmController, cErr := CreateVmController(mgr)
```
akrejcir, line 150: nit, just for info: It is ok to use `err` variable name instead of `cErr`. It would shadow the `err` the outer context, but that does not matter, because if `cErr != nil` we always return the error.

You don't have to fix this.

akrejcir (commented):
Looks good. Just one nit, that you don't have to fix.

/lgtm
/approve

akrejcir:
@machadovilaca I forgot to ask, did you verify this locally? There is no simple way how to write a functional test for it.

machadovilaca:
yes, I ran SSP without KubeVirt CRDs in the cluster!

PR #686, opened 2023-09-07
Reduce VirtualMachineCRCErrors noise

machadovilaca:

Instead of issuing one alert per VM, we now issue one alert, with the count of VMs with problems

Fixes #

**Release note**:

```release-note
Reduce VirtualMachineCRCErrors noise
```


0xFelix (commented):
/approve

Thanks

PR #696, opened 2023-10-05
fix: remove "cdi.kubevirt.io/storage.bind.immediate.requested" annotation from pipelines

ksimon1:
fix: remove "cdi.kubevirt.io/storage.bind.immediate.requested" annotation from pipelines
CDI does not honor this annotation when WFFC VolumeBindingMode is set. This is causing a problem, that PVCs are in pending state and pipelines fail.
this commit removes the annotations, because the pvcs are going to be consumed by VM.

Fixes https://github.com/kubevirt/containerized-data-importer/issues/2918

**Release note**:
```
fix: remove "cdi.kubevirt.io/storage.bind.immediate.requested" annotation from pipelines
```
/hold WIP

### DIFF
--- a/data/tekton-pipelines/windows-customize-pipeline.yaml
+++ b/data/tekton-pipelines/windows-customize-pipeline.yaml
@@ -54,8 +54,6 @@ spec:
             apiVersion: cdi.kubevirt.io/v1beta1
             kind: DataVolume
             metadata:
-              annotations:
-                "cdi.kubevirt.io/storage.bind.immediate.requested": "true"
               name: $(params.baseDvName)
               namespace: $(params.baseDvNamespace)
             spec:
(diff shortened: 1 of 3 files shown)


data/tekton-pipelines/windows-customize-pipeline.yaml
```diff
@@ -54,8 +54,6 @@ spec:
             apiVersion: cdi.kubevirt.io/v1beta1
             kind: DataVolume
             metadata:
-              annotations:
```
akalenyu, line 57: You should keep this annotation on the `spec.source.pvc` DVs
It's only broken not honored in `spec.source.blank` Block DVs

akalenyu: Ofc if youre creating a VM out of those DVs then the annotation just doesn't make much sense
so up to you

PR #715, opened 2023-10-27
fix: instance types tests by waiting for ssp to be deployed

ksimon1:
fix: instance types tests by waiting for ssp to be deployed

**Release note**:
```
NONE
```


0xFelix (approved):
/approve
/lgtm

PR #731, opened 2023-11-13
reconcile: Restore test that go deleted by accident

0xFelix:

Restore a test that got deleted by accident in
de115ba7e020d08719baecb963df78ade839f0b6.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/internal/operands/data-sources/reconcile_test.go
+++ b/internal/operands/data-sources/reconcile_test.go
@@ -220,6 +220,37 @@ var _ = Describe("Data-Sources operand", func() {
 				Expect(request.Client.Get(request.Context, client.ObjectKeyFromObject(&testDataSources[0]), ds)).To(Succeed())
 				Expect(ds.Spec).To(Equal(testDataSources[0].Spec))
 			})
+
+			It("should not restore DataSource if DataImportCron prefers snapshots sources", func() {
+				_, err := operand.Reconcile(&request)
+				Expect(err).ToNot(HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


0xFelix:
/hold

Clarifying why de115ba was needed in the first place.

akrejcir:
The PR https://github.com/kubevirt/ssp-operator/pull/728 fixed a bug: https://issues.redhat.com/browse/CNV-34614

Previously SSP assumed that if `.spec.source.PVC` field on `DataSource` is `nil`, then there is no `DataImportCron` that fills this field. The logic was correct when the `.spec.source.snapshot` filed did not exist. But after it was added, this logic is incorrect.

The PR fixed it, and now `DataSource` is reconciled by SSP only when there is no `DataImportCron`.

I've removed the unit test, because after the fix I thought it was not needed, and that the case of not reconciling `DataSource` by SSP is covered in a different test case.
Now I see that we are missing unit test for it.

@0xFelix, can you please change the test, so that it is not specific to snapshot? It should check that SSP does not overwrite changes to `DataSource` done by CDI.

akalenyu:
> so that it is not specific to snapshot? It should check that SSP does not overwrite changes to `DataSource` done by CDI.

I would sleep better at night if we had a snapshot permutation to the test/keep this one

akrejcir:
> I would sleep better at night if we had a snapshot permutation to the test/keep this one

Ok we can keep this test.

PR #742, opened 2023-11-28
Fix VirtualMachineCRCErrors not stop firing

machadovilaca:

The controller keeps reporting old metric values even after the VM is deleted. This PR updates the metric name and labels so that we can set up the metric value to 0, and no longer trigger the alert on VM deletion

Reduces VirtualMachineCRCErrors operator impact to none

Fixes #

**Release note**:

```release-note
Fix VirtualMachineCRCErrors not stop firing
```

### DIFF
--- a/controllers/vm_controller.go
+++ b/controllers/vm_controller.go
@@ -58,7 +58,11 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 	vm := kubevirtv1.VirtualMachine{}
 	if err := r.client.Get(ctx, req.NamespacedName, &vm); err != nil {
 		if errors.IsNotFound(err) {
-			// VM was deleted, so we can ignore it
+			// VM was deleted
+			vm.Name = req.Name
+			vm.Namespace = req.Namespace
+			metrics.SetVmWithVolume(&vm, nil, nil)
+
 			return ctrl.Result{}, nil
 		}
 		r.log.Error(err, "Error getting VM", "vm", req.NamespacedName)
--- a/internal/operands/metrics/resources.go
+++ b/internal/operands/metrics/resources.go
@@ -166,7 +166,7 @@ func getAlertRules() ([]promv1.Rule, error) {
 		},
 		{
 			Alert: "VirtualMachineCRCErrors",
-			Expr:  intstr.FromString("(count(kubevirt_ssp_vm_rbd_volume{volume_mode=\"Block\", rxbounce_enabled=\"false\"} > 0) or vector(0)) > 0"),
+			Expr:  intstr.FromString("(count(kubevirt_ssp_vm_rbd_block_volume_without_rxbounce > 0) or vector(0)) > 0"),
 			Annotations: map[string]string{
 				"description": "{{ $value }} Virtual Machines are in risk of causing CRC errors and major service outages",
 				"summary":     "When running VMs using ODF storage with 'rbd' mounter or 'rbd.csi.ceph.com provisioner', it will report bad crc/signature errors and cluster performance will be severely degraded if krbd:rxbounce is not set.",
@@ -174,7 +174,7 @@ func getAlertRules() ([]promv1.Rule, error) {
 			},
 			Labels: map[string]string{
 				severityAlertLabelKey:     "warning",
-				healthImpactAlertLabelKey: "warning",
+				healthImpactAlertLabelKey: "none",
 				partOfAlertLabelKey:       partOfAlertLabelValue,
 				componentAlertLabelKey:    componentAlertLabelValue,
 			},
--- a/tests/monitoring_test.go
+++ b/tests/monitoring_test.go
@@ -20,6 +20,7 @@ import (
 	promConfig "github.com/prometheus/common/config"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -175,7 +176,7 @@ var _ = Describe("Prometheus Alerts", func() {
 
 		AfterEach(func() {
 			vmError := apiClient.Delete(ctx, vm)
-			Expect(vmError).ToNot(HaveOccurred())
+			Expect(client.IgnoreNotFound(vmError)).ToNot(HaveOccurred())
 
 			if pvc != nil {
 				pvcError := apiClient.Delete(ctx, pvc)
@@ -271,22 +272,26 @@ var _ = Describe("Prometheus Alerts", func() {
 			return vmName
 		}
 
-		It("[test_id:TODO] Should not create kubevirt_ssp_vm_rbd_volume when not using DataVolume or PVC", func() {
-			vmName := createResources(false, true)
-			seriesShouldNotBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_volume{name='%s'}", vmName))
-			alertShouldNotBeActive("VirtualMachineCRCErrors")
-		})
-
 		It("[test_id:TODO] Should not fire VirtualMachineCRCErrors when rxbounce is enabled", func() {
 			vmName := createResources(true, true)
-			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_volume{name='%s', rxbounce_enabled='true'}", vmName))
+			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_block_volume_without_rxbounce{name='%s'} == 0", vmName))
 			alertShouldNotBeActive("VirtualMachineCRCErrors")
 		})
 
 		It("[test_id:10549] Should fire VirtualMachineCRCErrors when rxbounce is disabled", func() {
 			vmName := createResources(true, false)
-			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_volume{name='%s', rxbounce_enabled='false'}", vmName))
+			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_block_volume_without_rxbounce{name='%s'} == 1", vmName))
 			waitForAlertToActivate("VirtualMachineCRCErrors")
+
+			err := apiClient.Delete(ctx, vm)
+			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func(g Gomega) error {
+				return apiClient.Get(ctx, types.NamespacedName{Name: vmName, Namespace: strategy.GetNamespace()}, vm)
+			}).Should(MatchError(k8serrors.IsNotFound, "IsNotFound"))
+
+			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_block_volume_without_rxbounce{name='%s'} == 0", vmName))
+			alertShouldNotBeActive("VirtualMachineCRCErrors")
 		})
 	})
 })
@@ -319,7 +324,7 @@ func alertShouldNotBeActive(alertName string) {
 		if err != nil {
 			return err
 		}
-		if alert == nil {
+		if alert == nil || alert.State == "inactive" {
 			return nil
 		}
 		return fmt.Errorf("alert %s found", alertName)
@@ -330,7 +335,7 @@ func alertShouldNotBeActive(alertName string) {
 		if err != nil {
 			return err
 		}
-		if alert == nil {
+		if alert == nil || alert.State == "inactive" {
 			return nil
 		}
 		return fmt.Errorf("alert %s found", alertName)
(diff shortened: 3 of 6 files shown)


tests/monitoring_test.go
```diff
@@ -287,6 +287,11 @@ var _ = Describe("Prometheus Alerts", func() {
 			vmName := createResources(true, false)
 			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_volume{name='%s', rxbounce_enabled='false'}", vmName))
 			waitForAlertToActivate("VirtualMachineCRCErrors")
+
+			Eventually(func() error {
```
0xFelix, line 291: Get VM and check for error is not found instead?

machadovilaca: updated

machadovilaca:
> This PR changes quite a lot. Can you explain a little how the fix works and add the explanation also to the PR and commit descriptions?

@0xFelix The problem was that the controller kept reporting the metric for old VMs even after they were deleted. In the controller, we can understand when a VM is deleted, but at that moment we don't have access to the details present in the old labels (pv name, pv type, etc...). This PR changes the metric so that it only uses the VM name and namespace as labels, so that we can update its value, when a VM is deleted, to a no-error value. Alert expression is updated accordingly.

0xFelix:
This also replaces `kubevirt_ssp_vm_rbd_volume` with a new metric named `kubevirt_ssp_vm_rbd_block_volume_without_rxbounce`. Is that necessary for the fix you described?

Again, please update the PR and commit descriptions.

machadovilaca:
> This also replaces `kubevirt_ssp_vm_rbd_volume` with a new metric named `kubevirt_ssp_vm_rbd_block_volume_without_rxbounce`. Is that necessary for the fix you described?
> 
> Again, please update the PR and commit descriptions.

yes, old name was not accurate after removing the previous labels
updated the PR description

internal/operands/metrics/resources.go
```diff
@@ -166,15 +166,15 @@ func getAlertRules() ([]promv1.Rule, error) {
...
+			Expr:  intstr.FromString("(count(kubevirt_ssp_vm_rbd_block_volume_without_rxbounce > 0) or vector(0)) > 0"),
 			Annotations: map[string]string{
 				"description": "{{ $value }} Virtual Machines are in risk of causing CRC errors and major service outages",
 				"summary":     "When running VMs using ODF storage with 'rbd' mounter or 'rbd.csi.ceph.com provisioner', it will report bad crc/signature errors and cluster performance will be severely degraded if krbd:rxbounce is not set.",
 				"runbook_url": fmt.Sprintf(runbookURLTemplate, "VirtualMachineCRCErrors"),
 			},
 			Labels: map[string]string{
 				severityAlertLabelKey:     "warning",
-				healthImpactAlertLabelKey: "warning",
+				healthImpactAlertLabelKey: "none",
```
0xFelix, line 177: No health impact anymore?

tests/monitoring_test.go
```diff
@@ -175,7 +175,7 @@ var _ = Describe("Prometheus Alerts", func() {
 
 		AfterEach(func() {
 			vmError := apiClient.Delete(ctx, vm)
-			Expect(vmError).ToNot(HaveOccurred())
+			Expect(vmError == nil || client.IgnoreNotFound(vmError) == nil).To(BeTrue())
```
0xFelix, line 178: What about this?

```suggestion
			Expect(client.IgnoreNotFound(vmError)).ToNot(HaveOccured())
```

tests/monitoring_test.go
```diff
@@ -271,22 +271,24 @@ var _ = Describe("Prometheus Alerts", func() {
...
 			vmName := createResources(true, false)
-			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_volume{name='%s', rxbounce_enabled='false'}", vmName))
+			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_block_volume_without_rxbounce{name='%s'}", vmName))
 			waitForAlertToActivate("VirtualMachineCRCErrors")
+
+			err := apiClient.Delete(ctx, vm)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func() error {
+				return apiClient.Get(ctx, types.NamespacedName{Name: vmName, Namespace: strategy.GetNamespace()}, vm)
+			}).ShouldNot(Succeed())
```
0xFelix, line 289: This does not check if the returned error is a not found error.

internal/operands/metrics/resources.go
machadovilaca: https://issues.redhat.com/browse/CNV-35848

tests/monitoring_test.go
machadovilaca: The VM was created in the beginning of the test, VM should exist and be deleted, we don't expect a not found error here

tests/monitoring_test.go
machadovilaca: updated

internal/operands/metrics/resources.go
machadovilaca: done

tests/monitoring_test.go
machadovilaca: misunderstood the first time, you are right, updated

tests/monitoring_test.go
```diff
@@ -271,22 +271,27 @@ var _ = Describe("Prometheus Alerts", func() {
...
+			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_block_volume_without_rxbounce{name='%s'}", vmName))
 			waitForAlertToActivate("VirtualMachineCRCErrors")
+
+			err := apiClient.Delete(ctx, vm)
+			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func(g Gomega) {
+				err := apiClient.Get(ctx, types.NamespacedName{Name: vmName, Namespace: strategy.GetNamespace()}, vm)
+				g.Expect(err).To(HaveOccurred())
+				g.Expect(errors.IsNotFound(err)).To(BeTrue())
```
0xFelix, line 291: Please use `To(MatchError(...))` instead.

machadovilaca: I see that `MatchError` is usually used for checking the error contains a given string
Here I think it is better to take advantage of all the checks `errors.IsNotFound` provides

0xFelix: `MatchError` is able to take the matcher fn: https://onsi.github.io/gomega/#matcherrorexpected-interface

machadovilaca: but `k8s.io/apimachinery/pkg/api/errors` does not provide a way to compare to `metav1.StatusReasonNotFound`, its encapsulated in a `APIStatus` struct:

```
if status, ok := err.(APIStatus); ok || errors.As(err, &status) {
	return status.Status().Reason, status.Status().Code
}
```

0xFelix: `g.Expect(err).To(MatchError(errors.IsNotFound, "IsNotFound"))`

Does this work?

machadovilaca: it does :tada: 
updated

tests/monitoring_test.go
```diff
@@ -162,7 +162,7 @@ var _ = Describe("Prometheus Alerts", func() {
 		})
 	})
 
-	Context("VirtualMachineCRCErrors", func() {
+	FContext("VirtualMachineCRCErrors", func() {
```
0xFelix, line 165: Please drop this change

tests/monitoring_test.go
```diff
@@ -271,22 +271,27 @@ var _ = Describe("Prometheus Alerts", func() {
...
-			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_volume{name='%s', rxbounce_enabled='false'}", vmName))
+			waitForSeriesToBeDetected(fmt.Sprintf("kubevirt_ssp_vm_rbd_block_volume_without_rxbounce{name='%s'}", vmName))
 			waitForAlertToActivate("VirtualMachineCRCErrors")
+
+			err := apiClient.Delete(ctx, vm)
+			Expect(err).ToNot(HaveOccurred())
+
+			Eventually(func(g Gomega) {
+				err := apiClient.Get(ctx, types.NamespacedName{Name: vmName, Namespace: strategy.GetNamespace()}, vm)
+				g.Expect(err).To(HaveOccurred())
```
0xFelix, line 290: You don't need this line any more, it's covered by `MatchError` already.

tests/monitoring_test.go
```diff
@@ -301,7 +306,7 @@ func checkAlert(alertName string) (*promApiv1.Alert, error) {
 }
 
 func waitForAlertToActivate(alertName string) {
-	Eventually(func() error {
+	EventuallyWithOffset(1, func() error {
```
0xFelix, line 309: Is this necessary for the actual fix?

tests/monitoring_test.go
```diff
@@ -314,7 +319,7 @@ func waitForAlertToActivate(alertName string) {
 }
 
 func alertShouldNotBeActive(alertName string) {
-	Eventually(func() error {
+	EventuallyWithOffset(1, func() error {
```
0xFelix, line 322: Same

tests/monitoring_test.go
```diff
@@ -325,7 +330,7 @@ func alertShouldNotBeActive(alertName string) {
 		return fmt.Errorf("alert %s found", alertName)
 	}, env.Timeout(), time.Second).ShouldNot(HaveOccurred())
 
-	Consistently(func() error {
+	ConsistentlyWithOffset(1, func() error {
```
0xFelix, line 333: Same

tests/monitoring_test.go
```diff
@@ -338,31 +343,13 @@ func alertShouldNotBeActive(alertName string) {
 }
 
 func waitForSeriesToBeDetected(seriesName string) {
-	Eventually(func() bool {
+	EventuallyWithOffset(1, func() bool {
```
0xFelix, line 346: Same

tests/monitoring_test.go
```diff
@@ -435,7 +422,7 @@ func getAuthorizationTokenForPrometheus() string {
 			if strings.HasPrefix(secret.Name, "prometheus-k8s-token") {
 				tokenBytes, ok = secret.Data["token"]
 				if !ok {
-					return errors.New("token not found in secret data")
+					return fmt.Errorf("token not found in secret %s", secret.Name)
```
0xFelix, line 425: Change not related to this fix?

machadovilaca: removed

tests/monitoring_test.go
machadovilaca: removed

tests/monitoring_test.go
machadovilaca: removed

controllers/vm_controller.go
```diff
@@ -58,7 +58,11 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 	vm := kubevirtv1.VirtualMachine{}
 	if err := r.client.Get(ctx, req.NamespacedName, &vm); err != nil {
 		if errors.IsNotFound(err) {
-			// VM was deleted, so we can ignore it
+			// VM was deleted
+			vm.Name = req.Name
```
0xFelix, line 62: If the VM was deleted, why do we add it to the metrics? Also it looks error-prone to call `metrics.SetVmWithVolume` with nil-pointers, as IIUC this func does not check for nil pointers sufficiently.

tests/monitoring_test.go
0xFelix: OK, can you at least move these changes into a separate commit please?

controllers/vm_controller.go
machadovilaca: - the issue was that the controller kept exposing the metric that the VM, although already deleted, as the problematic volume configuration
this way when a VM is deleted, the controller no longer has a true value for the metric

- I think the nil check in the function covers this case completely:

```
if pv == nil || pvc == nil {
  vmRbdVolume.WithLabelValues(vm.Name, vm.Namespace).Set(0)
  return
}
```

tests/monitoring_test.go
machadovilaca: done!

controllers/vm_controller.go
0xFelix: https://github.com/kubevirt/ssp-operator/blob/33b6ca71959e903fc1f7abf813ef38d110061a10/pkg/monitoring/metrics/rbd_metrics.go#L29

The function that is called looks different, than what you quoted?

machadovilaca: I also changed that function on this PR:
https://github.com/kubevirt/ssp-operator/pull/742/files#diff-387c26f005e40fae497f3a5c01da1f877fc4eee90da11b7d8cf47512d8474b82R29-R31

0xFelix: Right, my mistake.

0xFelix (commented):
/approve

0xFelix:
If you want this to be in 4.15 you need to cherry-pick it to release-v0.19.

Issue #746, opened 2023-12-01
Code duplication in ssp_webhook.go


dominikholler:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind enhancement

**What happened**:

ValidateCreate and ValidateUpdate shares some logic, so it would be useful to share the code, too.

PR #747, opened 2023-12-04
fix: webhook: Don't check common template namespace

akrejcir:
The namespace of common templates is not checked in webhook.

Even if the webhook passes, user can still remove the namespace later.

Fixes: https://github.com/kubevirt/ssp-operator/issues/746
Fixes: https://issues.redhat.com/browse/CNV-35920

**Release note**:
```release-note
Validation webhook does not check if namespace for templates exists.
```


0xFelix (commented):
Looks good to me, thanks!

/approve

PR #803, opened 2023-12-25
fix: enable arm64 operator build

### LINKED ISSUE #802: Developer can't build SSP on an arm64 machine
Enable building of ssp-operator and template-validator on arm64 machine. It's essential when developing the operator.

codingben:

Enable developer building operator on an arm64 machine as an additional option.

Jira-Url: https://issues.redhat.com/browse/CNV-36775

Fixes #802

**Release note**:

```release-note
None
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
 
 RUN microdnf install -y make tar gzip which && microdnf clean all
 
-RUN curl -L https://go.dev/dl/go1.20.11.linux-amd64.tar.gz | tar -C /usr/local -xzf -
+RUN curl -L https://go.dev/dl/go1.20.11.linux-arm64.tar.gz | tar -C /usr/local -xzf -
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
 
 RUN microdnf install -y make tar gzip which && microdnf clean all
 
-RUN curl -L https://go.dev/dl/go1.20.11.linux-amd64.tar.gz | tar -C /usr/local -xzf -
```
codingben, line 6: @akrejcir @0xFelix How do you suggest to check if it's arm64 or amd64 here? Isn't architecture usually based on base images?

akrejcir: There may be some way to detect which architecture is used and download the right package based on that.
This was the reason why I dropped my original [PR](https://github.com/kubevirt/ssp-operator/pull/485). I didn't want to spend too much time figuring this out, if no one would use it.

codingben:
I'm closing this PR because it's not needed - I'm using VM to build SSP.

PR #826, opened 2024-01-11
feat: dependabot configuration

ksimon1:
this PR removes configuration for release branches,
because it created a lot of PRs with indirect dependencies.
It allows grouping of updates into single PRs

**Release note**:
```
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,35 +6,14 @@ updates:
       interval: "daily"
     allow:
       - dependency-type: "all"
-  - package-ecosystem: "gomod"
-    directory: "/"
-    target-branch: "release-v0.18"
-    schedule:
-      interval: "daily"
-    allow:
-      - dependency-type: "all"
-  - package-ecosystem: "gomod"
-    directory: "/"
-    target-branch: "release-v0.17"
-    schedule:
-      interval: "daily"
-    allow:
-      - dependency-type: "all"
-  - package-ecosystem: "gomod"
-    directory: "/"
-    target-branch: "release-v0.16"
-    schedule:
-      interval: "daily"
-    allow:
-      - dependency-type: "all"
-  - package-ecosystem: "gomod"
-    directory: "/"
-    target-branch: "release-v0.15"
-    schedule:
-      interval: "daily"
-    allow:
-      - dependency-type: "all"
+    groups:
+      production-dependencies:
+        dependency-type: "production"
+      development-dependencies:
+        dependency-type: "development"
+    open-pull-requests-limit: 3
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "weekly"  
+    open-pull-requests-limit: 3


jcanocan:
Thanks @ksimon1 ! 
/lgtm

.github/dependabot.yml
```diff
@@ -6,35 +6,54 @@ updates:
...
+        dependency-type: "development"
+    open-pull-requests-limit: 3
+  - package-ecosystem: "gomod"
+    directory: "/"
+    target-branch: "release-v0.19"
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-type: "direct"
+    open-pull-requests-limit: 0
```
0xFelix, line 22: If we set this to 0, do we need the configuration for this and others branches at all?

ksimon1: yes, because this should not affect security issue PRs.
```
This option has no impact on security updates, which have a separate, internal limit of ten open pull requests.
```
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

akrejcir: From the documentation, I've understood that the security update PRs are only opened for the default branch:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#configuration-options-for-the-dependabotyml-file

> Security updates are raised for vulnerable package manifests only on the default branch. When configuration options are set for the same branch (true unless you use target-branch), and specify a package-ecosystem and directory for the vulnerable manifest, then pull requests for security updates use relevant options.

0xFelix (commented):
/approve

Looks good to me!

.github/dependabot.yml
```diff
@@ -6,35 +6,14 @@ updates:
       interval: "daily"
     allow:
       - dependency-type: "all"
```
akrejcir, line 8: What do you think about limiting this only to direct dependencies:
```yaml
dependency-type: "direct"
```

And then we can add another `updates` block for security updates, which would also update indirect dependencies:
```yaml
- package-ecosystem: "gomod"
    directory: "/"
    schedule:
      interval: "daily"
    allow:
      - dependency-type: "all"
    open-pull-requests-limit: 0
```

akrejcir:
I was too late with review :)

0xFelix:
Are you able to configure security updates at all in this file? I'm still confused and under the impression that security updates are completely separate.

akrejcir:
This is my reasoning:

`allow` parameter affects also security updates. So if we set it to only allow `direct` dependencies, then I think indirect will be skipped also for security updates: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow

The trick is to add a second `updates` block with `allow` set to `all`, but  `open-pull-requests-limit: 0`. Security updates ignore this limit, so we should get security updates only for indirect dependencies.

But we have no way to test it.

ksimon1:
I am not sure I understand it correctly. We agreed to bump both direct and indirect dependencies in main branch (even security updates should be introduced for direct and indirect dependencies - remember HTTP2 CVE) and disable it in release branches. What is the benefit from your approach?

akrejcir:
I think bumping indirect dependencies can cause incompatibilities.
But ok, let's wait and see how the bot will behave and if we get any issues.

PR #840, opened 2024-01-17
fix: increase partition1 size in windows efi configmaps

ksimon1:
fix: increase partition1 size in windows efi configmaps
The windows11-autounattend configMap set the partition1 to 300MB, which is the minimum according to documentation. In practice, however, this causes the installer to create an extra partition at the end of the C: partition, blocking it from expanding. Then during windows update the disk fills up. This commit updates it to 700MB.

Fixes https://issues.redhat.com/browse/CNV-36289

**Release note**:
```
fix: increase partition1 size in windows efi configmaps
```


0xFelix (approved):
/lgtm

PR #845, opened 2024-01-18
fix: set boot source labels for DVs created by windows pipelines

ksimon1:
fix: set boot source labels for DVs created by windows pipelines

Set default instanceType and preferences labels to result DVs created by windows pipelines. With this fix, preferences will have default windows boot source when user triggers pipeline.

Fixes https://issues.redhat.com/browse/CNV-34661

**Release note**:
```
fix: set boot source labels for DVs created by windows pipelines
```

### DIFF
--- a/data/tekton-pipelines/windows-bios-installer-pipeline.yaml
+++ b/data/tekton-pipelines/windows-bios-installer-pipeline.yaml
@@ -55,6 +55,11 @@ spec:
             apiVersion: cdi.kubevirt.io/v1beta1
             kind: DataVolume
             metadata:
+              labels:
+                "instancetype.kubevirt.io/default-instancetype-kind": $(params.instanceTypeKind)
+                "instancetype.kubevirt.io/default-instancetype": $(params.instanceTypeName)
+                "instancetype.kubevirt.io/default-preference-kind": $(params.virtualMachinePreferenceKind)
+                "instancetype.kubevirt.io/default-preference": $(params.preferenceName)
               name: $(params.baseDvName)
               namespace: $(params.baseDvNamespace)
             spec:
--- a/data/tekton-pipelines/windows-efi-installer-pipeline.yaml
+++ b/data/tekton-pipelines/windows-efi-installer-pipeline.yaml
@@ -180,6 +180,11 @@ spec:
             metadata:
               annotations:
                 "cdi.kubevirt.io/storage.bind.immediate.requested": "true"
+              labels:
+                "instancetype.kubevirt.io/default-instancetype-kind": $(params.instanceTypeKind)
+                "instancetype.kubevirt.io/default-instancetype": $(params.instanceTypeName)
+                "instancetype.kubevirt.io/default-preference-kind": $(params.virtualMachinePreferenceKind)
+                "instancetype.kubevirt.io/default-preference": $(params.preferenceName)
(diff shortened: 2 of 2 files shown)


data/tekton-pipelines/windows-efi-installer-pipeline.yaml
```diff
@@ -44,6 +44,11 @@ spec:
       description: Name of Windows ISO datavolume
       name: isoDVName
       type: string
+    - default: windows.11.virtio
+      description: Name of the preference which will be used for this DV by default.
+      name: defaultPreferenceLabel
```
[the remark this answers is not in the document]
ksimon1: preferenceName parameter is used for information about VM which is created to install the iso. The defaultPreferenceLabel is used for result DV. These two parameters might be different. E.g. for installation you want to use windows-11-some-fency-feature-enabled, but the result you want to use with different preference (e.g. windows-11-regular-user-disk)

data/tekton-pipelines/windows-bios-installer-pipeline.yaml
```diff
@@ -55,6 +59,9 @@ spec:
             apiVersion: cdi.kubevirt.io/v1beta1
             kind: DataVolume
             metadata:
+              labels:
+                "instancetype.kubevirt.io/default-instancetype": "u1.large"
```
0xFelix, line 63: I'd still vote to not hardcode this.

data/tekton-pipelines/windows-bios-installer-pipeline.yaml
```diff
@@ -55,6 +71,11 @@ spec:
             apiVersion: cdi.kubevirt.io/v1beta1
             kind: DataVolume
             metadata:
+              labels:
```
0xFelix, line 74: What do you think about conditionally adding the kind, only if the param was set? The default will always be clusterwide.

data/tekton-pipelines/windows-bios-installer-pipeline.yaml
```diff
@@ -44,6 +44,22 @@ spec:
...
+      type: string
+    - default: virtualmachineclusterpreference
+      description: Kind of the preference which will be used for this DV by default. Possible values - virtualmachineclusterpreference, virtualmachinepreference
+      name: defaultPreferenceKindLabel
+      type: string
+    - default: u1.large
+      description: Name of the instance type which will be used for this DV by default.
+      name: defaultInstancetypeLabel
+      type: string
+    - default: virtualmachineclusterinstancetype
```
0xFelix, line 59: You could make this an optional param without a default value.

data/tekton-pipelines/windows-efi-installer-pipeline.yaml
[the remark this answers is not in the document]
dominikholler: I agree that the two parameters are technically different, but it might be simple and intuitive to the user, if the DV would reflect the instancetype and preference which were used to create the DV. Avoiding the additional parameter would also keep the "surface" to the user smaller. Since this pipeline targets users who starts learning about kubevirt-tekton-tasks, I would vote to keep it as simple as possible, even at the price of flexibility.

0xFelix (commented):
/approve

Thanks

PR #850, opened 2024-01-22
Update VM controller error handling

machadovilaca:

Sometimes some resources might not exist yet and we see a huge amount of logs stating that an error created the VM metrics.

This PR:

- adds 5 seconds backoff between failed reconcile loops of VM controller to reduce the number of logs 
- removes 'error' wording from logs when the VM controller does not create the VM-related metrics, since it might not represent more serious problems, and might deceive users
- skips metric creation when VM-related resources are still in Provisioning status

jira-ticket: https://issues.redhat.com/browse/CNV-35806

**Release note**:

```release-note
Update VM controller error handling
```

### DIFF
--- a/controllers/vm_controller.go
+++ b/controllers/vm_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"time"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -65,13 +66,26 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 
 			return ctrl.Result{}, nil
 		}
-		r.log.Error(err, "Error getting VM", "vm", req.NamespacedName)
-		return ctrl.Result{}, err
+
+		r.log.Error(err, "Could not find VM", "vm", req.NamespacedName)
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 5 * time.Second,
+		}, err
+	}
+
+	if vm.Status.PrintableStatus == kubevirtv1.VirtualMachineStatusProvisioning {
+		// The status Provisioning is set when not all resources are created yet.
+		// This way we will avoid reconciliation looping while waiting for the resources to be created.
+		return ctrl.Result{}, nil
 	}
 
 	if err := r.setVmVolumesMetrics(ctx, &vm); err != nil {
-		r.log.Error(err, "Error setting vm volumes metrics", "vm", req.NamespacedName)
-		return ctrl.Result{}, err
+		r.log.Error(err, "Could not set vm volumes metrics", "vm", req.NamespacedName)
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 5 * time.Second,
+		}, err
 	}
 
 	return ctrl.Result{}, nil


akrejcir:
If a PVC or PV does not exist, we want to try checking again? Will these resources be created eventually?

machadovilaca:
> If the PVC or PV does not exist, do we want to try checking again? Will these resources be created eventually?

The PVC is associated to the VM, is just not created yet, the backoff ensure we will try the reconcile loop again in a few seconds

akrejcir:
Instead of retrying later, can you check the `.status.printableStatus` field?
The status `Provisioning` is set when not all resources are created yet: 
https://github.com/kubevirt/kubevirt/blob/9d9885c655d465657f515c8f73c23b00c9f16898/staging/src/kubevirt.io/api/core/v1/types.go#L1511-L1513

If a VM has this `printableStatus`, it can be skipped. Later, if the `.status` filed changes, the VM will trigger reconciliation again.
This way we will avoid infinite looping while waiting for the resources to be created.

controllers/vm_controller.go
```diff
@@ -65,13 +66,20 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 5 * time.Second,
+		}, err
 	}
 
 	if err := r.setVmVolumesMetrics(ctx, &vm); err != nil {
-		r.log.Error(err, "Error setting vm volumes metrics", "vm", req.NamespacedName)
-		return ctrl.Result{}, err
+		r.log.Error(err, "could not set vm volumes metrics", "vm", req.NamespacedName)
```
sradco, line 78: Nit: `could` -> `Could`

controllers/vm_controller.go
```diff
@@ -65,13 +66,20 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 
 			return ctrl.Result{}, nil
 		}
-		r.log.Error(err, "Error getting VM", "vm", req.NamespacedName)
-		return ctrl.Result{}, err
+
+		r.log.Error(err, "could not find VM", "vm", req.NamespacedName)
```
sradco, line 70: Nit: `could` -> `Could`

controllers/vm_controller.go
```diff
@@ -65,13 +66,20 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 5 * time.Second,
```
sradco, line 73: @machadovilaca  Why do we wait here 5sec and for the Volumes we wait 30sec?

machadovilaca: waiting time for the PVC to be create, but not needed anymore

machadovilaca:
update @akrejcir

controllers/vm_controller.go
```diff
@@ -65,13 +66,26 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
...
-		return ctrl.Result{}, err
+
+		r.log.Error(err, "Could not find VM", "vm", req.NamespacedName)
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 5 * time.Second,
+		}, err
+	}
+
+	if vm.Status.PrintableStatus != kubevirtv1.VirtualMachineStatusProvisioning {
```
akrejcir, line 77: Typo: Should be `==`.

akrejcir (commented):
Thanks. Can you add more text to the commit message?

machadovilaca:
> Thanks. Can you add more text to the commit message?

done!

PR #854, opened 2024-01-24
feat: update golang to 1.21

ksimon1:
feat: update golang to 1.21

**Release note**:
```
update golang to 1.21
```


.github/workflows/pull-request.yaml
```diff
@@ -8,10 +8,10 @@ jobs:
   validate_metrics_docs:
     runs-on: ubuntu-latest
     steps:
-      - name: Install Golang 1.20
+      - name: Install Golang 1.21
         uses: actions/setup-go@v5
         with:
-          go-version: '1.20'
+          go-version: '1.21'
```
0xFelix, line 14: You could make it use the golang version set in `go.mod`, so we don't have to touch this file in the future.

0xFelix: https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

0xFelix (approved):
/approve
/lgtm

Thanks!

.github/workflows/pull-request.yaml
```diff
@@ -22,10 +22,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - name: Install Golang 1.20
+    - name: Install Golang
       uses: actions/setup-go@v5
       with:
-        go-version: '1.20'
+        go-version-file: 'go.mod'
```
akrejcir, line 28: This action `setup-go` needs to be executed after the below `actions/checkout`, because at this stage the code has not yet been checked out, so `go.mod` file does not exist.

ksimon1:
/approve
Merging without e2e tests, because CI is using old go version from main branch

PR #879, opened 2024-02-09
fix: Remove namespaceSelector from webhook definition

akrejcir:
The `ValidatingWebhookConfiguration` created by OLM can have `namespaceSelector` field set. We don't want it, because the webhook checks that there is only one SSP resource in the cluster.

Jira: https://issues.redhat.com/browse/CNV-36636

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/webhook_controller.go
+++ b/controllers/webhook_controller.go
@@ -1,0 +1,110 @@
+package controllers
+
+import (
+	"context"
+	"slices"
+
+	admissionv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	sspv1beta2 "kubevirt.io/ssp-operator/api/v1beta2"
+)
+
+const (
+	OlmNameLabel      = "olm.webhook-description-generate-name"
+	OlmNameLabelValue = "validation.ssp.kubevirt.io"
+)
+
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;update
+
+// CreateWebhookConfigurationController creates a controller
+// that watches ValidatingWebhookConfiguration created by OLM,
+// and removes any namespaceSelector defined in it.
+//
+// The OLM limits the webhook scope to the namespaces that are defined in the OperatorGroup
+// by setting namespaceSelector in the ValidatingWebhookConfiguration.
+// We would like our webhook to intercept requests from all namespaces.
+//
+// The SSP operator already watches all ValidatingWebhookConfigurations, because
+// of template validator operand, so this controller is not a performance issue.
+func NewWebhookConfigurationController(apiClient client.Client) ControllerReconciler {
+	return &webhookCtrl{
+		apiClient: apiClient,
+	}
+}
+
+type webhookCtrl struct {
+	apiClient client.Client
+}
+
+var _ ControllerReconciler = &webhookCtrl{}
+
+var _ reconcile.Reconciler = &webhookCtrl{}
+
+func (w *webhookCtrl) Start(_ context.Context, mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(w.Name()).
+		For(&admissionv1.ValidatingWebhookConfiguration{}, builder.WithPredicates(
+			predicate.NewPredicateFuncs(hasExpectedLabel),
+		)).
+		Complete(w)
+}
+
+func (w *webhookCtrl) Name() string {
+	return "validating-webhook-controller"
+}
+
+func (w *webhookCtrl) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	webhookConfig := &admissionv1.ValidatingWebhookConfiguration{}
+	if err := w.apiClient.Get(ctx, request.NamespacedName, webhookConfig); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	if !hasExpectedLabel(webhookConfig) {
+		return reconcile.Result{}, nil
+	}
+
+	if changed := updateWebhookConfiguration(webhookConfig); !changed {
+		return reconcile.Result{}, nil
+	}
+
+	err := w.apiClient.Update(ctx, webhookConfig)
+	return reconcile.Result{}, err
+}
+
+func hasExpectedLabel(obj client.Object) bool {
+	return obj.GetLabels()[OlmNameLabel] == OlmNameLabelValue
+}
+
+func updateWebhookConfiguration(webhookConfig *admissionv1.ValidatingWebhookConfiguration) bool {
+	var changed bool
+	for i := range webhookConfig.Webhooks {
+		webhook := &webhookConfig.Webhooks[i]
+		if webhook.NamespaceSelector == nil {
+			continue
+		}
+		// Check if the webhook reacts to SSP resource.
+		var hasSspRule bool
+		for _, rule := range webhook.Rules {
+			if slices.Contains(rule.APIGroups, sspv1beta2.GroupVersion.Group) {
+				hasSspRule = true
+				break
+			}
+		}
+		if !hasSspRule {
+			continue
+		}
+
+		webhook.NamespaceSelector = nil
+		changed = true
+	}
+	return changed
+}
--- a/controllers/wehook_test.go
+++ b/controllers/wehook_test.go
@@ -1,0 +1,120 @@
+package controllers
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	admissionv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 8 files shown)


webhooks/ssp_webhook.go
```diff
@@ -53,6 +66,65 @@ func Setup(mgr ctrl.Manager) error {
 		Complete()
 }
 
+// The OLM limits the webhook scope to the namespaces that are defined in the OperatorGroup
```
0xFelix, line 69: Is that an issue for core KubeVirt too? How do we avoid the creation of multiple KubeVirt CRs when deployed via HCO? Does it use the same approach?

webhooks/ssp_webhook.go
```diff
@@ -53,6 +66,65 @@ func Setup(mgr ctrl.Manager) error {
...
+		}
+
+		patchBytes, err := json.Marshal(patchOperations)
+		if err != nil {
+			return fmt.Errorf("faied to encode json patch for ValidatingWebhookConfiguration: %w", err)
+		}
+
+		patch := client.RawPatch(types.JSONPatchType, patchBytes)
+		err = mgr.GetClient().Patch(ctx, webhookConfig, patch)
+		if err != nil {
```
0xFelix, line 121: Inline check?

webhooks/ssp_webhook.go
```diff
@@ -53,6 +66,65 @@ func Setup(mgr ctrl.Manager) error {
...
+// The OLM limits the webhook scope to the namespaces that are defined in the OperatorGroup
+// by setting namespaceSelector in the ValidatingWebhookConfiguration. We would like our webhook to intercept
+// requests from all namespaces. Luckily the OLM does not watch and reconcile the ValidatingWebhookConfiguration,
+// so we can simply reset the namespaceSelector
+func patchValidatingWebhookConfiguration(ctx context.Context, mgr ctrl.Manager) error {
+	configList := &admissionv1.ValidatingWebhookConfigurationList{}
+	err := mgr.GetAPIReader().List(ctx, configList,
+		client.MatchingLabels{"olm.webhook-description-generate-name": "validation.ssp.kubevirt.io"},
+	)
+	if err != nil {
```
0xFelix, line 78: Inline check?

webhooks/ssp_webhook.go
```diff
@@ -37,9 +42,17 @@ import (
...
 )
 
+// RBAC Rule needed by webhook
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=list;patch
+
 var ssplog = logf.Log.WithName("ssp-resource")
 
-func Setup(mgr ctrl.Manager) error {
+func Setup(ctx context.Context, mgr ctrl.Manager) error {
+	err := patchValidatingWebhookConfiguration(ctx, mgr)
```
0xFelix, line 51: Is this prone to flakiness? Is it possible that the webhook server is deployed and running before the `ValidatingWebhookConfiguration` was created?

tests/webhook_test.go
```diff
@@ -206,4 +211,121 @@ var _ = Describe("Validation webhook", func() {
...
+			Expect(apiClient.Create(ctx, webhookConfig)).To(Succeed())
+			testWebhookConfigName = webhookConfig.Name
+
+			DeferCleanup(func() {
+				err := apiClient.Delete(ctx, webhookConfig)
+				if !errors.IsNotFound(err) {
+					Expect(err).ToNot(HaveOccurred())
+				}
+				testWebhookConfigName = ""
+			})
```
0xFelix, line 260: Is it possible to use AfterEach instead? (to avoid the nesting)

webhooks/ssp_webhook.go
akrejcir: It is not. SSP `ValidatingWebhookConfiguration` is created by OLM. The webhook for `Kubevirt` resource is created by `virt-operator` and it only checks deletion, not create or update.

webhooks/ssp_webhook.go
akrejcir: Possibly. I will change the code to use a controller and watch the configuration.

akrejcir:
I've changed the logic completely. It uses a controller now.
@0xFelix @jcanocan can you take another look?

tests/webhook_controller_test.go
```diff
@@ -0,0 +1,89 @@
...
+				AdmissionReviewVersions: []string{"v1"},
+			}},
+		}
+	})
+
+	It("[test_id:TODO] should delete namespaceSelector from webhook configuration with OLM label", func() {
+		Expect(apiClient.Create(ctx, webhook)).To(Succeed())
+		DeferCleanup(func() {
+			Expect(apiClient.Delete(ctx, webhook)).To(Succeed())
+		})
```
jcanocan, line 64: As far as I understand, the intention here is that these `DeferCleanup` function calls are executed before the global `AfterEach`:

https://github.com/kubevirt/ssp-operator/blob/eb60b158ff1d5914f6ec3cfdd32d51b5e884f4e3/tests/cleanup_test.go#L29-L31

Could you verify that this is what actually happens?  As mentioned before, these calls have an issue which makes them to be executed after the global `AfterEach`:  https://github.com/onsi/ginkgo/issues/1284#issuecomment-1756314394

akrejcir: That issue is probably not relevant to this specific test file, because it does not use `AfterEach()`.

I will read `ginkgo` source to be sure how the ordering is done.

Btw, what do you mean by "global `AfterEach()`" ? Do you mean the `AfterSuite()` function:
https://github.com/kubevirt/ssp-operator/blob/b1770163ba65c1327926a47de79c6815c3a75eb6/tests/tests_suite_test.go#L419-L421

jcanocan: Sorry for the noise, SSP does not have such `AfterEach` executed after each test case as the `KubeVirt` testsuite. So in this case, this shouldn't be a problem.

0xFelix: In this case it is most likely not a problem, but a global awareness of the caveats of `DeferCleanup` is certainly helpful.

controllers/webhook_controller.go
```diff
@@ -0,0 +1,107 @@
...
+	OlmNameLabelValue = "validation.ssp.kubevirt.io"
+)
+
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingwebhookconfigurations,verbs=get;update
+
+// CreateWebhookConfigurationController creates a controller
+// that watches ValidatingWebhookConfiguration created by OLM,
+// and removes any namespaceSelector defined in it.
+//
+// The OLM limits the webhook scope to the namespaces that are defined in the OperatorGroup
```
0xFelix, line 29: Btw. is this a bug than should be opened on OLM?

controllers/webhook_controller.go
```diff
@@ -0,0 +1,107 @@
...
+				break
+			}
+		}
+		if !hasSspRule {
+			continue
+		}
+
+		webhook.NamespaceSelector = nil
+		changed = true
+	}
```
0xFelix, line 95: Extract this to func?

controllers/wehook_test.go
```diff
@@ -0,0 +1,120 @@
...
+		Expect(fakeClient.Create(context.Background(), webhookConfig)).To(Succeed())
+
+		_, err := testController.Reconcile(context.Background(), testRequest)
+		Expect(err).ToNot(HaveOccurred())
+
+		updatedConfig := &admissionv1.ValidatingWebhookConfiguration{}
+		Expect(fakeClient.Get(context.Background(), client.ObjectKeyFromObject(webhookConfig), updatedConfig)).To(Succeed())
+
+		Expect(updatedConfig).To(Equal(webhookConfig))
+	})
```
0xFelix, line 119: Maye put these two into a table?

controllers/webhook_controller.go
akrejcir: It is not an OLM bug. Our downstream CSV manifes has intentionally set `installModes` filed to `OwnNamespace` and `SingleNamespace`. We want to limit the scope of our operators.

SSP webhook is special compared to regular webhooks, that it should also prevent the creation of SSP CRs in other namespaces, so it needs to be triggered if these objects are created in different namespace.

akrejcir: I don't know. It looks like a design choice that should be discussed with OLM team.

controllers/wehook_test.go
akrejcir: I'd prefer not to. It is only 2 tests. If there were more similar test cases, then I would do it.

controllers/webhook_controller.go
akrejcir: done

controllers/webhook_controller.go
```diff
@@ -0,0 +1,110 @@
...
+		return reconcile.Result{}, err
+	}
+
+	if !hasExpectedLabel(webhookConfig) {
+		return reconcile.Result{}, nil
+	}
+
+	if changed := updateWebhookConfiguration(webhookConfig); !changed {
+		return reconcile.Result{}, nil
+	}
```
0xFelix, line 77: nit

```suggestion
	if !updateWebhookConfiguration(webhookConfig) {
		return reconcile.Result{}, nil
	}
```

controllers/webhook_controller.go
0xFelix: Would you be able to investigate it? Do other repos in the kubevirt org have the same issue (e.g. HCO)?

0xFelix (commented):
/approve

controllers/webhook_controller.go
akrejcir: HCO has the same issue. My previous revision of this PR was copied from: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/7ac82e83ad975bf369f7ad1425936047ae892eba/pkg/webhooks/setup.go#L71

Kubevirt does not have a webhook that checks `Kubevirt` resource, and I'm not sure about CDI.

controllers/webhook_controller.go
akrejcir: This is really small. I don't want to retrigger CI because of it.

akrejcir:
@jcanocan can you please take another look?

PR #870, opened 2024-02-01
Refactor recording rules and alerts code

avlitman:
**What this PR does / why we need it**:
Following the work started in https://github.com/kubevirt/kubevirt/pull/10044 , and according to the https://github.com/kubevirt/community/pull/219 proposal, this PR refactors monitoring recording rules and alerts

**Which issue(s) this PR fixes**: 

Fixes # 
jira-ticket: https://issues.redhat.com/browse/CNV-37376
jira-ticket: https://issues.redhat.com/browse/CNV-34290
**Special notes for your reviewer**:

**Release note**:

```
Refactor recording-rules and alerts code.

```

PR #896, opened 2024-02-26
Re-add cnv:vmi_status_running:count recording rule

machadovilaca:

Re-add `cnv:vmi_status_running:count,` incorrectly removed in https://github.com/kubevirt/ssp-operator/pull/870

Fixes #

**Release note**:

```release-note
Re-add cnv:vmi_status_running:count recording rule
```

### DIFF
--- a/pkg/monitoring/rules/recordingrules/recordingrules.go
+++ b/pkg/monitoring/rules/recordingrules/recordingrules.go
@@ -5,5 +5,6 @@ import "github.com/machadovilaca/operator-observability/pkg/operatorrules"
 func Register() error {
 	return operatorrules.RegisterRecordingRules(
 		operatorRecordingRules(),
+		vmiRecordingRules,
 	)
 }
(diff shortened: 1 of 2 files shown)


docs/metrics.md
```diff
@@ -1,5 +1,8 @@
 # SSP Operator metrics
 
+### cnv:vmi_status_running:count
```
ksimon1, line 3: is the name correct?

akrejcir:
The metrics linter fails with error message:
```
cnv:vmi_status_running:count: metric names should not contain ':'
cnv:vmi_status_running:count: name need to start with 'kubevirt_'
```

docs/metrics.md
machadovilaca: yes, it does follow the conventions we are using now, but its a old metric, so we would need a deprecation system

machadovilaca: yes, its a old name, to change it we would need a deprecation policy

PR #925, opened 2024-03-14
Replace test 'Eventually-Consistently' with 'Eventually-MustPassRepeatedly

machadovilaca:

address failures such as https://main-jenkins-csb-cnvqe.apps.ocp-c1.prod.psi.redhat.com/job/test-ssp-cnv-4.14/

Fixes #

**Release note**:

```release-note
none
```

### DIFF
--- a/tests/monitoring_test.go
+++ b/tests/monitoring_test.go
@@ -331,18 +331,7 @@ func alertShouldNotBeActive(alertName string) {
 			return nil
 		}
 		return fmt.Errorf("alert %s found", alertName)
-	}, env.Timeout(), time.Second).ShouldNot(HaveOccurred())
-
-	ConsistentlyWithOffset(1, func() error {
-		alert, err := checkAlert(alertName)
-		if err != nil {
-			return err
-		}
-		if alert == nil || alert.State == "inactive" {
-			return nil
-		}
-		return fmt.Errorf("alert %s found", alertName)
-	}, env.ShortTimeout(), time.Second).ShouldNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


tests/monitoring_test.go
```diff
@@ -331,18 +331,7 @@ func alertShouldNotBeActive(alertName string) {
...
-		alert, err := checkAlert(alertName)
-		if err != nil {
-			return err
-		}
-		if alert == nil || alert.State == "inactive" {
-			return nil
-		}
-		return fmt.Errorf("alert %s found", alertName)
-	}, env.ShortTimeout(), time.Second).ShouldNot(HaveOccurred())
+	}, env.Timeout(), 10*time.Second).MustPassRepeatedly(10).ShouldNot(HaveOccurred())
```
akrejcir, line 334: The shortest amount of time for this to pass is 100 seconds. Is that acceptable? Can the time be lower?

machadovilaca: I was actually not sure what an acceptable middle term would be, do you think 5 times would be fine? 3 is too low because Prometheus is scrapping metrics just every 30s seconds

akrejcir: I'm also not sure. I don't have enough knowledge about prometheus. If you think a lower number would be too unstable, than this is ok.

machadovilaca: lowering the time could cause a similar problem to what we had before, of some state in-between causing flakiness

akrejcir: Ok, thanks for explanation.

PR #929, opened 2024-03-15
feat: Don't deploy tekton tasks and pipelines

akrejcir:
**This is an incomplete draft.** Currently only `tekton-pipelines` operand was changed. I will add more commits.

The tekton tasks and pipelines can now be used directly from Artifacthub. 
The tasks and pipelines previously installed by SSP are now deprecated. The operator will not remove them, but it will add an annotation to signal deprecation, and it will stop watching them.

Fixes: https://issues.redhat.com/browse/CNV-38642
Fixes: https://issues.redhat.com/browse/CNV-35909
Fixes: https://issues.redhat.com/browse/CNV-35910
Fixes: https://issues.redhat.com/browse/CNV-36129

**Release note**:
```release-note
Tekton tasks and pipelines are no longer installed by this operator.
```


akrejcir:
This is now ready for review.

/cc @ksimon1 @0xFelix @jcanocan

api/v1beta2/ssp_types.go
```diff
@@ -85,9 +85,13 @@ type SSPSpec struct {
 	CommonInstancetypes *CommonInstancetypes `json:"commonInstancetypes,omitempty"`
 
 	// TektonPipelines is the configuration of the tekton-pipelines operand
+	// +kubebuilder:deprecatedversion:warning="tektonPipelines filed is ignored"
+	// Deprecated: This filed is ignored.
```
0xFelix, line 89:
```suggestion
	// Deprecated: This field is ignored.
```

api/v1beta2/ssp_types.go
```diff
@@ -85,9 +85,13 @@ type SSPSpec struct {
...
 	TektonPipelines *TektonPipelines `json:"tektonPipelines,omitempty"`
 
 	// TektonTasks is the configuration of the tekton-tasks operand
+	// +kubebuilder:deprecatedversion:warning="tektonTasks filed is ignored"
+	// Deprecated: This filed is ignored.
```
0xFelix, line 94:
```suggestion
	// Deprecated: This field is ignored.
```

api/v1beta2/ssp_types.go
```diff
@@ -106,6 +110,8 @@ type TektonTasks struct {
 
 // FeatureGates for SSP
 type FeatureGates struct {
+	// +kubebuilder:deprecatedversion:warning="tekton task resources are no longer deployed by SSP"
+	// Deprecated: This filed is ignored.
```
0xFelix, line 114:
```suggestion
	// Deprecated: This field is ignored.
```

config/rbac/role.yaml
```diff
@@ -101,6 +101,15 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
0xFelix, line 104: Why do we add new RBAC on removal of a feature?

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,175 @@
...
+	*T
+	client.Object
+}](request *common.Request) error {
+	resources, err := common.ListOwnedResources[L, T, PtrL, PtrT](request, client.MatchingLabelsSelector{Selector: labels.NewSelector().Add(
+		newLabelRequirementOrPanic(common.AppKubernetesManagedByLabel, selection.Equals, []string{common.AppKubernetesManagedByValue}),
+		newLabelRequirementOrPanic(common.AppKubernetesComponentLabel, selection.In, []string{common.AppComponentTektonPipelines.String(), common.AppComponentTektonTasks.String()}),
+		newLabelRequirementOrPanic(common.AppKubernetesNameLabel, selection.In, []string{operandPipelinesName, operandTasksName}),
+	)})
+	if err != nil {
+		return fmt.Errorf("faield to list owned resources: %w", err)
```
0xFelix, line 153:
```suggestion
		return fmt.Errorf("failed to list owned resources: %w", err)
```

internal/common/owner.go
```diff
@@ -26,3 +28,36 @@ func CheckOwnerAnnotation(obj client.Object, owner client.Object) bool {
...
 	return ok && namespacedNameVal == sspNamespacedName.String()
 }
+
+func ListOwnedResources[L any, T any, PtrL interface {
+	*L
+	client.ObjectList
+}, PtrT interface {
+	*T
+	client.Object
+}](request *Request, listOpts ...client.ListOption) ([]T, error) {
```
0xFelix, line 38: This reads very hard.

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,175 @@
...
+
+func cleanupResources(request *common.Request) error {
+	cleanupFuncs := []func(*common.Request) error{
+		cleanupResource[rbac.ClusterRoleList, rbac.ClusterRole],
+		cleanupResource[rbac.RoleBindingList, rbac.RoleBinding],
+		cleanupResource[v1.ServiceAccountList, v1.ServiceAccount],
+		cleanupResource[v1.ConfigMapList, v1.ConfigMap],
+	}
+	if request.CrdList.CrdExists(tektonCrd) {
+		cleanupFuncs = append(cleanupFuncs, cleanupResource[pipeline.PipelineList, pipeline.Pipeline]) //nolint:staticcheck
```
0xFelix, line 127: What does the linter complain about here?

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,175 @@
...
+	return nil
+}
+
+func cleanupResource[L any, T any, PtrL interface {
+	*L
+	client.ObjectList
+}, PtrT interface {
+	*T
+	client.Object
+}](request *common.Request) error {
```
0xFelix, line 146: Again, not the easiest signature to understand.

0xFelix (commented):
Do I understand correctly, that this is intended to delete resources deployed by SSP or TTO in the past? Is it not okay to leave them behind with a deprecation notice as we did with other resources in the past?

akrejcir:
> Do I understand correctly, that this is intended to delete resources deployed by SSP or TTO in the past? Is it not okay to leave them behind with a deprecation notice as we did with other resources in the past?

No, this PR does not remove existing resources. It leaves them there, adds a deprecation annotation, and removes the owner annotations. Then SSP will no longer watch those resources.

config/rbac/role.yaml
akrejcir: It's just reshuffling of existing permissions. I will look into it.

internal/operands/tekton-cleanup/reconcile.go
akrejcir: `pipeline.Pipeline` is deprecated.

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,175 @@
...
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=list;watch;create;update
+
+// Below are RBAC for deployed ClusterRoles. We still need these permissions so we can update annotations on existing ClusterRoles
+
+// RBAC rules for "cleanup-vm-task"
+// +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachineinstances;virtualmachines,verbs=get;list;watch;delete
+// +kubebuilder:rbac:groups=subresources.kubevirt.io,resources=virtualmachines/restart;virtualmachines/start;virtualmachines/stop,verbs=update
+
+// RBAC rules for "copy-template-task"
+// +kubebuilder:rbac:groups=template.openshift.io,resources=templates,verbs=get;list;watch;create;update
```
ksimon1, line 32: Are all of these rbacs necessary? Since operator will not deploy anything and just udate each object.

akrejcir: Yes, I've tried it. When the operator updates a `ClusterRole` it needs to have all the RBAC permissions that are in the `ClusterRole`.

akrejcir: I will try to use `Patch` to only change the annotations.

internal/operands/tekton-cleanup/reconcile.go
akrejcir: done

api/v1beta2/ssp_types.go
akrejcir: done

api/v1beta2/ssp_types.go
akrejcir: done

config/rbac/role.yaml
akrejcir: done

internal/operands/tekton-cleanup/reconcile.go
akrejcir: I've tried using `Patch` instead of `Update` but the RBAC permissions are still required.

akrejcir: I've tried using `Patch` instead of `Update` but the RBAC permissions are still required.

akrejcir:
I will change it, so old resources are removed when SSP is removed.

/hold

akrejcir:
I've changed the code. Now the old tekton resources are deleted when the SSP CR is deleted.

@ksimon1 , @0xFelix , can you please take another look?

/unhold

internal/common/owner.go
0xFelix: Better leave it like it is, a single line would make it worse IMO. Also, let's stay away from reflection if we can. :+1:

internal/operands/tekton-cleanup/reconcile.go
0xFelix: ACK

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,195 @@
...
+// +kubebuilder:rbac:groups=template.openshift.io,resources=templates,verbs=get;list;watch;create;update;delete
+
+const (
+	operandName          = "tekton-cleanup"
+	operandPipelinesName = "tekton-pipelines"
+	operandTasksName     = "tekton-tasks"
+
+	tektonCrd = "tasks.tekton.dev"
+
+	tektonDeprecated = "tekton.dev/deprecated"
```
0xFelix, line 44: Is that a label used by Tekton? Or should it reflect that those were part of KubeVirt, e.g. `ssp.kubevirt.io/deprecated`?

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,195 @@
...
+		selector.Add(newLabelRequirementOrPanic(common.AppKubernetesPartOfLabel, selection.Equals, []string{sspPartOf}))
+	}
+
+	return client.MatchingLabelsSelector{Selector: selector}
+}
+
+func newLabelRequirementOrPanic(key string, op selection.Operator, vals []string) labels.Requirement {
+	requirement, err := labels.NewRequirement(key, op, vals)
+	if err != nil {
+		panic(fmt.Errorf("invalid requirement: %w", err))
```
0xFelix, line 192: Shouldn't it return an `err` instead of panicking?

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,195 @@
...
+
+// +kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;update;delete
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=create
+// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;create;patch
+// +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachineinstances;virtualmachines,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=kubevirt.io,resources=virtualmachines/finalizers,verbs=get
+// +kubebuilder:rbac:groups=cdi.kubevirt.io,resources=datasources,verbs=get;create;delete
+// +kubebuilder:rbac:groups=subresources.kubevirt.io,resources=virtualmachines/restart;virtualmachines/start;virtualmachines/stop,verbs=update
+// +kubebuilder:rbac:groups=template.openshift.io,resources=processedtemplates,verbs=create
+// +kubebuilder:rbac:groups=template.openshift.io,resources=templates,verbs=get;list;watch;create;update;delete
```
lyarwood, line 35: Most of these can be removed now right?

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func matchingLabelsOption(ssp *ssp.SSP) client.MatchingLabelsSelector {
+	selector := labels.NewSelector().Add(
+		newLabelRequirementOrPanic(common.AppKubernetesManagedByLabel, selection.Equals, []string{common.AppKubernetesManagedByValue}),
+		newLabelRequirementOrPanic(common.AppKubernetesComponentLabel, selection.In, []string{common.AppComponentTektonPipelines.String(), common.AppComponentTektonTasks.String()}),
+		newLabelRequirementOrPanic(common.AppKubernetesNameLabel, selection.In, []string{operandPipelinesName, operandTasksName}),
+	)
+
+	if sspPartOf, exists := ssp.Labels[common.AppKubernetesPartOfLabel]; exists {
```
lyarwood, line 182: Remind me again why this is optional, do we not add it upstream?

internal/operands/tekton-cleanup/reconcile.go
```diff
@@ -0,0 +1,195 @@
...
+// +kubebuilder:rbac:groups=template.openshift.io,resources=processedtemplates,verbs=create
+// +kubebuilder:rbac:groups=template.openshift.io,resources=templates,verbs=get;list;watch;create;update;delete
+
+const (
+	operandName          = "tekton-cleanup"
+	operandPipelinesName = "tekton-pipelines"
+	operandTasksName     = "tekton-tasks"
+
+	tektonCrd = "tasks.tekton.dev"
+
```
lyarwood, line 43: supernit - whitespace between the two `tekton*` constants

lyarwood (commented):
This looks good as an approach, some quick comments about the leftover RBAC rules and a few nits but otherwise looks great.

internal/operands/tekton-cleanup/reconcile.go
ksimon1: This is Tekton's labels, not KubeVirt's.

akrejcir: I think it is used by tekton, or the UI. The annotation was mentioned in this jira task: https://issues.redhat.com/browse/CNV-36129

@ksimon1, is it correct?

internal/operands/tekton-cleanup/reconcile.go
akrejcir: I've added this helper function, because it is always called with arguments that are compile time constants. Any error returned by `labels.NewRequirement()` is a programmer's mistake. In my opinion it is ok to panic in this case.

internal/operands/tekton-cleanup/reconcile.go
akrejcir: No, these rules are still needed, because the current code updates an existing `ClusterRole` objects.
The operator needs to have all the permissions that are in the roles to be allowed to update them.

internal/operands/tekton-cleanup/reconcile.go
akrejcir: HCO adds it upstream. Keeping it optional is more convenient in development environments, so we can create SSP object without it.

In my opinion requiring a label to be present on an object is not a good design. It should be in `.spec` if it contains required data.

internal/common/owner.go
akrejcir: Ok, I will leave it as is for now. 
If you will have any idea how to improve it in the future, please tell me. I don't like that golang requires separate generic parameters for `T` and pointer to `T`.

internal/operands/tekton-cleanup/reconcile.go
akrejcir: You were faster, thank you.

codingben:
Please also update docs of `kubevirt/hyperconverged-cluster-operator` [1] and [2].

[1] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/api.md
[2] https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#deploytektontaskresources-feature-gate

0xFelix (commented):
/approve

ksimon1:
Can you please remove this file https://github.com/kubevirt/ssp-operator/blob/main/.github/workflows/update-virtio-version.yaml too?

akrejcir:
> Can you please remove this file https://github.com/kubevirt/ssp-operator/blob/main/.github/workflows/update-virtio-version.yaml too? Can be done in follow up PR

I've opened a new PR: https://github.com/kubevirt/ssp-operator/pull/939

PR #950, opened 2024-04-09
fix: Increase reconcile metric only if SSP CR was not changed

akrejcir:
The metric `kubevirt_ssp_common_templates_restored_total` should be increased if user changes a template and the operator restores it back to what it should be.

If the SSP CR is changed, for example one of the labels that are propagated to the templates are changed, then the reconciliation should not increase the metric.

Fixes: https://issues.redhat.com/browse/CNV-39928

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/ssp_controller.go
+++ b/controllers/ssp_controller.go
@@ -149,18 +149,20 @@ func (r *sspReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 		// Error reading the object - requeue the request.
 		return ctrl.Result{}, err
 	}
-	r.clearCacheIfNeeded(instance)
+
+	sspChanged := r.clearCacheIfNeeded(instance)
 
 	sspRequest := &common.Request{
-		Request:        req,
-		Client:         r.client,
-		UncachedReader: r.uncachedReader,
-		Context:        ctx,
-		Instance:       instance,
-		Logger:         reqLogger,
-		VersionCache:   r.subresourceCache,
-		TopologyMode:   r.topologyMode,
-		CrdList:        r.crdList,
+		Request:         req,
+		Client:          r.client,
+		UncachedReader:  r.uncachedReader,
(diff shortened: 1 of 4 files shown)


0xFelix (commented):
/approve

Looks good to me. But allow me a question, is there really a benefit to exposing this metric at all?

sradco:
> 
> /approve
> 
> Looks good to me. But allow me a question, is there really a benefit to exposing this metric at all?
> 
>> I think it is a question for @machadovilaca or @sradco .

This metric is used for an alert.

0xFelix:
> > /approve
> > Looks good to me. But allow me a question, is there really a benefit to exposing this metric at all?
> > > I think it is a question for @machadovilaca or @sradco .
> 
> This metric is used for an alert.

Let me rephrase, is there a benefit to having that alert?

sradco:
> /approve
> Looks good to me. But allow me a question, is there really a benefit to exposing this metric at all?
> 
> 
> I think it is a question for @machadovilaca or @sradco .
> > 
> >> This metric is used for an alert.
> >> 
> > >>Let me rephrase, is there a benefit to having that alert?

This is not for me or @machadovilaca to decide.
It up to your team.

From what I understand the alert is educational, But I see it has "warning" severity. 
Is that the correct severity for it? 

I think it should be "info" if you keep it, since it doesn't impact the SSP operator functionality or even puts it at risk for losing functionality. https://github.com/kubevirt/monitoring/blob/main/docs/runbooks/SSPCommonTemplatesModificationReverted.md

0xFelix:
@sradco Thanks for the clarification. I don't see a benefit in this alert to be honest. We should reevaluate the need for it. Also, you are right, it should be of severity info.

akrejcir:
I've opened a new issue for us to reevaluate which metrics are needed: https://github.com/kubevirt/ssp-operator/issues/956

PR #952, opened 2024-04-11
Fix common-instancetypes func tests with HCO

lyarwood:
/cc @0xFelix 

The functional tests currently assume that SSP is deployed standalone without HCO. However these same tests are also used downstream in a slightly different context where we initially deploy the environment using HCO. As such the common-instancetypes tests are invalid and need to be skipped. This series adds some limited coverage for this scenario ensuring that SSP ignores the virt-operator deployed resources.

More comprehensive coverage needs to be added to HCO to assert the default behaviour of virt-operator and ssp-operator in environments where HCO is still actively reconciling both.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/common_instancetypes_test.go
+++ b/tests/common_instancetypes_test.go
@@ -132,6 +132,39 @@ var _ = Describe("Common Instance Types", func() {
 			}
 		})
 	})
+
+	Context("when SSP is deployed by HCO", func() {
+
+		BeforeEach(func() {
+			strategy.SkipIfNotDeployedByHCO()
+		})
+
+		It("should not claim ownership or remove existing resources deployed by virt-operator", func() {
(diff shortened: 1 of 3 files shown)


0xFelix (commented):
Not sure this is the right approach. Can't we make sure to set the appropriate FGs during the tests and reset them to their original value after the tests ran? The functionality is there and a user could still choose to enable it by modifying the FGs with jsonpatch annotations on HCO's CR.

Of course this won't work if HCO keeps reconciling them, but I don't think the test is at fault when HCO interferes with it.

lyarwood:
> Not sure this is the right approach. Can't we make sure to set the appropriate FGs during the tests and reset them to their original value after the tests ran? The functionality is there and a user could still choose to enable it by modifying the FGs with jsonpatch annotations on HCO's CR.
> 
> Of course this won't work if HCO keeps reconciling them, but I don't think the test is at fault when HCO interferes with it.

Yeah fair point, with HCO disabled we can just disable the KV FG for virt-operator deployment in a `BeforeAll` and restore in `AfterAll`. That should allow these tests to work regardless of HCO being involved in the initial deployment. I guess we should also wait for virt-operator to clean up these resources? Let me rework this.

/hold

PR #960, opened 2024-04-15
[release-v0.19] fix: Increase reconcile metric only if SSP CR was not changed

akrejcir:
Manual cherry-pick of: https://github.com/kubevirt/ssp-operator/pull/950
It was needed, because metrics code was refactored in `main`.

The metric "kubevirt_ssp_common_templates_restored_total" should be increased if user changes a template and the operator restores it back to what it should be.

If the SSP CR is changed, for example one of the labels that are propagated to the templates are changed,
then the reconciliation should not increase the metric.

Fixes: https://issues.redhat.com/browse/CNV-39928

**Release note**:
```release-note
None
```


0xFelix (approved):
/approve

Thanks!

PR #986, opened 2024-05-21
common instancetypes: Deprecate operand 

lyarwood:

Deployment of the `common-instancetypes` resources will be handled by `virt-operator` moving forward so we can finally deprecate the operand within SSP for removal in a future release.

Fixes #

**Release note**:

```release-note
Deployment of `common-instancetypes` using this operator is now deprecated and will be removed in a future release.
```


akrejcir:
I think you need to run
```
make fmt vet container-build
```
that should regenerate need files.

0xFelix (approved):
/approve

PR #987, opened 2024-05-23
refactor: Prepare controller code for watching single namespace

akrejcir:
This is a refactoring PR before limiting the operator to watching SSPs only from a single namespace.
The main goal of this is to create controllers before creating the Manager. Then in a future PR, each controller will expose information needed to configure cache to only watch certain namespaces or labels. It will improve memory load of the operator.

Jira: https://issues.redhat.com/browse/CNV-38130

**Release note**:
```release-note
None
```


codingben:
> It will improve memory load of the operator.

Do you have some data about it, how much it improved memory load of the operator? What it was before, and will be after this PR is merged?

akrejcir:
> > It will improve memory load of the operator.
> 
> Do you have some data about it, how much it improved memory load of the operator? What it was before, and will be after this PR is merged?

I don't have any data, but it makes sense. Currently the operator is watching all objects of certain type that are in the cluster, like `Deployments`, `ServiceAccounts`, `Services`, etc. Even those that are created by users and not related to CNV.
This means that memory requirements of the operator scale with the cluster size, which is not needed.

The only data I have is in the linked jira ticket: https://issues.redhat.com/browse/CNV-38130
The graph shows that after creating 2000 user workloads (Deployments, Services, ...) the memory load is around 190 MB.

akrejcir:
This PR is only a refactoring. I'm still working on the change that will limit which objects are cached.

internal/controllers/setup.go
```diff
@@ -82,23 +81,39 @@ func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controller
...
 	}
 
-	// Add VMController necessary VirtualMachine CRD
-	vmKind := strings.ToLower(kubevirtv1.VirtualMachineGroupVersionKind.Kind) + "s"
-	vmCRD := vmKind + "." + kubevirtv1.VirtualMachineGroupVersionKind.Group
-	requiredCrds = append(requiredCrds, vmCRD)
+	return []Controller{
+		serviceController,
+		NewWebhookConfigurationController(),
+		CreateVmController(),
```
0xFelix, line 100: `CreateVmController` -> `NewVmController` IMO is a nice refactoring to be consistent. (maybe also in a follow up)

internal/controllers/ssp_controller.go
```diff
@@ -68,42 +69,50 @@ var kvsspCRDs = map[string]string{
...
 	operands         []operands.Operand
 	lastSspSpec      ssp.SSPSpec
 	subresourceCache common.VersionCache
 	topologyMode     osconfv1.TopologyMode
-	crdList          crd_watch.CrdList
 	areCrdsMissing   bool
+
+	client         client.Client
+	uncachedReader client.Reader
+	crdList        crd_watch.CrdList
```
0xFelix, line 83: Sometimes vars of this type are called `crdWatch`, sometimes `crdList`. Stick to one to be consistent?

internal/controllers/ssp_controller.go
```diff
@@ -113,32 +122,55 @@ func (r *sspReconciler) setupController(mgr ctrl.Manager) error {
 	builder := ctrl.NewControllerManagedBy(mgr)
 	watchSspResource(builder)
 
-	r.areCrdsMissing = len(r.crdList.MissingCrds()) > 0
+	s.areCrdsMissing = len(s.crdList.MissingCrds()) > 0
```
0xFelix, line 125: This could be missing CRDs of other Controllers too, right?

internal/controllers/vm_controller.go
```diff
@@ -91,7 +97,12 @@ func (r *VmReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 	return ctrl.Result{}, nil
 }
 
-func (r *VmReconciler) setVmVolumesMetrics(ctx context.Context, vm *kubevirtv1.VirtualMachine) error {
+func getVmCrd() string {
+	vmKind := strings.ToLower(kubevirtv1.VirtualMachineGroupVersionKind.Kind) + "s"
```
0xFelix, line 101: Same as `kubevirtv1.VirtualMachineGroupVersionKind.GroupKind().ToString()`?

internal/env/env.go
```diff
@@ -22,6 +20,8 @@ const (
 	TemplateValidatorImageKey = "VALIDATOR_IMAGE"
 	VmConsoleProxyImageKey    = "VM_CONSOLE_PROXY_IMAGE"
```
0xFelix, line 22: Drop newline between?

internal/env/env.go
```diff
@@ -63,12 +63,10 @@ func GetInfrastructureTopology(ctx context.Context, c client.Reader) (osconfv1.T
 	return infraConfig.Status.InfrastructureTopology, nil
 }
 
-func GetOperatorNamespace(logger logr.Logger) (string, error) {
-	nsBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
0xFelix, line 67: Why did you choose to remove reading the namespace from the current service account?

I think it's cleaner not to rely on an additional env variable, that needs to be set to some value by someone external.

internal/operands/vm-console-proxy/reconcile_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"kubevirt.io/ssp-operator/internal/env"
```
0xFelix, line 10: Move import to L30

main.go
```diff
@@ -25,13 +25,14 @@ import (
...
 	"path"
 
-	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
-	// to ensure that exec-entrypoint and run can make use of them.
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
 	ocpconfigv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/library-go/pkg/crypto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
0xFelix, line 31: Move this import to block below.

tests/service_controller_test.go
```diff
@@ -7,14 +7,15 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	env2 "kubevirt.io/ssp-operator/internal/env"
```
0xFelix, line 10: Move to bottom import block

0xFelix: Maybe call it `ssp_env`.

internal/env/env.go
0xFelix: Ah, I missed the downward API part. So it comes down to preference. Actually, I think both approaches are fine.

internal/controllers/ssp_controller.go
akrejcir: Yes, thanks for noticing. I will open a separate issue to fix this. 
It does not break anything, but the resulting error message can be confusing.

internal/controllers/vm_controller.go
akrejcir: No, because there is the `s` at the end for plural. The CRD is called:
```
virtualmachines.kubevirt.io
```

And `kubevirtv1.VirtualMachineGroupVersionKind.GroupKind().ToString()` returns:
```
virtualmachine.kubevirt.io
```

tests/service_controller_test.go
akrejcir: done

main.go
akrejcir: done

internal/operands/vm-console-proxy/reconcile_test.go
akrejcir: done

internal/env/env.go
akrejcir: The newline separates exported and private constants.

internal/controllers/setup.go
akrejcir: done

internal/controllers/setup.go
```diff
@@ -45,22 +44,22 @@ func CreateAndStartReconciler(ctx context.Context, mgr controllerruntime.Manager
 	return nil
 }
 
-func setupManager(ctx context.Context, cancel context.CancelFunc, mgr controllerruntime.Manager) error {
-	runningOnOpenShift, err := common.RunningOnOpenshift(ctx, mgr.GetAPIReader())
+func CreateControllres(ctx context.Context, apiReader client.Reader) ([]Controller, error) {
```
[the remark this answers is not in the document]
akrejcir: thanks

internal/controllers/services_controller.go
```diff
@@ -67,53 +71,84 @@ func ServiceObject(namespace string, appKubernetesPartOfValue string) *v1.Servic
...
 
-func (r *serviceReconciler) createMetricsService(ctx context.Context) error {
-	appKubernetesPartOfValue := r.deployment.GetLabels()[common.AppKubernetesPartOfLabel]
-	service := ServiceObject(r.serviceNamespace, appKubernetesPartOfValue)
-	err := r.setServiceOwnerReference(service)
+func (s *serviceReconciler) setServiceOwnerReference(service *v1.Service) error {
+	return controllerutil.SetOwnerReference(s.deployment, service, s.client.Scheme())
+}
+
+func (s *serviceReconciler) createMetricsService(ctx context.Context, deployment *apps.Deployment, apiClient client.Client) error {
```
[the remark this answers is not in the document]
akrejcir: done

internal/controllers/ssp_controller.go
0xFelix: ACK, that got me confused.

internal/controllers/ssp_controller.go
0xFelix: ACK

internal/controllers/vm_controller.go
0xFelix: Silly question, but does using the plural make a difference?

0xFelix (commented):
/approve

Thanks!

PR #988, opened 2024-05-27
chore: Temporarily disable dependabot until update to golang 1.22

akrejcir:
The PRs opened by dependabot need golang 1.22, but we cannot update yet, because downstream containers don't work with this version.

**Release note**:
```release-note
None
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,26 +1,27 @@
+# TODO: Enable dependabot when we update to golang 1.22 in downstream.
 version: 2
-updates:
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
Is there a way to stop dependabot only from updating certain dependencies? IIRC there is `@dependabot ignore this dependency`.

codingben:
I think we're supporting Go 1.22 so you can close this PR.

akrejcir:
Yes. 
/close

PR #996, opened 2024-06-17
chore: Update controller-gen

akrejcir:
The old version, `v0.10.0`, failed to run when compiled using golang 1.22. That broke compilation of SSP operator.

This PR updates the version to `v0.14.0` which is not the latest, because the latest required golang 1.22. We still want to be able to compile SSP using golang 1.21.

**Release note**:
```release-note
None
```


config/crd/bases/ssp.kubevirt.io_ssps.yaml
```diff
@@ -46,31 +50,45 @@ spec:
...
-                      it to a specific reference. It is recommended that the reference
-                      be a specific commit or tag to ensure the generated contents
-                      does not change over time. As such it is recommended not to
-                      use branches as the ref for the time being. \n * Only VirtualMachineClusterPreference
-                      and VirtualMachineClusterInstancetype resources generated from
-                      the URL are deployed by the operand. \n See the following Kustomize
-                      documentation for more details: \n remote targets https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md"
+                    description: |-
+                      URL of a remote Kustomize target from which to generate and deploy resources.
+
```
0xFelix, line 55: This adds a lot of unneccessary newlines, can it be avoided?

akrejcir: I'm not sure how. Maybe there is a config option in the `controller-gen`.

akrejcir: No, there isn't any parameter.

0xFelix (approved):
/approve

config/crd/bases/ssp.kubevirt.io_ssps.yaml
0xFelix: Maybe a bug in controller-gen?

PR #1008, opened 2024-07-18
deps: Update golang version and dependencies

akrejcir:
**What this PR does / why we need it**:
These changes are in this PR:
- Updated golang version to `1.22.4`.
- Created golang workspace, because this is a multi-module repository.
- Removed unused indirect dependencies in the `api` module.
- Updated various dependencies.

**Release note**:
```release-note
None
```

PR #1010, opened 2024-07-19
deps: Update ci-builder to go 1.22.5

akrejcir:
Updated the CI builder to use golang 1.22.5.

This has to be done before any PR that needs it: https://github.com/kubevirt/ssp-operator/pull/1008

**Release note**:
```release-note
None
```

### DIFF
--- a/ci-builder/Dockerfile
+++ b/ci-builder/Dockerfile
@@ -1,7 +1,7 @@
 # This Dockerfile is used in openshift CI
-FROM quay.io/fedora/fedora:37
+FROM quay.io/fedora/fedora:40
 
-RUN curl -L https://go.dev/dl/go1.21.6.linux-amd64.tar.gz | tar -C /usr/local -xzf -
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
I noticed that `go1.22.5` is out already. Is there a reason to still use `go1.22.4`?

akrejcir:
Yes, downstream only uses `1.22.4`.

ci-builder/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This Dockerfile is used in openshift CI
 FROM quay.io/fedora/fedora:37
```
0xFelix, line 2: Upgrading fedora to a supported version might make sense too.

0xFelix (commented):
Downstream could also have additional patches ontop of the used `go1.22.4` which are missing in the public `go1.22.4`. I'd vote to use the latest dot release in this case.

ci-builder/Dockerfile
akrejcir: done

0xFelix (approved):
/approve
/lgtm

PR #1019, opened 2024-07-29
fix: Remove workspace vendoring, because downstream does not support it

akrejcir:
Downstream builds fail, because workspace vendoring is not supported.

**Release note**:
```release-note
None
```


0xFelix (approved):
/approve

PR #1020, opened 2024-07-29
chore: add renovate config

ksimon1:
Add renovate configuration, which updates main and release branches. release branches are configure to have only security patches.

This config was tested on: https://github.com/ksimon1/ssp-operator/pulls
**Release note**:
```
NONE
```


renovate.json
```diff
@@ -0,0 +1,75 @@
...
+           "release-v0.19",
+           "release-v0.18",
+           "release-v0.17"
+        ],
+        "matchPackagePatterns":[
+           "*"
+        ]
+     },
+     {
+        "matchPackageNames":[
```
akrejcir, line 49: According to the docs: https://docs.renovatebot.com/configuration-options/#matchpackagenames
`matchPackageNames` should contain names of packages. Is `golang` a special name for the compiler version?

renovate.json
```diff
@@ -0,0 +1,75 @@
...
+  "$schema":"https://docs.renovatebot.com/renovate-schema.json",
+  "baseBranches":[
+     "main",
+     "release-v0.20",
+     "release-v0.19",
+     "release-v0.18",
+     "release-v0.17"
+  ],
+  "constraints":{
+     "go":"1.22"
```
akrejcir, line 11: Older branches use older go versions. Can we configure this constraint for each branch?

renovate.json
```diff
@@ -0,0 +1,75 @@
...
+        "groupSlug":"all",
+        "matchBaseBranches":[
+           "main"
+        ],
+        "matchPackagePatterns":[
+           "*"
+        ]
+     },
+     {
+        "enabled":false,
```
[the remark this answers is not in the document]
ksimon1: This disables updates of regular packages for release branches and keeps only security updates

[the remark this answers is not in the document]
akrejcir: I see, thanks.

renovate.json
```diff
@@ -0,0 +1,75 @@
+{
+  "$schema":"https://docs.renovatebot.com/renovate-schema.json",
+  "baseBranches":[
+     "main",
+     "release-v0.20",
```
lmilbaum, line 5: You can specify the branch names with a wild card.
For example:
  "baseBranches": ["/^rhel-9.\\d$|^rhel-10.\\d|^rhel-10.0-beta$/"]

renovate.json
```diff
@@ -0,0 +1,75 @@
...
+     "release-v0.19",
+     "release-v0.18",
+     "release-v0.17"
+  ],
+  "constraints":{
```
lmilbaum, line 10: Not sure what the impact of this configuration

renovate.json
```diff
@@ -0,0 +1,75 @@
...
+  "prConcurrentLimit":3,
+  "groupName":"all dependencies",
+  "groupSlug":"all",
+  "lockFileMaintenance":{
+     "enabled":false
+  },
+  "labels":[
+     "release-note-none"
+  ],
+  "extends":[
```
lmilbaum, line 22: I like this section to be at the top of the configuration. It is not a must.

ksimon1: fixed

renovate.json
ksimon1: I removed it

renovate.json
ksimon1: We want to update the security dependencies only in some branches, not in all release-v... branches. Thats why I put there the names instead of regex.

PR #1032, opened 2024-08-06
fix: Correctly deploy instance types from URL

akrejcir:
This PR fixes 2 issues:
- When setting, unsetting, and setting again the `.spec.commonInstancetypes.URL` field, resources would not be deployed when URL is set for the second time. This was because the URL was cached the first time and not cleared.
- When instance types were deployed from URL, the operator would not revert any changes to them.

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/operands/common-instancetypes/reconcile.go
+++ b/internal/operands/common-instancetypes/reconcile.go
@@ -242,23 +242,25 @@ func (c *CommonInstancetypes) reconcileRemovedResources(request *common.Request)
 func (c *CommonInstancetypes) reconcileFromURL(request *common.Request) ([]common.ReconcileResult, error) {
 	// TODO - In the future we should handle cases where the URL remains the same but the provided resources change.
 	//nolint:staticcheck
-	if c.resourceURL != "" && c.resourceURL == *request.Instance.Spec.CommonInstancetypes.URL {
-		request.Logger.Info(fmt.Sprintf("Skipping reconcile of common-instancetypes from URL %s, force with a restart of the service.", *request.Instance.Spec.CommonInstancetypes.URL))
-		return nil, nil
+	if c.resourceURL == "" || c.resourceURL != *request.Instance.Spec.CommonInstancetypes.URL {
+		// Cache the URL, so we don't have to load it on every reconciliation.
+		//nolint:staticcheck
+		c.resourceURL = *request.Instance.Spec.CommonInstancetypes.URL
+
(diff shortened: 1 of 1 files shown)


internal/operands/common-instancetypes/reconcile.go
```diff
@@ -242,23 +242,25 @@ func (c *CommonInstancetypes) reconcileRemovedResources(request *common.Request)
...
+		var err error
+		c.virtualMachineClusterInstancetypes, c.virtualMachineClusterPreferences, err = c.FetchResourcesFromURL(c.resourceURL)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		request.Logger.Info(fmt.Sprintf("Skipping reading common-instancetypes from URL %s. Using cached resources. Force with a restart of the service.", *request.Instance.Spec.CommonInstancetypes.URL))
 	}
 
+	request.Logger.Info(fmt.Sprintf("Reconciling common-instancetypes from URL %s", c.resourceURL))
```
0xFelix, line 261:
```suggestion
	request.Logger.Info(fmt.Sprintf("Reconciling common-instancetypes from cached URL %s", c.resourceURL))
```

0xFelix (commented):
Code change looks good. Can you add tests for it?

akrejcir:
I've separated the first commit into separate PR #1033 , so we can merge it sooner.

codingben:
> I've separated the first commit into separate PR #1033 , so we can merge it sooner.

Will you rebase once the new PR is merged?

akrejcir:
I will, but not yet. Let's wait for Lee to know if this PR solves a bug, or if it was intended. I don't want to waste CI.

codingben:
> I will, but not yet. Let's wait for Lee to know if this PR solves a bug, or if it was intended. I don't want to waste CI.

Sure thing. I'd rebase it and mark it as draft to disable CI.

lyarwood:
> Now I'm not sure if it was intentional that resources from URL are not reconciled. Let's wait for @lyarwood .

Thanks for looking at this @akrejcir , I think my issue with continuously reconciling from the same URL was just the cost in building the resources each time. The call to `kustomize` not cheap and would definitely slow down the time taken by SSP as a whole to reconcile.

With v0.21 out the door I wanted to finally remove this functionality from SSP anyway for v0.22. We can still land this PR to resolve the set, unset and set use case you set out in the description but for it to be useful we would then need to backport that all the way back to v0.19. I'm not sure it's worth the effort tbh.

PR #1033, opened 2024-08-06
fix: common-instancetypes: Clear URL when deploying from bundle

akrejcir:
Previously, this order of operations would trigger a bug:
- set `.spec.commonInstancetypes.url` to a URL - The operator will deploy resources from the URL
- unset `.spec.commonInstancetypes.url` - The operator will deploy resources from internal bundle
- set `.spec.commonInstancetypes.url` to the same URL - The operator will ignore the resources and not deploy anything

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/operands/common-instancetypes/reconcile.go
+++ b/internal/operands/common-instancetypes/reconcile.go
@@ -272,6 +272,11 @@ func (c *CommonInstancetypes) reconcileFromURL(request *common.Request) ([]commo
 
 func (c *CommonInstancetypes) reconcileFromBundle(request *common.Request) ([]common.ReconcileResult, error) {
 	request.Logger.Info("Reconciling common-instancetypes from internal bundle")
+
+	// We need to clear the resourceURL, otherwise when switching between using and not using URL,
+	// the operator would not deploy resources from the URL.
+	c.resourceURL = ""
+
 	var err error
 	c.virtualMachineClusterInstancetypes, c.virtualMachineClusterPreferences, err = c.fetchResourcesFromBundle()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


0xFelix (commented):
The fix is quite easy, but in this case, doesn't ssp work like intended?

> `Skipping reconcile of common-instancetypes from URL %s, force with a restart of the service.`

It still holds true.

akrejcir:
Yes, the comment is true, but the behavior is not correct in my opinion. It is very unexpected from the user's point of view.
Even when user specifies the URL parameter, the operator will keep the resources from the internal bundle, ignoring the URL.

0xFelix:
Can you add a test?

codingben:
+1 unit test, was about also to comment, is it covered by unit tests?

internal/operands/common-instancetypes/reconcile.go
```diff
@@ -310,6 +310,11 @@ func (c *CommonInstancetypes) Reconcile(request *common.Request) ([]common.Recon
 	if request.Instance.Spec.CommonInstancetypes != nil && request.Instance.Spec.CommonInstancetypes.URL != nil {
 		return c.reconcileFromURL(request)
 	}
+
+	// We need to clear the resourceURL, otherwise when switching between using and not using URL,
+	// the operator would not deploy resources from the URL.
+	c.resourceURL = ""
```
codingben, line 316: Do you think we can clear URL inside `c.reconcileFromBundle()`?

akrejcir:
I've added the test.

0xFelix (commented):
/approve

codingben (approved):
/lgtm

PR #1038, opened 2024-08-09
tests: Show clearer error message when cleanup tests fail

akrejcir:
Previously the error message contained the whole object, but the text was too long and it was truncated, so often we could not read the name and namespace of the objects.

Now we show only group, version, kind, namespace and name.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/cleanup_test.go
+++ b/tests/cleanup_test.go
@@ -1,9 +1,10 @@
 package tests
 
 import (
+	"fmt"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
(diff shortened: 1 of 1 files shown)


jcanocan:
There's any chance to also print the status or something that will help to understand why it wasn't deleted?

tests/cleanup_test.go
```diff
@@ -61,7 +62,23 @@ var _ = Describe("Cleanup", func() {
...
 			})).To(Succeed())
-			Expect(list.Items).To(BeEmpty(), "Some resources were not deleted.")
+
+			if len(list.Items) == 0 {
+				continue
+			}
+
+			// Collect resources to more readable names
+			objectTypesAndNames := make([]string, 0, len(list.Items))
+			for _, item := range list.Items {
```
codingben, line 72: Can you move it to a new function that will return readable names?

```
objectTypesAndNames := getObjectTypesAndNames(list.Items)
Expect(objectTypesAndNames).To(BeEmpty(), "Some resources were not deleted.")
```

0xFelix (commented):
/approve

IMO it is okay to inline or to extract it as function, I'm fine with both.

PR #1070, opened 2024-09-10
feat: update validator target to be based on ubi-micro

nestoracunablanco:
Since Ubi-Micro has fewer packages, it has fewer security risks; therefore, no microdnf updates are necessary. This minimizes the need to execute an architecture-specific command while building the target image.

Fixes #

**Release note**:

```release-note NONE

```

### DIFF
--- a/validator.Dockerfile
+++ b/validator.Dockerfile
@@ -27,10 +27,8 @@ RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -a -ldflags="-X 'kubevirt.i
 -X 'kubevirt.io/ssp-operator/internal/template-validator/version.BRANCH=$BRANCH'\
 -X 'kubevirt.io/ssp-operator/internal/template-validator/version.REVISION=$REVISION'" -o kubevirt-template-validator internal/template-validator/main.go
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal
+FROM registry.access.redhat.com/ubi9/ubi-micro
 RUN mkdir -p /etc/webhook/certs
-
-RUN microdnf update -y  && microdnf clean all
 
 WORKDIR /
 COPY --from=builder /workspace/kubevirt-template-validator /usr/sbin/kubevirt-template-validator


validator.Dockerfile
```diff
@@ -27,11 +27,9 @@ RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -a -ldflags="-X 'kubevirt.i
 -X 'kubevirt.io/ssp-operator/internal/template-validator/version.BRANCH=$BRANCH'\
 -X 'kubevirt.io/ssp-operator/internal/template-validator/version.REVISION=$REVISION'" -o kubevirt-template-validator internal/template-validator/main.go
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal
+FROM registry.access.redhat.com/ubi9/ubi-micro
 RUN mkdir -p /etc/webhook/certs
 
-RUN microdnf update -y  && microdnf clean all
```
akrejcir, line 33: @ksimon1, you've added this line a long time ago: https://github.com/kubevirt/ssp-operator/pull/348

Do you remember why?

ksimon1: it can happen, that the released image contains vulnerabilities, which can be fixed by updating the packages with microdnf update -y

nestoracunablanco: @ksimon1, I appreciate your valid point. However, it's worth noting that the ubi-micro image contains significantly fewer packages, weighing in at 22.4 MB compared to 96.2 MB in the latest release.

Another important consideration is that running microdnf update can break the build if the host architecture differs from that of the container image, especially if no emulation is enabled.

Here’s an example of what I’m aiming to achieve: https://github.com/kubevirt/vm-console-proxy/blob/main/Dockerfile

In summary, my approach involves the following steps:

1. Build everything from the host architecture in the builder image.
2. Perform cross-compilation to the target architecture.
3. Copy the binaries to a image with the target architecture.

What are your thoughts on this approach?

ksimon1: I am not saying your changes are wrong, I am just explaining why I added the command 2 years ago :)
It is great, that you are working on making it available on multiple architecture.
The approach you mentioned looks good :+1:

nestoracunablanco: Thank you so much! I truly appreciate your feedback :+1:

PR #1119, opened 2024-11-01
feat: validator: Load TLS configuration from ConfigMap

akrejcir:
This PR moves the TLS configuration of `template-validator` to a ConfigMap. It is mounded as a file in the pod, and validator is able to update its configuration without restarting the pod.

Jira: https://issues.redhat.com/browse/CNV-28716

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/operands/template-validator/resources.go
+++ b/internal/operands/template-validator/resources.go
@@ -2,7 +2,6 @@ package template_validator
 
 import (
 	"fmt"
-	"strings"
 
 	templatev1 "github.com/openshift/api/template/v1"
 	admission "k8s.io/api/admissionregistration/v1"
@@ -16,7 +15,6 @@ import (
 	kubevirt "kubevirt.io/api/core"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
-	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/env"
 	common_templates "kubevirt.io/ssp-operator/internal/operands/common-templates"
 	metrics "kubevirt.io/ssp-operator/internal/operands/metrics"
@@ -37,6 +35,7 @@ const (
 	ServiceName                   = VirtTemplateValidator
 	MetricsServiceName            = "template-validator-metrics"
 	DeploymentName                = VirtTemplateValidator
+	ConfigMapName                 = VirtTemplateValidator
 	PrometheusLabel               = "prometheus.ssp.kubevirt.io"
 	kubernetesHostnameTopologyKey = "kubernetes.io/hostname"
 )
@@ -145,9 +144,11 @@ func newPodAntiAffinity(key, topologyKey string, operator metav1.LabelSelectorOp
 	}
 }
 
-func newDeployment(namespace string, replicas int32, image string, sspTLSOptions *common.SSPTLSOptions) *apps.Deployment {
-	const volumeName = "tls"
+func newDeployment(namespace string, replicas int32, image string) *apps.Deployment {
+	const secretVolumeName = "tls"
+	const configMapVolumeName = "config-map"
 	const certMountPath = "/etc/webhook/certs"
+	const configMapMountPath = "/tls-options"
 	trueVal := true
 	falseVal := false
 
@@ -196,19 +197,13 @@ func newDeployment(namespace string, replicas int32, image string, sspTLSOptions
 							fmt.Sprintf("--port=%d", ContainerPort),
 							fmt.Sprintf("--cert-dir=%s", certMountPath),
 						},
-						Env: []core.EnvVar{
-							{
-								Name:  tlsinfo.CiphersEnvName,
-								Value: strings.Join(sspTLSOptions.OpenSSLCipherNames, ","),
-							},
-							{
-								Name:  tlsinfo.TLSMinVersionEnvName,
-								Value: sspTLSOptions.MinTLSVersion,
-							},
-						},
 						VolumeMounts: []core.VolumeMount{{
-							Name:      volumeName,
+							Name:      secretVolumeName,
 							MountPath: certMountPath,
+							ReadOnly:  true,
+						}, {
+							Name:      configMapVolumeName,
+							MountPath: configMapMountPath,
 							ReadOnly:  true,
 						}},
 						SecurityContext: &core.SecurityContext{
@@ -240,10 +235,19 @@ func newDeployment(namespace string, replicas int32, image string, sspTLSOptions
 						},
 					}},
 					Volumes: []core.Volume{{
-						Name: volumeName,
+						Name: secretVolumeName,
 						VolumeSource: core.VolumeSource{
 							Secret: &core.SecretVolumeSource{
 								SecretName: SecretName,
+							},
+						},
+					}, {
+						Name: configMapVolumeName,
+						VolumeSource: core.VolumeSource{
+							ConfigMap: &core.ConfigMapVolumeSource{
+								LocalObjectReference: core.LocalObjectReference{
+									Name: ConfigMapName,
+								},
 							},
 						},
 					}},
@@ -252,6 +256,19 @@ func newDeployment(namespace string, replicas int32, image string, sspTLSOptions
 					},
 				},
 			},
+		},
+	}
+}
+
+func newConfigMap(namespace string, tlsOptionsJson string) *core.ConfigMap {
+	return &core.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ConfigMapName,
+			Namespace: namespace,
+			Labels:    CommonLabels(),
+		},
+		Data: map[string]string{
+			tlsinfo.TLSOptionsFilename: tlsOptionsJson,
 		},
 	}
 }
--- a/internal/template-validator/tlsinfo/tlsinfo.go
+++ b/internal/template-validator/tlsinfo/tlsinfo.go
@@ -34,10 +34,6 @@ type TLSInfo struct {
 	sspTLSOptions  common.SSPTLSOptions
 }
 
-func (ti *TLSInfo) IsEnabled() bool {
-	return ti.CertsDirectory != ""
-}
-
 func (ti *TLSInfo) Init() {
 	ti.initCerts()
 	ti.initCryptoConfig()
@@ -50,10 +46,6 @@ func (ti *TLSInfo) initCryptoConfig() {
 }
 
 func (ti *TLSInfo) initCerts() {
-	if !ti.IsEnabled() {
-		return
-	}
-
 	directory := ti.CertsDirectory
 	filesChanged, watcherCloser, err := watchDirectory(directory)
 	if err != nil {
--- a/internal/template-validator/tlsinfo/tlsinfo.go
+++ b/internal/template-validator/tlsinfo/tlsinfo.go
@@ -153,25 +153,24 @@ func (ti *TLSInfo) getCertificate() (*tls.Certificate, error) {
 	return ti.cert, nil
 }
 
-func (ti *TLSInfo) CreateTlsConfig() *tls.Config {
+func (ti *TLSInfo) CreateTlsConfig() (*tls.Config, error) {
+	cert, err := ti.getCertificate()
+	if err != nil {
+		return nil, fmt.Errorf("error getting certificate: %w", err)
+	}
+
 	tlsConfig := &tls.Config{
-		GetCertificate: func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
-			cert, err := ti.getCertificate()
-			if err != nil {
-				return nil, fmt.Errorf("error getting certificate: %w", err)
-			}
-			return cert, nil
-		},
+		Certificates: []tls.Certificate{*cert},
 	}
 
 	if !ti.sspTLSOptions.IsEmpty() {
 		tlsConfig.CipherSuites = common.CipherIDs(ti.sspTLSOptions.OpenSSLCipherNames, nil)
 		minVersion, err := ti.sspTLSOptions.MinTLSVersionId()
 		if err != nil {
-			panic(fmt.Sprintf("TLS Configuration broken, min version misconfigured %v", err))
+			return nil, fmt.Errorf("TLS Configuration broken, min version misconfigured %v", err)
 		}
 		tlsConfig.MinVersion = minVersion
 	}
 
-	return tlsConfig
+	return tlsConfig, nil
 }
--- a/internal/template-validator/tlsinfo/tlsinfo.go
+++ b/internal/template-validator/tlsinfo/tlsinfo.go
@@ -3,40 +3,43 @@ package tlsinfo
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strings"
 	"sync"
 	"time"
 
-	"github.com/fsnotify/fsnotify"
-
 	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/template-validator/filewatch"
 	"kubevirt.io/ssp-operator/internal/template-validator/logger"
 )
 
 const (
 	CertFilename         = "tls.crt"
 	KeyFilename          = "tls.key"
-	retryInterval        = 1 * time.Minute
 	CiphersEnvName       = "TLS_CIPHERS"
 	TLSMinVersionEnvName = "TLS_MIN_VERSION"
 )
 
 type TLSInfo struct {
 	CertsDirectory string
-	cert           *tls.Certificate
-	certLock       sync.Mutex
-	stopCertReload chan struct{}
-	sspTLSOptions  common.SSPTLSOptions
+
+	watch     filewatch.Watch
+	stopWatch chan struct{}
+
+	cert          *tls.Certificate
+	certError     error
+	certLock      sync.Mutex
+	sspTLSOptions common.SSPTLSOptions
 }
 
-func (ti *TLSInfo) Init() {
-	ti.initCerts()
+func (ti *TLSInfo) Init() error {
+	if err := ti.initCerts(); err != nil {
+		return err
+	}
 	ti.initCryptoConfig()
+	return nil
 }
 
 func (ti *TLSInfo) initCryptoConfig() {
@@ -45,100 +48,73 @@ func (ti *TLSInfo) initCryptoConfig() {
 	ti.sspTLSOptions.MinTLSVersion, _ = os.LookupEnv(TLSMinVersionEnvName)
 }
 
-func (ti *TLSInfo) initCerts() {
-	directory := ti.CertsDirectory
-	filesChanged, watcherCloser, err := watchDirectory(directory)
-	if err != nil {
-		panic(err)
+func (ti *TLSInfo) initCerts() error {
+	if ti.stopWatch != nil {
+		return fmt.Errorf("certificate watcher was already initialized")
 	}
 
-	ti.stopCertReload = make(chan struct{})
-	notify(filesChanged)
+	directory := ti.CertsDirectory
+	ti.updateCertificates(directory)
 
+	ti.watch = filewatch.New()
+	err := ti.watch.Add(directory, func() {
+		ti.updateCertificates(directory)
+	})
+	if err != nil {
+		return fmt.Errorf("error adding directory to watch: %w", err)
+	}
+
+	ti.stopWatch = make(chan struct{})
 	go func() {
-		defer watcherCloser.Close()
 		for {
 			select {
-			case <-filesChanged:
-				err := ti.updateCertificates(directory)
-				if err != nil {
-					go func() {
-						time.Sleep(retryInterval)
-						notify(filesChanged)
-					}()
-				}
-			case <-ti.stopCertReload:
+			case <-ti.stopWatch:
 				return
+			default:
+			}
+
+			watchErr := ti.watch.Run(ti.stopWatch)
+			if watchErr == nil {
+				return
+			}
+			// Log error and restart watch
+			logger.Log.Info("failed watching files", "error", watchErr)
+
+			select {
+			case <-ti.stopWatch:
+				return
+			case <-time.After(time.Second):
 			}
 		}
 	}()
+
+	return nil
 }
 
 func (ti *TLSInfo) Clean() {
-	if ti.stopCertReload != nil {
-		close(ti.stopCertReload)
+	if ti.stopWatch != nil {
+		close(ti.stopWatch)
+		ti.stopWatch = nil
 	}
 }
 
-func watchDirectory(directory string) (chan struct{}, io.Closer, error) {
-	watcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		logger.Log.Error(err, "Failed to create an inotify watcher")
-		return nil, nil, err
-	}
-
-	err = watcher.Add(directory)
-	if err != nil {
-		watcher.Close()
-		logger.Log.Error(err, "Failed to establish a watch", "directory", directory)
-		return nil, nil, err
-	}
-
-	filesChanged := make(chan struct{}, 1)
-	go func() {
-		for {
-			select {
-			case event, ok := <-watcher.Events:
-				if !ok {
-					return
-				}
-				if event.Op != fsnotify.Chmod {
-					notify(filesChanged)
-				}
-			case err, ok := <-watcher.Errors:
-				if !ok {
-					return
-				}
-				logger.Log.Error(err, "An error occurred while watching", "directory", directory)
-			}
-		}
-	}()
-
-	return filesChanged, watcher, nil
-}
-
-func notify(channel chan struct{}) {
-	select {
-	case channel <- struct{}{}:
-	default:
-	}
-}
-
-func (ti *TLSInfo) updateCertificates(directory string) error {
+func (ti *TLSInfo) updateCertificates(directory string) {
 	cert, err := loadCertificates(directory)
-	if err != nil {
-		logger.Log.Info("failed to load certificates",
-			"directory", directory,
-			"error", err)
-		return err
-	}
 
 	ti.certLock.Lock()
 	defer ti.certLock.Unlock()
-	ti.cert = cert
+	if err != nil {
+		ti.cert = nil
+		ti.certError = err
+		logger.Log.Info("failed to load certificates",
+			"directory", directory,
+			"error", err)
+		return
+	}
 
+	ti.cert = cert
+	ti.certError = nil
 	logger.Log.Info("certificate retrieved", "directory", directory, "name", cert.Leaf.Subject.CommonName)
-	return nil
 }
 
 func loadCertificates(directory string) (serverCrt *tls.Certificate, err error) {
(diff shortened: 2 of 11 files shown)


codingben:
@akrejcir Is it simialr to the implementation as in `kubevirt/vm-console-proxy`, e.g. [filewatch.go](https://github.com/kubevirt/vm-console-proxy/blob/main/pkg/filewatch/filewatch.go)? Maybe you can reuse this existing file to from `kubevirt/vm-console-proxy` to avoid code duplication?

akrejcir:
It is exactly the same package. I've mentioned it in the commit message: https://github.com/kubevirt/ssp-operator/pull/1119/commits/3b2e200f14a9ced14b647fdb231e68c5d4487a91

Do you mean that SSP should import `vm-console-proxy`? I would rather avoid it, because it will cause unnecessary indirect dependencies.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -3,40 +3,44 @@ package tlsinfo
...
+	cert          *tls.Certificate
+	certError     error
+	certLock      sync.Mutex
+	sspTLSOptions common.SSPTLSOptions
 }
 
-func (ti *TLSInfo) Init() {
-	ti.initCerts()
+func (ti *TLSInfo) Init() error {
+	err := ti.initCerts()
```
0xFelix: Could be inlined

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -45,100 +49,73 @@ func (ti *TLSInfo) initCryptoConfig() {
...
-	}
 
-	err = watcher.Add(directory)
-	if err != nil {
-		watcher.Close()
-		logger.Log.Error(err, "Failed to establish a watch", "directory", directory)
-		return nil, nil, err
-	}
+			watchErr := ti.watch.Run(ti.stopWatch)
+			if watchErr == nil {
```
0xFelix, line 70: Inline this?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -45,100 +49,73 @@ func (ti *TLSInfo) initCryptoConfig() {
...
-				if err != nil {
-					go func() {
-						time.Sleep(retryInterval)
-						notify(filesChanged)
-					}()
-				}
-			case <-ti.stopCertReload:
+			case <-ti.stopWatch:
 				return
+			default:
```
0xFelix, line 66: Is this required to satisfy the linter?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -45,100 +49,73 @@ func (ti *TLSInfo) initCryptoConfig() {
...
-				}
-				if event.Op != fsnotify.Chmod {
-					notify(filesChanged)
-				}
-			case err, ok := <-watcher.Errors:
-				if !ok {
-					return
-				}
-				logger.Log.Error(err, "An error occurred while watching", "directory", directory)
+			case <-ti.stopWatch:
```
0xFelix, line 77: Wouldn't it be enough to only delay it by a second?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -45,100 +49,73 @@ func (ti *TLSInfo) initCryptoConfig() {
...
+		close(ti.stopWatch)
+		ti.stopWatch = nil
 	}
 }
 
-func (ti *TLSInfo) updateCertificates(directory string) error {
+func (ti *TLSInfo) updateCertificates(directory string) {
 	cert, err := loadCertificates(directory)
+
+	ti.certLock.Lock()
```
0xFelix: Can you help me to understand what this mutex does?

internal/template-validator/validator/app.go
```diff
@@ -89,7 +89,11 @@ func (app *App) Run() {
 			CertsDirectory: app.certsDir,
 		}
 
-		tlsInfo.Init()
+		err := tlsInfo.Init()
```
0xFelix, line 96: Inline this?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -16,53 +16,45 @@ import (
...
 
-	directory := ti.CertsDirectory
-	ti.updateCertificates(directory)
-
 	ti.watch = filewatch.New()
-	err := ti.watch.Add(directory, func() {
-		ti.updateCertificates(directory)
-	})
+
+	err := ti.initCerts()
```
0xFelix, line 51: Inline this?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -92,6 +84,32 @@ func (ti *TLSInfo) initCerts() error {
...
 
+func (ti *TLSInfo) initCryptoConfig() error {
+	directory := ti.TLSOptionsDirectory
+	ti.updateTLSOptions(directory)
+
+	err := ti.watch.Add(directory, func() {
+		ti.updateTLSOptions(directory)
+	})
+	if err != nil {
+		return fmt.Errorf("error adding directory to watch: %w", err)
```
0xFelix, line 95: Consider a const for this err msg fmt?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -143,34 +161,80 @@ func loadCertificates(directory string) (serverCrt *tls.Certificate, err error)
...
+	optionsJson, err := os.ReadFile(optionsFilePath)
 	if err != nil {
-		return nil, fmt.Errorf("error getting certificate: %w", err)
+		return nil, err
 	}
 
-	tlsConfig := &tls.Config{
-		Certificates: []tls.Certificate{*cert},
+	tlsOptions := &common.SSPTLSOptions{}
+	err = json.Unmarshal(optionsJson, tlsOptions)
```
0xFelix, line 214: Inline

internal/operands/template-validator/resources.go
```diff
@@ -37,6 +35,7 @@ const (
 	ServiceName                   = VirtTemplateValidator
 	MetricsServiceName            = "template-validator-metrics"
 	DeploymentName                = VirtTemplateValidator
+	ConfigMapName                 = VirtTemplateValidator
```
0xFelix, line 38: Use `VirtTemplateValidator` directly?

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: no. It is to make the `select` non-blocking. It only checks if the `ti.stopWatch` channel is closed, and if not, it continues.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: Without the early stop? I'm not sure, IMHO code should unblock once the stop channel is closed.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: It guards access to the fields `ti.cert` and `ti.certError`.
These fields are accessed from 2 goroutines:
- The file watcher goroutine, that updates these fields when the mouned secret is changed.
- The TLS listener goroutine, that reads the certificate when a new incoming connection is accepted.

internal/template-validator/tlsinfo/tlsinfo.go
0xFelix: Yes that's what I was suggesting. Not sure if it makes a huge difference to have the early stop parallel to the delay. The next loop in the goroutine will also begin by checking if the stop was triggered.

internal/operands/template-validator/resources.go
akrejcir: I wanted to keep the pattern as other constants. I can change it if you want.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: done

internal/template-validator/validator/app.go
akrejcir: done

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: done

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: done

internal/template-validator/tlsinfo/tlsinfo.go
0xFelix: ACK

internal/template-validator/tlsinfo/tlsinfo.go
0xFelix: ACK

internal/operands/template-validator/resources.go
0xFelix: Let's stay consistent for now. It could be changed in a follow up.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -3,150 +3,138 @@ package tlsinfo
...
+	return nil
 }
 
-func watchDirectory(directory string) (chan struct{}, io.Closer, error) {
-	watcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		logger.Log.Error(err, "Failed to create an inotify watcher")
-		return nil, nil, err
-	}
+const watchErrorFmt = "error adding directory to watch: %w"
```
0xFelix, line 86: Personal preference, but move it to the beginning of the file?

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -174,31 +162,80 @@ func loadCertificates(directory string) (serverCrt *tls.Certificate, err error)
...
+	tlsOptions, err := loadTLSOptions(directory)
+
+	ti.lock.Lock()
+	defer ti.lock.Unlock()
+
+	if err != nil {
+		ti.minTLSVersion = 0
+		ti.cipherSuites = nil
+		ti.tlsOptionsError = err
+		logger.Log.Info("failed to TLS options",
```
0xFelix, line 175: Is it an error of level Info? Shouldn't it be higher?

0xFelix: Also, what did it fail to do? :)

internal/template-validator/tlsinfo/tlsinfo_test.go
```diff
@@ -15,79 +16,196 @@ import (
...
+			}, time.Second).Should(HaveOccurred())
+		})
 
-		Eventually(func() (*tls.Certificate, error) {
-			return tlsConfig.GetCertificate(nil)
-		}, time.Second).ShouldNot(BeNil())
-	})
+		It("should load certificate", func() {
+			Expect(tlsInfo.Init()).To(Succeed())
+			defer tlsInfo.Clean()
```
0xFelix, line 90: `defer` vs `DeferCleanup`?

internal/template-validator/tlsinfo/tlsinfo_test.go
```diff
@@ -15,79 +16,196 @@ import (
...
+				g.Expect(tlsConfig.Certificates).To(HaveLen(1))
+				g.Expect(tlsConfig.Certificates[0].Leaf.Subject.CommonName).To(Equal(certSubject))
+			}, time.Second).Should(Succeed())
+
+			Expect(os.WriteFile(filepath.Join(certDir, CertFilename), []byte{}, 0777)).To(Succeed())
+
+			Eventually(func() error {
+				_, err := tlsInfo.CreateTlsConfig()
+				return err
+			}, time.Second).Should(HaveOccurred())
```
0xFelix, line 138: Which error should occur?

internal/template-validator/tlsinfo/tlsinfo_test.go
```diff
@@ -15,79 +16,196 @@ import (
...
+			Expect(os.Remove(filepath.Join(certDir, CertFilename))).To(Succeed())
+			Expect(os.Remove(filepath.Join(certDir, KeyFilename))).To(Succeed())
+
+			Expect(tlsInfo.Init()).To(Succeed())
+			defer tlsInfo.Clean()
+
+			Consistently(func() error {
+				_, err := tlsInfo.CreateTlsConfig()
+				return err
+			}, time.Second).Should(HaveOccurred())
```
0xFelix, line 110: Which error?

internal/template-validator/tlsinfo/tlsinfo_test.go
```diff
@@ -15,79 +16,196 @@ import (
...
+		It("should fail if TLS options file does not exist", func() {
+			Expect(os.Remove(filepath.Join(tlsOptionsDir, TLSOptionsFilename))).To(Succeed())
+
+			Expect(tlsInfo.Init()).To(Succeed())
+			defer tlsInfo.Clean()
+
+			Consistently(func() error {
+				_, err := tlsInfo.CreateTlsConfig()
+				return err
+			}, time.Second).Should(HaveOccurred())
```
0xFelix, line 152: Same as above, and also below

internal/template-validator/tlsinfo/tlsinfo_test.go
akrejcir: `defer` is called when `func() {}` returns. `DeferCleanup()` is called after the `It()` test case, similarly to `AfterEach()`.
 It is useful to call in `BeforeEach()` to register cleanups that will be called after the test.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: I prefer it to be close the where it is used.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: Yes, changed it to error.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -174,31 +162,78 @@ func loadCertificates(directory string) (serverCrt *tls.Certificate, err error)
...
+		ti.minTLSVersion = 0
+		ti.cipherSuites = nil
+		ti.tlsOptionsError = fmt.Errorf("TLS Configuration broken, min version misconfigured: %w", err)
+		logger.Log.Error(ti.tlsOptionsError, "TLS options are not valid",
+			"directory", directory)
+		return
+	}
+
+	ti.minTLSVersion = minVersion
+	ti.cipherSuites = common.CipherIDs(tlsOptions.OpenSSLCipherNames, nil)
```
0xFelix, line 200: What happens if this leaves us with an empty slice in case of user misconfiguration?

internal/template-validator/tlsinfo/tlsinfo_test.go
```diff
@@ -15,79 +16,200 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
+	ocpconfigv1 "github.com/openshift/api/config/v1"
+
+	"kubevirt.io/ssp-operator/internal/common"
 )
 
+const certSubject = "test.kubevirt.io"
```
0xFelix, line 25: Can it be moved into the `Describe` below?

internal/template-validator/tlsinfo/tlsinfo_test.go
0xFelix: Does it make a difference using `DeferCleanup` here?

tests/crypto_policy_test.go
```diff
@@ -196,41 +215,62 @@ func tryToAccessEndpoint(pod core.Pod, subpath string, port uint16, tlsConfig cl
...
-	attemptedUrl = fmt.Sprintf("https://%s.%s.svc:%d%s", strategy.GetSSPWebhookServiceName(), strategy.GetSSPDeploymentNameSpace(), port, subpath)
+	// ${serviceName}.${serviceNamespace}.svc is used to access the endpoints we are testing.
+	attemptedUrl = fmt.Sprintf("https://%s.%s.svc:%d%s", serviceName, pod.Namespace, port, subpath)
 	_, err = httpClient.Get(attemptedUrl)
 	return attemptedUrl, err
 }
 
-func (c tlsConfigTestPermutation) testTLSEndpointAccessible(pod core.Pod, subpath string, port uint16, tlsConfig clientTLSOptions) {
-	_, err := tryToAccessEndpoint(pod, subpath, port, tlsConfig)
-	Expect(err).ToNot(HaveOccurred(), "Can't access pod %s, at port %d, with tlsConfig %#v", pod.Name, port, tlsConfig)
```
0xFelix, line 209: Not sure I understand this change. Why not keep the Expects? Looks more concise to me.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: Good point. If user passes unknown cipher names, they will be silently left out. That is how it is implemented in the library function `crypto.OpenSSLToIANACipherSuites()`. I will add code to return an error instead.

internal/template-validator/tlsinfo/tlsinfo_test.go
akrejcir: No, because it is used in function `writeCertificate()` that is outside the `Describe`.

internal/template-validator/tlsinfo/tlsinfo_test.go
akrejcir: I think only a performance difference.

tests/crypto_policy_test.go
akrejcir: Because the next commit calls this function in an `Eventually` loop. The `Expect()` would fail the test immediately.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: Thinking about this, it may be desired behavior. Usually, on TCP connection, the server gives client a list of supported ciphers and the client chooses. Any unknown ciphers are simply ignored.
Maybe it is ok to use the same approach for configuration file. User could specify multiple ciphers, and it is not an error if the validator does not know some of them.

What do you think?

internal/template-validator/tlsinfo/tlsinfo_test.go
0xFelix: Can it be moved to `AfterEach()`?

internal/template-validator/tlsinfo/tlsinfo.go
0xFelix: I was thinking what happens if you configure the API server with an empty slice if using TLS 1.1/1.2? Will it fallback to default ciphers? I'm not sure we should swallow this error silently. It is misconfiguration of the API server, not a failure in negotiating a cipher during a handshake.

internal/template-validator/tlsinfo/tlsinfo_test.go
akrejcir: Yes, but I prefer it this way. It is also how it's used in the validator code `app.go`.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: If we configure the server ciphers with empty slice `[]` and not use TLS 1.3, then no ciphers will be enabled, so it will not accept any connection. If we configure it with `nil` slice, then it will use default ciphers.
This is IMHO a bad API design in the standard library. Usually in other places there is no functional difference between `[]` and `nil`.

Ok, I will change the code, so it fails with error if ciphers in config file are not known.

akrejcir: I've implemented the error checking and noticed that OpenShift's TLS security profiles also contain ciphers that are not implemented in the TLS in golang and they are silently ignored.
For example, here is the TLS `Intermediate` profile:
https://github.com/openshift/api/blob/a2817b89f7e0989016967055a142d5e88bed18de/config/v1/types_tlssecurityprofile.go#L288-L303
```go
TLSProfileIntermediateType: {
	Ciphers: []string{
		"TLS_AES_128_GCM_SHA256",
		"TLS_AES_256_GCM_SHA384",
		"TLS_CHACHA20_POLY1305_SHA256",
		"ECDHE-ECDSA-AES128-GCM-SHA256",
		"ECDHE-RSA-AES128-GCM-SHA256",
		"ECDHE-ECDSA-AES256-GCM-SHA384",
		"ECDHE-RSA-AES256-GCM-SHA384",
		"ECDHE-ECDSA-CHACHA20-POLY1305",
		"ECDHE-RSA-CHACHA20-POLY1305",
		"DHE-RSA-AES128-GCM-SHA256",
		"DHE-RSA-AES256-GCM-SHA384",
	},
	MinTLSVersion: VersionTLS12,
},
```

It contains ciphers `DHE-RSA-AES128-GCM-SHA256` and `DHE-RSA-AES256-GCM-SHA384`, which are not available in the golang TLS library.
The code that does the translation between names and cipher IDs ignores them: https://github.com/openshift/library-go/blob/64780a4acea5ee3e565683ad96b1f4f0799f0bd2/pkg/crypto/crypto.go#L149-L180

So it is not trivial to check if a cipher name is valid. We don't want it to fail if users set one of the OpenShift's profiles, but they already contain unknown ciphers. So we would need to specifically allow them, even if they are not implemented.

What do you think?

0xFelix: As long as an empty slice that is not nil does not result in default values being used, I'm fine with it.

But strange that these profiles ship ciphers that are not supported. Are they supported in downstream distributions of golang?

akrejcir: The downstream TLS library is the same as upstream, there are no changes.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -174,31 +162,78 @@ func loadCertificates(directory string) (serverCrt *tls.Certificate, err error)
...
+	if ti.certError != nil {
+		return nil, fmt.Errorf("error getting certificate: %w", ti.certError)
+	}
+
+	if ti.tlsOptionsError != nil {
+		return nil, fmt.Errorf("error getting TLS options: %w", ti.tlsOptionsError)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates: []tls.Certificate{*ti.cert},
```
0xFelix, line 233: What if `ti.cert` is nil? You check `ti.certError` above but it might be worth to avoid a nil-ptr access by adding a second check.

0xFelix: Same for `cipherSuites == nil`.

tests/crypto_policy_test.go
```diff
@@ -109,9 +111,15 @@ var _ = Describe("Crypto Policy", func() {
...
+
 			applyTLSConfig(tlsConfigTestPermutation.openshiftTLSPolicy)
-			testMetricsEndpoint(pod, tlsConfigTestPermutation)
-			testWebhookEndpoint(pod, tlsConfigTestPermutation)
+			Expect(testMetricsEndpoint(pod, tlsConfigTestPermutation)).To(Succeed())
+			Expect(testWebhookEndpoint(pod, tlsConfigTestPermutation)).To(Succeed())
+
+			Eventually(func() error {
+				return testValidatorEndpoint(validatorPod, tlsConfigTestPermutation)
+			}, env.ShortTimeout(), time.Second).Should(Succeed())
```
0xFelix, line 122: Why do other endpoints not need an eventually?

akrejcir: Because `applyTLSConfig()` waits until the new configuration is read by SSP. SSP needs to go to `Deploying` state and then back to `Deployed`.
We don't have any way to check if template-validator reloaded the configuration.

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: In my opinion, it is better to panic in that case. If `ti.certError == nil` and `ti.cert == nil` , then it is a programmer's error and an error message is probably useless to the user. It's better that we catch the bug sooner if we see nil-ptr panic.

0xFelix: Won't this result in a crash loop possibly? A Pod could otherwise appear as not Ready instead, which would look cleaner to me. Not going to block the PR though.

akrejcir: You are right, I've added the check.

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -174,31 +162,82 @@ func loadCertificates(directory string) (serverCrt *tls.Certificate, err error)
...
-		}
-		tlsConfig.MinVersion = minVersion
+	if ti.tlsOptionsError != nil {
+		return nil, fmt.Errorf("error getting TLS options: %w", ti.tlsOptionsError)
+	}
+
+	if ti.cert == nil {
+		return nil, fmt.Errorf("the TLS certificate is unexpectedly nil")
+	}
+
```
0xFelix, line 235: What about `ti.cipherSuites`? Are we ever allowing them to be `nil`?

akrejcir: Yes, they are set to `nil` if the config file is empty. See line 183.

akrejcir: I don't understand your question. Can you explain what do you mean?

0xFelix: If the config file is empty, then TLS is going to be disabled and therefore `CreateTlsConfig` is not going to get called, correct? To me that means that `CreateTlsConfig` should not accept `ti.cipherSuites` to be `nil`.

akrejcir: No, TLS is always enabled. If the config file is empty, it just means that SSP did not pass any specific configuration to validator, and it should use the default.

jcanocan:
Nothing to comment from my site. The PR looks great. Nice job!
/lgtm

tests/crypto_policy_test.go
```diff
@@ -196,41 +217,62 @@ func tryToAccessEndpoint(pod core.Pod, subpath string, port uint16, tlsConfig cl
 		subpath = "/" + subpath
 	}
 
-	// ${webhookServiceName}.${deploymentNamespace}.svc is used to access the endpoints we are testing.
-	// It is used here for the metrics as well for simplicity, as it is the CN in the ca_cert
-	// and the metrics just sit on a different port on the same pod.
-	attemptedUrl = fmt.Sprintf("https://%s.%s.svc:%d%s", strategy.GetSSPWebhookServiceName(), strategy.GetSSPDeploymentNameSpace(), port, subpath)
+	// ${serviceName}.${serviceNamespace}.svc is used to access the endpoints we are testing.
```
codingben, line 220: nit: You've mentioned `${serviceName}.${serviceNamespace}` but instead do `${serviceName}.${pod.Namespace}`.

tests/crypto_policy_test.go
```diff
@@ -196,41 +217,62 @@ func tryToAccessEndpoint(pod core.Pod, subpath string, port uint16, tlsConfig cl
...
+	attemptedUrl = fmt.Sprintf("https://%s.%s.svc:%d%s", serviceName, pod.Namespace, port, subpath)
```
codingben, line 221: What is `subpath`? Specifically I mean how `%d%s` looks like?

codingben (approved):
/lgtm

internal/template-validator/tlsinfo/tlsinfo.go
0xFelix: ACK

0xFelix (approved):
/approve

/hold

Putting hold to give chance to address Ben's comment.

tests/crypto_policy_test.go
akrejcir: Yes. I think the Service must be in the same namespace as Deployment and Pods. It is possible to point Service to a pod in different namespace, but not with simple selector that we use.
https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors

codingben: Okay :+1:

PR #1131, opened 2024-11-14
chore: Allow dependabot to update ginkgo

akrejcir:
Updated dependabot config to allow updating ginkgo.
The ignore config did not work.

Makefile was changed to read ginkgo version from go.mod

**Release note**:
```release-note
None
```


0xFelix (approved):
/approve
/lgtm

PR #1146, opened 2024-11-28
tests: TLS policy: Use insecure TLS for connection to template-validator

akrejcir:
The current code in functional tests that loads root CA for TLS connection to template-validator does not work when SSP is deployed by OLM.

This commit is a simple fix, where we ignore the validity of the certificate passed by template-validator. This is ok for these tests, because different tests that check behavior of the template-validator would fail if the certificate was invalid.

Fixes: https://issues.redhat.com/browse/CNV-47164

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/crypto_policy_test.go
+++ b/tests/crypto_policy_test.go
@@ -11,16 +11,17 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


jcanocan:
Thanks!
/lgtm

0xFelix (approved):
/approve

PR #1148, opened 2024-11-28
test: delete ssp deployment after end-to-end tests

nestoracunablanco:
Experiments indicate that certain tests require a fresh environment to yield reliable and consistent results. One potential solution is to remove the SSP deployment during the cleanup phase, but only if the "skip cleanup" flag is not set.

Fixes #

**Release note**:

```release-note
NONE
```


akrejcir:
This is not what we want to do.
Instead your test script should remove the deployment after running the tests, if you need it.

/hold

akrejcir:
The reason why this is not what we want is that Deployment should not carry any state. The state is in the SSP object, which is removed at the end of the tests. Unless we run the tests in an existing cluster, where the SSP object was created by HCO.

nestoracunablanco:
Thanks for the explanation. Will close.

PR #1150, opened 2024-11-29
feat: architecture dependent template reading

nestoracunablanco:
This change requires templates to define their architecture.

**What this PR does / why we need it**: s390x enablement

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/template-bundle/bundle.go
+++ b/internal/template-bundle/bundle.go
@@ -2,9 +2,11 @@ package template_bundle
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 
 	templatev1 "github.com/openshift/api/template/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -16,6 +18,16 @@ import (
 type Bundle struct {
 	Templates   []templatev1.Template
 	DataSources []cdiv1beta1.DataSource
+}
+
+type templateArchitecture struct {
+	Spec struct {
+		Template struct {
+			Spec struct {
+				Architecture string `json:"architecture"`
+			} `json:"spec"`
+		} `json:"template"`
+	} `json:"spec"`
 }
 
 func ReadBundle(filename string) (Bundle, error) {
@@ -51,7 +63,17 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
 		if err != nil {
 			return nil, err
 		}
-		if template.Name != "" {
+		if template.Name == "" {
+			continue
+		}
+		templateArchitecture := templateArchitecture{}
+		if err = json.Unmarshal(template.Objects[0].Raw, &templateArchitecture); err != nil {
+			return nil, err
+		}
+		// DISCLAIMER: This is a temporary solution for delivering templates related to the host architecture.
+		// Once the common templates are released based on architecture, this change will no longer be necessary.
+		// Instead, a modification will be required in setup.go to specify the bundle to read from.
+		if templateArchitecture.Spec.Template.Spec.Architecture == runtime.GOARCH {
 			bundle = append(bundle, template)
 		}
 	}
--- a/internal/template-bundle/bundle_test.go
+++ b/internal/template-bundle/bundle_test.go
@@ -1,48 +1,59 @@
 package template_bundle
 
 import (
+	"runtime"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+func retrieveSuffix() string {
+	switch runtime.GOARCH {
+	case "s390x":
+		return "-s390x"
+	default:
+		return ""
+	}
+}
+
 var _ = Describe("Template bundle", Ordered, func() {
 	var (
 		testBundle Bundle
+		nameSuffix string
 	)
 
 	BeforeAll(func() {
 		var err error
 		testBundle, err = ReadBundle("template-bundle-test.yaml")
 		Expect(err).ToNot(HaveOccurred())
+		nameSuffix = retrieveSuffix()
 	})
 
 	It("should correctly read templates", func() {
 		templates := testBundle.Templates
 		Expect(templates).To(HaveLen(4))
-
 		{
 			templ := templates[0]
-			Expect(templ.Name).To(Equal("centos-stream8-server-medium"))
+			Expect(templ.Name).To(Equal("centos-stream8-server-medium" + nameSuffix))
 			Expect(templ.Annotations).To(HaveKey("name.os.template.kubevirt.io/centos-stream8"))
 			Expect(templ.Objects).To(HaveLen(1))
 		}
 		{
 			templ := templates[1]
-			Expect(templ.Name).To(Equal("centos-stream8-desktop-large"))
+			Expect(templ.Name).To(Equal("centos-stream8-desktop-large" + nameSuffix))
 			Expect(templ.Annotations).To(HaveKey("name.os.template.kubevirt.io/centos-stream8"))
 			Expect(templ.Objects).To(HaveLen(1))
 		}
 		{
 			templ := templates[2]
-			Expect(templ.Name).To(Equal("windows10-desktop-medium"))
+			Expect(templ.Name).To(Equal("windows10-desktop-medium" + nameSuffix))
 			Expect(templ.Annotations).To(HaveKey("name.os.template.kubevirt.io/win10"))
 			Expect(templ.Objects).To(HaveLen(1))
 		}
 		{
 			templ := templates[3]
-			Expect(templ.Name).To(Equal("rhel8-saphana-tiny"))
+			Expect(templ.Name).To(Equal("rhel8-saphana-tiny" + nameSuffix))
 			Expect(templ.Annotations).To(HaveKey("name.os.template.kubevirt.io/rhel8.4"))
 			Expect(templ.Objects).To(HaveLen(1))
 		}
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -56,6 +56,12 @@ var (
 	sspWebhookServiceName  = "ssp-webhook-service"
 )
 
+const (
+	amd64 = "amd64"
+	s390x = "s390x"
+	arm   = "aarch64"
+)
+
 type TestSuiteStrategy interface {
 	Init()
 	Cleanup()
@@ -367,7 +373,9 @@ var (
 	sspListerWatcher   cache.ListerWatcher
 	portForwarder      PortForwarder
 	deploymentTimedOut bool
+	nodeArchitecture   string
 	clusterMachineType string
+	templatesSuffix    string
 )
 
 var _ = BeforeSuite(func() {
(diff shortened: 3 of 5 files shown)


nestoracunablanco:
@ksimon1 @lyarwood The XL size was necessary because I had to modify the template-bundle-test.yaml file. I apologize for this, but I did not find any other way.

internal/template-bundle/bundle_test.go
```diff
@@ -1,48 +1,61 @@
...
+		return "-s390x"
+	default:
+		return ""
+	}
+}
+
 var _ = Describe("Template bundle", Ordered, func() {
 	var (
 		testBundle Bundle
+		suffix     string
```
akrejcir, line 25: nit - Please give it a more descriptive name, like `nameSuffix`.

tests/tests_suite_test.go
```diff
@@ -54,6 +54,9 @@ var (
 	sspDeploymentName      = "ssp-operator"
 	sspDeploymentNamespace = "kubevirt"
 	sspWebhookServiceName  = "ssp-webhook-service"
+	amd64Arch              = "amd64"
+	s390xArch              = "s390x"
+	armArch                = "aarch64"
```
akrejcir, line 59: These can be constants.

internal/template-bundle/bundle.go
```diff
@@ -52,11 +54,26 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
...
-			bundle = append(bundle, template)
+			templateArch, _ := retrieveTemplateArch(template)
+			if templateArch == runtime.GOARCH {
+				bundle = append(bundle, template)
+			}
 		}
 	}
 }
 
+func retrieveTemplateArch(templ templatev1.Template) (string, error) {
```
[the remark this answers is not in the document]
nestoracunablanco: I acknowledge that I added this feature to the common templates. However, until these templates are delivered, my current approach is to obtain the architecture in this manner. Once the multiarch common templates are delivered available, we can simplify the code.
Even better, if we deliver architecture-dependent template files [here](https://github.com/kubevirt/ssp-operator/tree/main/data/common-templates-bundle) we can modify only this [line](https://github.com/kubevirt/ssp-operator/blob/main/internal/controllers/setup.go#L53) based on the runtime architecture.

tests/tests_suite_test.go
nestoracunablanco: Done.

internal/template-bundle/bundle_test.go
nestoracunablanco: Done.

internal/template-bundle/bundle.go
```diff
@@ -52,11 +54,26 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
...
+	var data map[string]interface{}
+	unmarschal_err := json.Unmarshal(templ.Objects[0].Raw, &data)
+	if unmarschal_err != nil {
+		return "", unmarschal_err
+	}
+	object_spec := data["spec"].(map[string]interface{})
```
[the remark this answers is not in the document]
nestoracunablanco: Done.

internal/template-bundle/bundle.go
```diff
@@ -52,11 +54,26 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
...
+			if templateArch == runtime.GOARCH {
+				bundle = append(bundle, template)
+			}
 		}
 	}
 }
 
+func retrieveTemplateArch(templ templatev1.Template) (string, error) {
+	var data map[string]interface{}
+	unmarschal_err := json.Unmarshal(templ.Objects[0].Raw, &data)
```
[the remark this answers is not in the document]
nestoracunablanco: Done.

internal/template-bundle/bundle.go
```diff
@@ -52,11 +54,26 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
 			return nil, err
 		}
 		if template.Name != "" {
-			bundle = append(bundle, template)
+			templateArch, _ := retrieveTemplateArch(template)
```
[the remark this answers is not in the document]
nestoracunablanco: Done.

internal/template-bundle/bundle.go
```diff
@@ -52,11 +54,26 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
 			return nil, err
 		}
 		if template.Name != "" {
```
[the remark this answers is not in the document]
nestoracunablanco: Nice suggestion, already implemented.

internal/template-bundle/bundle.go
```diff
@@ -51,12 +53,31 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
 		if err != nil {
 			return nil, err
 		}
-		if template.Name != "" {
+		if template.Name == "" {
+			continue
+		}
+		templateArch, retrieveTemplateArchErr := retrieveTemplateArch(template)
```
0xFelix, line 59: Is there a need for a new err variable?

internal/template-bundle/bundle.go
```diff
@@ -51,12 +53,31 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
...
+		}
+		if templateArch == runtime.GOARCH {
 			bundle = append(bundle, template)
 		}
 	}
 }
 
+func retrieveTemplateArch(templ templatev1.Template) (string, error) {
+	var data map[string]interface{}
+	err := json.Unmarshal(templ.Objects[0].Raw, &data)
```
0xFelix, line 71: Inline this into if below

0xFelix: Can you please try to unmarshal that into a `VirtualMachine` instead of an interface. Perhaps even using [UnmarshalStrict](https://pkg.go.dev/sigs.k8s.io/yaml#UnmarshalStrict) if possible.

internal/template-bundle/bundle_test.go
```diff
@@ -1,48 +1,61 @@
...
 	)
 
 	BeforeAll(func() {
 		var err error
 		testBundle, err = ReadBundle("template-bundle-test.yaml")
 		Expect(err).ToNot(HaveOccurred())
+		nameSuffix = retrieveSuffix()
 	})
 
 	It("should correctly read templates", func() {
```
0xFelix, line 35: This test should iterate over all supported archs, independent of the hosts architecture. Consider making it a table.

tests/tests_suite_test.go
```diff
@@ -407,7 +415,9 @@ var _ = BeforeSuite(func() {
 	setupApiClient()
 	strategy.Init()
 
-	clusterMachineType = retrieveQemuMachineType()
+	nodeArchitecture = retrieveNodeArchitecture()
```
0xFelix, line 418: Tbh, I'm not sure this will bite us in the future. It assumes the arch of all nodes by looking at the first node only.

internal/template-bundle/bundle.go
nestoracunablanco: I believe my background is influencing my approach. I was taught to use final variables whenever possible, and even in programming languages that don’t support them directly, I tend to assign values only once.
While I’m not a strict adherent of functional programming, it’s a mindset I carry with me. However, if this seems awkward to Go programmers, I’m willing to adapt.
I will make the requested changes.

nestoracunablanco: There was an error in the lines below. Thank you for pointing that out.

internal/template-bundle/bundle.go
nestoracunablanco: While I wasn't able to implement it exactly as you envisioned, I believe I understand your idea.

internal/template-bundle/bundle_test.go
nestoracunablanco: The is that the templates for this provisional solution are read in the readTemplates function based on the `runtime.GOARCH` constant. While I could attempt to mock this, it would come at the cost of readability.

If the common templates release files for different architectures, we can revert these changes and implement the necessary modifications in the `CreateControllers` function within `setup.go`. What are your thoughts on this?

internal/template-bundle/bundle.go
akrejcir: A template, in general, is not a valid `VirtualMachine` struct, so it cannot be unmarshalled. Because any field in the template can contain a variable that is substituted when the template is instantiated, something like: `${NAME}`. If this variable is on field that is not of type `string` the unmarshall would fail.

tests/tests_suite_test.go
nestoracunablanco: The primary concern here is that the s390x templates are provided through a suffix. As I mentioned earlier, if the common templates were delivered as three separate files—one generic, one for amd64, and one for s390x—the tests would become architecture-independent.
The goal is to introduce an intermediary step. Once the multi-architecture common templates are delivered to the ssp-operator, we can simplify the code while minimizing the risk of breaking anything.

internal/template-bundle/bundle.go
nestoracunablanco: @akrejcir @0xFelix I believe the trigger was the var data map[string]interface{}. Is that correct? If so, I have found an alternative solution.

0xFelix: > If this variable is on field that is not of type string the unmarshall would fail.

I see, but `interface{}` stings my eye. Anything we can do here to get around it?

nestoracunablanco: [Solved](https://github.com/kubevirt/ssp-operator/pull/1150/files#diff-32b371bd23ccfa0083d4fcdfed2d7d28f755ca79dab2d6cca92e5b4fcbb16a0aR23)

internal/template-bundle/bundle_test.go
nestoracunablanco: Done.

internal/template-bundle/bundle.go
```diff
@@ -51,7 +63,14 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
...
 		}
-		if template.Name != "" {
+		if template.Name == "" {
+			continue
+		}
+		templateArchitecture := TemplateArchitecture{}
+		if err = json.Unmarshal(template.Objects[0].Raw, &templateArchitecture); err != nil {
+			return nil, err
+		}
+		if templateArchitecture.Spec.Template.Spec.Architecture == runtime.GOARCH {
```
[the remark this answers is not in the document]
nestoracunablanco: Done.

tests/tests_suite_test.go
```diff
@@ -56,6 +56,12 @@ var (
 	sspWebhookServiceName  = "ssp-webhook-service"
 )
 
+const (
+	Amd64Arch = "amd64"
+	S390xArch = "s390x"
+	ArmArch   = "aarch64"
```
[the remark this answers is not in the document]
nestoracunablanco: I made some additional changes to `tests_suite_test.go,` but I used your suggestion as inspiration.

internal/template-bundle/bundle_test.go
```diff
@@ -1,48 +1,61 @@
...
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+func retrieveSuffix() string {
+	switch runtime.GOARCH {
+	case "amd64":
+		return ""
```
0xFelix, line 14: Same as default case, might drop that? :)

nestoracunablanco: Done

0xFelix (approved):
/lgtm

PR #1152, opened 2024-12-02
fix: test: Prepare SSP object for TLS policy tests

akrejcir:
Before running these tests, the SSP should not have any TLS policy defined. The tests wait for reconciliation, and it would not happen if the same TLS policy is already defined on SSP.

Jira: https://issues.redhat.com/browse/CNV-47164

**Release note**:
```release-note
None
```


codingben (approved):
/lgtm

0xFelix (commented):
/approve

PR #1154, opened 2024-12-03
Bump deps

jcanocan:

This is a manual bump of https://github.com/kubevirt/ssp-operator/pull/1149. The following dependencies have not been updated because they required go version >1.23.0. This version is not jet, available in downstream:

- [github.com/machadovilaca/operator-observability](https://github.com/machadovilaca/operator-observability)
- [github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring](https://github.com/prometheus-operator/prometheus-operator)

These dependencies will be updated in a follow-up PR.

Moreover, it fixes some problems caused by the latest dep updates, as well as updating the golangci-lint version.

Fixes #

**Release note**:

```release-note
NONE
```


jcanocan:
Change: rebase to main latest commits

akrejcir:
Thanks for fixing this!

There are some unrelated changes, probably from old generator. Can you try removing the `bin` directory in your local repo and regenerating files using `make container-build` ?

jcanocan:
Change: reverted some mistakes due to old local `/bin` files.

dominikholler:
> This version is not jet, available in downstream

Downstream tries to keep in sync with upstream kubernetes in 
https://github.com/kubernetes/kubernetes/blob/master/staging/publishing/rules.yaml , so it might be cleaner to refer to kubernetes directly. But it would not change the technical fact, just the way we see it.

Would renovate be able to generate PRs like this one, even renovate is blocked currently by https://issues.redhat.com/browse/CWFHEALTH-3336  or  https://issues.redhat.com/browse/CWFHEALTH-3173 ?

jcanocan:
> Thanks for fixing this!
> 
> There are some unrelated changes, probably from old generator. Can you try removing the `bin` directory in your local repo and regenerating files using `make container-build` ?

Thanks! That did the trick!

jcanocan:
> > This version is not jet, available in downstream
> 
> Downstream tries to keep in sync with upstream kubernetes in https://github.com/kubernetes/kubernetes/blob/master/staging/publishing/rules.yaml , so it might be cleaner to refer to kubernetes directly. But it would not change the technical fact, just the way we see it.

Adjusted. Thanks!

akrejcir:
I see that linter timed out:
```
/workspace/bin/golangci-lint run --timeout 5m
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
make: *** [Makefile:327: lint] Error 4
```

Sometimes when I run the `golangci-lint` locally, it consumes all memory and then OOM kills it. Maybe the same happens in CI now. Can you try running it locally?

jcanocan:
> I see that linter timed out:
> 
> ```
> /workspace/bin/golangci-lint run --timeout 5m
> level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> make: *** [Makefile:327: lint] Error 4
> ```
> 
> Sometimes when I run the `golangci-lint` locally, it consumes all memory and then OOM kills it. It probably has some memory leak. Maybe the same happens in CI now. Can you try running it locally?

Yeah, that happened to me with previous version of `golangci-lint` that's why I've updated it.
It works locally perfectly fine. Could you please try locally as well?

internal/controller/handler-hook/handler_hook.go
```diff
@@ -2,12 +2,12 @@ package handler_hook
 
 import (
 	"context"
-
```
akrejcir, line 5: No need to remove the line.

internal/controller/handler-hook/handler_hook.go
```diff
@@ -18,6 +18,12 @@ func New(wrapped handler.EventHandler, hookFunc HookFunc) handler.EventHandler {
 		hookFunc: hookFunc,
 	}
 }
+func New1(wrapped handler.EventHandler, hookFunc HookFunc) handler.EventHandler {
```
akrejcir, line 21: This is the exact same function as the one above.

jcanocan:
> > I see that linter timed out:
> > ```
> > /workspace/bin/golangci-lint run --timeout 5m
> > level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
> > make: *** [Makefile:327: lint] Error 4
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Sometimes when I run the `golangci-lint` locally, it consumes all memory and then OOM kills it. It probably has some memory leak. Maybe the same happens in CI now. Can you try running it locally?
> 
> Yeah, that happened to me with previous version of `golangci-lint` that's why I've updated it. It works locally perfectly fine. Could you please try locally as well?

Let's try with previous linter version. We can try to update this in a follow-up if it works.

akrejcir (commented):
I've tried running it locally and it passed ok.

internal/controller/handler-hook/handler_hook.go
jcanocan: Ups, you are right! That was a test that I have forgotten to remove :sweat_smile:
I will remove it.

jcanocan:
> I've tried running it locally and it passed ok.

Could be related with the fact that version `golangci-lint` 1.62.2 is using go version 1.23.3 and SSP 1.22.9. Let's see...

jcanocan:
@akrejcir Looks like it does not time out now. Let's stick with version 1.55.2 for now. I will try to update it in a follow-up PR.

internal/controller/handler-hook/handler_hook.go
jcanocan: Reverted.

internal/controller/handler-hook/handler_hook.go
jcanocan: Removed.

go.mod
```diff
@@ -1,45 +1,46 @@
...
-	k8s.io/api v0.30.3
-	k8s.io/apiextensions-apiserver v0.30.3
-	k8s.io/apimachinery v0.30.3
-	k8s.io/client-go v0.30.3
-	k8s.io/kube-aggregator v0.30.3
-	k8s.io/utils v0.0.0-20240502163921-fe8a2dddb1d0
-	kubevirt.io/api v1.3.0
-	kubevirt.io/containerized-data-importer-api v1.59.0
+	k8s.io/api v0.31.3
+	k8s.io/apiextensions-apiserver v0.31.2
```
0xFelix, line 27: Keep that in sync with other versions? `v0.31.3`?

0xFelix (approved):
/lgtm

PR #1156, opened 2024-12-03
feat: read architecture specific common templates file

nestoracunablanco:
This change prepares for architecture-specific delivery. The logic will first attempt to locate the architecture-specific common templates bundle. If it is not found, it will then fall back to the generic bundle.

**What this PR does / why we need it**: s390x enablement

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/template-bundle/bundle.go
+++ b/internal/template-bundle/bundle.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"runtime"
 
 	templatev1 "github.com/openshift/api/template/v1"
@@ -13,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/yaml"
 	cdiv1beta1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+	common_templates "kubevirt.io/ssp-operator/internal/operands/common-templates"
 )
 
 type Bundle struct {
@@ -45,6 +47,18 @@ func ReadBundle(filename string) (Bundle, error) {
 		Templates:   templates,
 		DataSources: sources,
 	}, nil
+}
+
+func RetrieveCommonTemplatesBundleFile(templateBundleDir string) (string, error) {
+	archDependentFileName := filepath.Join(templateBundleDir, fmt.Sprintf("common-templates-%s-%s.yaml", runtime.GOARCH, common_templates.Version))
+	if _, err := os.Stat(archDependentFileName); err == nil {
+		return archDependentFileName, nil
+	}
+	archIndependentFileName := filepath.Join(templateBundleDir, fmt.Sprintf("common-templates-%s.yaml", common_templates.Version))
+	if _, err := os.Stat(archIndependentFileName); err == nil {
+		return archIndependentFileName, nil
+	}
+	return "", fmt.Errorf("failed to find common-templates bundles, none of the files were found: %s, %s", archDependentFileName, archIndependentFileName)
 }
 
 func readTemplates(filename string) ([]templatev1.Template, error) {
--- a/internal/template-bundle/bundle_test.go
+++ b/internal/template-bundle/bundle_test.go
@@ -1,11 +1,15 @@
 package template_bundle
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"runtime"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	common_templates "kubevirt.io/ssp-operator/internal/operands/common-templates"
 )
 
 func retrieveSuffix() string {
@@ -19,8 +23,11 @@ func retrieveSuffix() string {
 
 var _ = Describe("Template bundle", Ordered, func() {
 	var (
-		testBundle Bundle
-		nameSuffix string
+		testBundle          Bundle
+		nameSuffix          string
+		tmpDir              string
+		archIndependentFile string
+		archDependentFile   string
 	)
 
 	BeforeAll(func() {
@@ -28,6 +35,12 @@ var _ = Describe("Template bundle", Ordered, func() {
 		testBundle, err = ReadBundle("template-bundle-test.yaml")
 		Expect(err).ToNot(HaveOccurred())
 		nameSuffix = retrieveSuffix()
+	})
+
+	BeforeEach(func() {
+		tmpDir = GinkgoT().TempDir()
+		archIndependentFile = filepath.Join(tmpDir, fmt.Sprintf("common-templates-%s.yaml", common_templates.Version))
+		archDependentFile = filepath.Join(tmpDir, fmt.Sprintf("common-templates-%s-%s.yaml", runtime.GOARCH, common_templates.Version))
 	})
 
 	It("should correctly read templates", func() {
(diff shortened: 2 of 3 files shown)


internal/template-bundle/bundle.go
```diff
@@ -47,6 +54,26 @@ func ReadBundle(filename string) (Bundle, error) {
 	}, nil
 }
 
+func RetrieveCommonTemplatesBundleFile(templateBundleDir string) (string, error) {
+	archDependentFileName := filepath.Join(templateBundleDir, commonTemplatesFilePrefix+runtime.GOARCH+"-"+common_templates.Version+commonTemplatesFileSuffix)
+	archIndependentFileName := filepath.Join(templateBundleDir, commonTemplatesFilePrefix+common_templates.Version+commonTemplatesFileSuffix)
```
akrejcir, line 59: The string concatenation is a little confusing here. Can you use `fmt.Sprintf()`? The format string could be:
```
common-templates-%s-%s.yaml`
```

internal/template-bundle/bundle.go
```diff
@@ -47,6 +54,26 @@ func ReadBundle(filename string) (Bundle, error) {
...
+	commonTemplatesExistingFiles := filterExistingTemplateFiles([]string{archDependentFileName, archIndependentFileName})
```
akrejcir, line 60: Can you simplify this by not using a filter function? IMHO directly checking if the 2 files exists is more readable.
```
if _, err := os.Stat(archDependentFileName); err == nil {
    return archDependentFileName, nil
}

if _, err := os.Stat(archIndependentFileName); err == nil {
    return archIndependentFileName, nil
}
```

internal/template-bundle/bundle_test.go
```diff
@@ -21,13 +23,19 @@ var _ = Describe("Template bundle", Ordered, func() {
...
 	BeforeAll(func() {
 		var err error
 		testBundle, err = ReadBundle("template-bundle-test.yaml")
 		Expect(err).ToNot(HaveOccurred())
 		nameSuffix = retrieveSuffix()
+		_, filename, _, ok := runtime.Caller(0)
+		Expect(ok).To(BeTrue())
+		fileAbsolutePath, err := filepath.Abs(filename)
+		Expect(err).NotTo(HaveOccurred())
+		testDir = filepath.Dir(fileAbsolutePath)
```
akrejcir, line 38: Instead of reading files from the repo, can you create them in a temp directory when the test starts?
Ginkgo can create the directory for you:
```
tmpDir := GinkgoT().TempDir()
```

internal/template-bundle/test-files/all-supported-archs/common-templates-amd64-v0.31.0.yaml
akrejcir, line 1: Please don't commit these empty files into the repo.

internal/template-bundle/bundle.go
nestoracunablanco: Done

internal/template-bundle/bundle.go
nestoracunablanco: Done

internal/template-bundle/bundle_test.go
nestoracunablanco: Nice idea, thanks for pointing that out

internal/template-bundle/test-files/all-supported-archs/common-templates-amd64-v0.31.0.yaml
nestoracunablanco: Done

internal/template-bundle/bundle_test.go
```diff
@@ -75,6 +87,37 @@ var _ = Describe("Template bundle", Ordered, func() {
...
 		Expect(ds2.Spec.Source.PVC.Namespace).To(Equal("kubevirt-os-images"))
 	})
+
+	It("should throw an error retrieving the bundle file", func() {
+		_, err := RetrieveCommonTemplatesBundleFile(tmpDir)
+		Expect(err.Error()).To(ContainSubstring("failed to find common-templates bundles, none of the files were found"))
+	})
+
+	It("should retrieve the bundle arch independent file", func() {
+		_, err := os.OpenFile(archIndependentFile, os.O_CREATE, 0644)
```
akrejcir, line 97: Please, close the file.

nestoracunablanco: I found a better way. It seems that os.WriteFile implicitly handles both opening and closing the file. Thank you for bringing that to my attention

dominikholler:
@nestoracunablanco
Is the assumption that the node, which runs the ssp-operator, has the same architecture as the nodes running the VMs?
Are you aware that this change might limit the usage of templates in clusters with worker nodes of multiple architectures?

nestoracunablanco:
Hello @dominikholler,

Thank you for reaching out. Regarding your questions:

1. _Are you aware that this change might limit the usage of templates in clusters with worker nodes of multiple architectures?_
- The assumption is not that the node running the same architecture as the SSP operator will have the same architecture as the nodes running the VMs. However, it is true that multi-architecture support is currently not available and would require significant design changes. With the existing design (which works well for amd64), the SSP operator attempts to import the data volumes even if they are not enabled for the host architecture (for example, Windows or OpenSUSE on s390x). To enhance the user experience for s390x, these options are not presented.
For amd64, this change is completely transparent. In the case of s390x, end users can clone a template and modify it according to their needs (in case they want to run amd64 workloads).

2.  _Does this mean you prefer to adopt the volume (instance types/preferences) approach for clusters using worker nodes of mixed architectures?_
- I do not have a personal opinion on this matter. My primary focus is to enable the existing features for the s390x architecture.

Looking ahead, here are some ideas to address the lack of multi-architecture worker node support (though they may not be fully feasible):
- Load the templates based on the available architectures across all nodes.
- Import the data volumes selectively.

In my opinion, enabling multi-architecture worker nodes would require a design change that necessitates consensus.

dominikholler:
On Thu, 05 Dec 2024 02:02:48 -0800
Nestor Acuna Blanco ***@***.***> wrote:

> Hello @dominikholler,
>
> Thank you for reaching out. Regarding your questions:
>
> 1. _Are you aware that this change might limit the usage of templates in clusters with worker nodes of multiple architectures?_
> - The assumption is not that the node running the same architecture as the SSP operator will have the same architecture as the nodes running the VMs. However, it is true that multi-architecture support is currently not available and would require significant design changes. With the existing design (which works well for amd64), the SSP operator attempts to import the data volumes even if they are not enabled for the host architecture (for example, Windows or OpenSUSE on s390x).

As mentioned more detailed below, the import of the DataImportCrons
respects the architecture if possible.

> To enhance the user experience for s390x, these options are not
> presented.

I do not understand. If the containerdisk imported by a DataImportCron
contains s390x, and the containerdisk is referring to the s390x
preference
(https://github.com/kubevirt/common-instancetypes/pull/290), wouldn't the OpenShift UI create a working VM from the related golden base image?

Are you looking for a way to prevent importing DataImportCrons which
does not support the architecture of the node?


> For amd64, this change is completely transparent. In the case of s390x, end users can clone a template and modify it according to their needs (in case they want to run amd64 workloads).
>
> 2.  _Does this mean you prefer to adopt the volume (instance types/preferences) approach for clusters using worker nodes of mixed architectures?_
> - I do not have a personal opinion on this matter. My primary focus is to enable the existing features for the s390x architecture.
>

I see.

> Looking ahead, here are some ideas to address the lack of multi-architecture worker node support (though they may not be fully feasible):
> - Load the templates based on the available architectures across all nodes.

Sounds interesting. What do you mean by "load" here, maybe showing
them in the UI?

> - Import the data volumes selectively.
>

If all worker nodes have the same architecture, this should be already
done. You are welcome to chime in
on https://issues.redhat.com/browse/CNV-52175?focusedId=26209275&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26209275

> In my opinion, enabling multi-architecture worker nodes would require a design change that necessitates consensus.
>

akrejcir:
@0xFelix , can you please take a look?

internal/template-bundle/bundle_test.go
```diff
@@ -30,6 +36,12 @@ var _ = Describe("Template bundle", Ordered, func() {
 		nameSuffix = retrieveSuffix()
 	})
 
+	BeforeEach(func() {
+		tmpDir = GinkgoT().TempDir()
+		archIndependentFile = tmpDir + fmt.Sprintf("/common-templates-%s.yaml", common_templates.Version)
+		archDependentFile = tmpDir + fmt.Sprintf("/common-templates-%s-%s.yaml", runtime.GOARCH, common_templates.Version)
```
0xFelix, line 42: Use filepath.Join

internal/template-bundle/bundle_test.go
```diff
@@ -75,6 +87,37 @@ var _ = Describe("Template bundle", Ordered, func() {
 		Expect(ds2.Spec.Source.PVC.Name).To(Equal("win10"))
 		Expect(ds2.Spec.Source.PVC.Namespace).To(Equal("kubevirt-os-images"))
 	})
+
+	It("should throw an error retrieving the bundle file", func() {
+		_, err := RetrieveCommonTemplatesBundleFile(tmpDir)
+		Expect(err.Error()).To(ContainSubstring("failed to find common-templates bundles, none of the files were found"))
```
0xFelix, line 93:
```suggestion
		Expect(err).To(MatchError(ContainSubstring("failed to find common-templates bundles, none of the files were found")))
```

internal/template-bundle/bundle_test.go
nestoracunablanco: Done

internal/template-bundle/bundle_test.go
nestoracunablanco: Done

0xFelix (approved):
/lgtm

PR #1162, opened 2024-12-05
feat: webhook: Check if DataImportCronTemplate is correct

akrejcir:
Added webhook code to check if DataImportCronTemplate is correct. 
The code is copied from CDI webhook. They don't export the code, so we cannot import it as a module.

Fixes: https://issues.redhat.com/browse/CNV-48792

**Release note**:
```release-note
None
```

### DIFF
--- a/webhooks/data-import-cron.go
+++ b/webhooks/data-import-cron.go
@@ -1,0 +1,216 @@
+package webhooks
+
+import (
+	"fmt"
+	neturl "net/url"
+	"reflect"
+
+	cronexpr "github.com/robfig/cron/v3"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
+	cdiv1beta1 "kubevirt.io/containerized-data-importer-api/pkg/apis/core/v1beta1"
+
+	sspv1beta2 "kubevirt.io/ssp-operator/api/v1beta2"
+)
+
+// These validations were mostly copied from CDI code, so that SSP catches some validation
+// errors in the webhook, before trying to create an invalid DataImportCron.
+// https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/apiserver/webhooks/dataimportcron-validate.go
+
+func validateDataImportCronTemplate(cron *sspv1beta2.DataImportCronTemplate) error {
+	if nameErrors := validation.IsDNS1035Label(cron.Name); len(nameErrors) > 0 {
+		return fmt.Errorf("invalid name: %q", nameErrors)
+	}
+
+	if cron.Spec.Template.Spec.Source == nil || cron.Spec.Template.Spec.Source.Registry == nil {
+		return fmt.Errorf("missing registry source")
+	}
+
+	if cron.Spec.Template.Spec.SourceRef != nil ||
+		cron.Spec.Template.Spec.ContentType != "" ||
+		len(cron.Spec.Template.Spec.Checkpoints) > 0 ||
+		cron.Spec.Template.Spec.FinalCheckpoint {
+		return fmt.Errorf("unsettable fields: SourceRef, ContentType, Checkpoints, FinalCheckpoint")
+	}
+
+	if err := validateDataVolumeSpec(&cron.Spec.Template.Spec); err != nil {
+		return fmt.Errorf("invalid template.spec: %w", err)
+	}
+
+	if cron.Spec.Schedule != "" {
+		if _, err := cronexpr.ParseStandard(cron.Spec.Schedule); err != nil {
+			return fmt.Errorf("illegal cron schedule")
+		}
+	}
+
+	if cron.Spec.ImportsToKeep != nil && *cron.Spec.ImportsToKeep < 0 {
+		return fmt.Errorf("illegal ImportsToKeep value")
+	}
+
+	if cron.Spec.GarbageCollect != nil &&
+		*cron.Spec.GarbageCollect != cdiv1beta1.DataImportCronGarbageCollectNever &&
+		*cron.Spec.GarbageCollect != cdiv1beta1.DataImportCronGarbageCollectOutdated {
+		return fmt.Errorf("illegal GarbageCollect value")
+	}
+
+	if dataSourceNameErrors := validation.IsDNS1035Label(cron.Spec.ManagedDataSource); len(dataSourceNameErrors) > 0 {
+		return fmt.Errorf("invalid managedDataSource: %q", dataSourceNameErrors)
+	}
+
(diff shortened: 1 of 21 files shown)


webhooks/data-import-cron.go
```diff
@@ -0,0 +1,217 @@
...
+	default:
+		return fmt.Errorf(`unsupported value: "%s": supported values: "ReadOnlyMany", "ReadWriteMany", "ReadWriteOnce", "ReadWriteOncePod"`, string(accessMode))
+	}
+	return nil
+}
+
+func validateNumberOfSources(source *cdiv1beta1.DataVolumeSource) error {
+	numberOfSources := 0
+	s := reflect.ValueOf(source).Elem()
+	for i := 0; i < s.NumField(); i++ {
```
0xFelix, line 170: nit: Use `for i := range s.NumField()`

webhooks/data-import-cron.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("source registry should have either URL or ImageStream")
+	}
+
+	if sourceURL != nil {
+		url, err := neturl.Parse(*sourceURL)
+		if err != nil {
+			return fmt.Errorf("illegal registry source URL %s: %w", *sourceURL, err)
+		}
+
+		scheme := url.Scheme
```
0xFelix, line 197: nit: Inline this var

webhooks/data-import-cron.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	if numberOfSources > 1 {
+		return fmt.Errorf("multiple DataVolume sources")
+	}
+	return nil
+}
+
+func validateDataVolumeSourceRegistry(sourceRegistry *cdiv1beta1.DataVolumeSourceRegistry) error {
+	sourceURL := sourceRegistry.URL
+	sourceIS := sourceRegistry.ImageStream
```
0xFelix, line 186: Up to you but I'd inline most of these vars

webhooks/data-import-cron_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+			dict.Spec.Template.Spec.PVC.AccessModes = []v1.PersistentVolumeAccessMode{
+				"unknown-access-mode",
+			}
+
+			err := ValidateDataImportCronTemplate(dict)
+			Expect(err).To(MatchError(ContainSubstring("unsupported value:")))
+		})
+
+		It("should fail if there are multiple access modes in PVC", func() {
```
0xFelix, line 162: This case is missing for Storage?

0xFelix: Can the cases not be split e.g. like other ones?

webhooks/ssp_webhook.go
```diff
@@ -173,11 +173,11 @@ func (s *sspValidator) validateOperandPlacement(ctx context.Context, namespace s
 	return s.apiClient.Create(ctx, deployment, &client.CreateOptions{DryRun: []string{metav1.DryRunAll}})
 }
 
-// TODO: also validate DataImportCronTemplates in general once CDI exposes its own validation
 func validateDataImportCronTemplates(ssp *sspv1beta2.SSP) error {
 	for _, cron := range ssp.Spec.CommonTemplates.DataImportCronTemplates {
-		if cron.Name == "" {
-			return fmt.Errorf("missing name in DataImportCronTemplate")
+		err := ValidateDataImportCronTemplate(&cron)
+		if err != nil {
```
0xFelix, line 179: nit: inline

webhooks/data-import-cron.go
akrejcir: Done.
TIL that `range` can have `int` argument. I assumed only slices and maps can be there.

webhooks/data-import-cron.go
akrejcir: done

webhooks/data-import-cron.go
akrejcir: I'd keep them this way. It would be too verbose if inlined.

webhooks/data-import-cron_test.go
akrejcir: > Can the cases not be split e.g. like other ones?

What do you mean?

webhooks/ssp_webhook.go
akrejcir: done

webhooks/data-import-cron.go
akrejcir: But `golangci-lint` complains about it:
```
golangci/golangci-lint info installed /workspace/bin/golangci-lint
/workspace/bin/golangci-lint run --timeout 5m
webhooks/data-import-cron.go:170:17: cannot range over s.NumField() (value of type int) (typecheck)
	for i := range s.NumField() {
	               ^
make: *** [Makefile:327: lint] Error 1
```

Even if it obviously works: https://go.dev/play/p/KbogyWbaifX?v=goprev

akalenyu:
> > The code is copied from CDI webhook. They don't export the code, so we cannot import it as a module.
> 
> Isn't exportable at all? What would be @akalenyu's opinion about this?

Yeah I don't think there is a good reason to export it. We did export the authorization bits at one point https://github.com/kubevirt/containerized-data-importer/pull/2636 but that is a must for any project that embeds DataVolumes for security reasons.

My general thoughts are, with all the recent focus on webhooks, I am not sure we want to copy the webhook,
when we could instead just propagate the creation error properly to the status. Would that not suffice?
Another option is dry running the creation.

webhooks/data-import-cron.go
0xFelix: IIRC that was introduced with golang 1.22. :)

0xFelix:
I like the idea of using the dry running option.

@akrejcir Can you explain how having these checks in SSP fix the initial issue? Won't it still fail its reconciliation because the supplied DIC templates are not valid?

webhooks/data-import-cron.go
jcanocan: Maybe the problem is that the linter is already a bit old. IIRC, we are using `golangci-lint` 1.55.2, which is based in golang 1.21.3. Could you please try to update the `golangci-lint` version up to 1.62.2 and try again?

akrejcir: I will do it in a separate PR.

akrejcir:
The error is already reported in the status.

I will post another PR to improve the error message, so it is clear which object caused the failure.

akrejcir:
Here is the new PR: https://github.com/kubevirt/ssp-operator/pull/1231

PR #1165, opened 2024-12-09
Update common-templates to v0.32.0

ksimon1:
Update common-templates bundle to v0.32.0 
**Release note**:
```release-note
Update common-templates bundle to v0.32.0
```


0xFelix:
This file contains all architectures (s390x,amd64). I don't think that is what we want for now. The files providing only a single architecture are missing from the common-templates release and this PR.

@fossedihelm

PR #1199, opened 2025-01-08
feat: Add VM delete protection

jcanocan:
VirtualMachine objects are often managed by automation, CLI commands, 3rd party tools, etc. These automations may result in deleting accidentally VMs that should have not been deleted. These deletions may lead to a service degradation or out of service. Moreover, the deleted VMs may lead to information loss if the underlining PVC is deleted as a result of a cascaded delete.

It adds the ability to protect VirtualMachine objects from being deleted. If the label `kubevirt.io/vm-delete-protection` is set to `True`, any attempt to delete the VM will be rejected by a VAP policy.

This protection enables a protection against non-intended VM deletions, providing security and confidence to users.

Fixes # [CNV-50741](https://issues.redhat.com/browse/CNV-50741)

**Release note**:

```release-note
Enables delete protection to VirtualMachine objects
```

### DIFF
--- a/internal/common/labels.go
+++ b/internal/common/labels.go
@@ -22,11 +22,12 @@ func (a AppComponent) String() string {
 }
 
 const (
-	AppComponentMonitoring      AppComponent = "monitoring"
-	AppComponentSchedule        AppComponent = "schedule"
-	AppComponentTemplating      AppComponent = "templating"
-	AppComponentTektonPipelines AppComponent = "tektonPipelines"
-	AppComponentTektonTasks     AppComponent = "tektonTasks"
+	AppComponentMonitoring           AppComponent = "monitoring"
+	AppComponentSchedule             AppComponent = "schedule"
+	AppComponentTemplating           AppComponent = "templating"
+	AppComponentTektonPipelines      AppComponent = "tektonPipelines"
+	AppComponentTektonTasks          AppComponent = "tektonTasks"
+	AppComponentVMDeletionProtection AppComponent = "vmDeleteProtection"
 )
 
 // AddAppLabels to the provided obj
--- a/internal/operands/vm-delete-protection/reconcile.go
+++ b/internal/operands/vm-delete-protection/reconcile.go
@@ -1,0 +1,105 @@
+package vm_delete_protection
+
+import (
+	"fmt"
+
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/operands"
+)
+
+// Define RBAC rules needed by this operand:
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingadmissionpolicies;validatingadmissionpolicybindings,verbs=get;list;create;watch;update;delete
+
+const (
+	operandName                              = "vm-delete-protection"
+	operandComponent                         = common.AppComponentVMDeletionProtection
+	virtualMachineDeleteProtectionPolicyName = "kubevirt-vm-deletion-protection"
+)
+
+func init() {
+	utilruntime.Must(admissionregistrationv1.AddToScheme(common.Scheme))
+}
+
+func WatchClusterTypes() []operands.WatchType {
+	return []operands.WatchType{
+		{Object: &admissionregistrationv1.ValidatingAdmissionPolicy{}},
+		{Object: &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}},
+	}
+}
+
+type VMDeleteProtection struct{}
+
+var _ operands.Operand = &VMDeleteProtection{}
+
+func New() operands.Operand { return &VMDeleteProtection{} }
+
+func (v *VMDeleteProtection) WatchTypes() []operands.WatchType { return nil }
+
+func (v *VMDeleteProtection) WatchClusterTypes() []operands.WatchType { return WatchClusterTypes() }
+
+func (v *VMDeleteProtection) Reconcile(request *common.Request) ([]common.ReconcileResult, error) {
+	return common.CollectResourceStatus(request,
+		reconcileVAP,
+		reconcileVAPB,
+	)
+}
+
+func (v *VMDeleteProtection) Cleanup(request *common.Request) ([]common.CleanupResult, error) {
+	return common.DeleteAll(request,
+		newValidatingAdmissionPolicy(),
+		newValidatingAdmissionPolicyBinding(),
+	)
+}
+
+func (v *VMDeleteProtection) Name() string { return operandName }
+
+func reconcileVAP(request *common.Request) (common.ReconcileResult, error) {
+	return common.CreateOrUpdate(request).
+		ClusterResource(newValidatingAdmissionPolicy()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAP := found.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			expectedVAP := expected.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+
+			foundVAP.Spec = expectedVAP.Spec
+		}).
+		StatusFunc(func(resource client.Object) common.ResourceStatus {
+			vap := resource.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			if vap.Status.TypeChecking == nil {
+				msg := fmt.Sprintf("Delete protection VAP type checking in progress")
+				return common.ResourceStatus{
+					Progressing:  &msg,
+					NotAvailable: &msg,
+					Degraded:     &msg,
+				}
+			}
+
+			if len(vap.Status.TypeChecking.ExpressionWarnings) != 0 {
+				msg := fmt.Sprintf("Incorrect VM delete protection VAP CEL expression %v",
+					vap.Status.TypeChecking)
+				return common.ResourceStatus{
+					NotAvailable: &msg,
+					Degraded:     &msg,
+				}
+			}
+			return common.ResourceStatus{}
+		}).
+		Reconcile()
+}
+
+func reconcileVAPB(request *common.Request) (common.ReconcileResult, error) {
+	return common.CreateOrUpdate(request).
+		ClusterResource(newValidatingAdmissionPolicyBinding()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAPB := found.(*admissionregistrationv1.ValidatingAdmissionPolicyBinding)
+			expectedVAPB := expected.(*admissionregistrationv1.ValidatingAdmissionPolicyBinding)
+
+			foundVAPB.Spec = expectedVAPB.Spec
+		}).
+		Reconcile()
+}
--- a/internal/operands/vm-delete-protection/reconcile_test.go
+++ b/internal/operands/vm-delete-protection/reconcile_test.go
@@ -1,0 +1,134 @@
+package vm_delete_protection
+
+import (
+	"context"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/google/cel-go/cel"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	ssp "kubevirt.io/ssp-operator/api/v1beta2"
+	"kubevirt.io/ssp-operator/internal/common"
+	. "kubevirt.io/ssp-operator/internal/test-utils"
+)
+
+var _ = Describe("VM delete protection operand", func() {
+	const (
+		namespace = "kubevirt"
+		name      = "test-ssp"
+	)
+
+	var (
+		request common.Request
+		operand = New()
+		key     = client.ObjectKey{Name: virtualMachineDeleteProtectionPolicyName}
+	)
+
+	BeforeEach(func() {
+		client := fake.NewClientBuilder().WithScheme(common.Scheme).Build()
+		request = common.Request{
+			Request: reconcile.Request{
+				NamespacedName: types.NamespacedName{
+					Namespace: namespace,
+					Name:      name,
+				},
+			},
+			Client:  client,
+			Context: context.Background(),
+			Instance: &ssp.SSP{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "SSP",
+					APIVersion: ssp.GroupVersion.String(),
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: namespace,
+				},
+			},
+			VersionCache: common.VersionCache{},
+		}
+	})
+
+	It("should create VM deletion protection resources", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(newValidatingAdmissionPolicy(), request)
+		ExpectResourceExists(newValidatingAdmissionPolicyBinding(), request)
+	})
+
+	It("should update VAP spec if changed", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		vap := &admissionregistrationv1.ValidatingAdmissionPolicy{}
+
+		Expect(request.Client.Get(request.Context, key, vap)).To(Succeed())
+
+		vap.Spec.Variables = []admissionregistrationv1.Variable{
+			{
+				Name:       "test-variable",
+				Expression: `test-expression`,
+			},
+		}
+
+		Expect(request.Client.Update(request.Context, vap)).ToNot(HaveOccurred())
+
+		_, err = operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(request.Client.Get(request.Context, key, vap)).To(Succeed())
+		Expect(vap.Spec).To(Equal(newValidatingAdmissionPolicy().Spec))
+
+	})
+
+	It("should update VAPB spec if changed", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		vapb := &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}
+
+		Expect(request.Client.Get(request.Context, key, vapb)).To(Succeed())
+
+		vapb.Spec.ValidationActions = []admissionregistrationv1.ValidationAction{
+			admissionregistrationv1.Warn,
+		}
+
+		Expect(request.Client.Update(request.Context, vapb)).To(Succeed())
+
+		_, err = operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		Expect(request.Client.Get(request.Context, key, vapb)).To(Succeed())
+		Expect(vapb.Spec).To(Equal(newValidatingAdmissionPolicyBinding().Spec))
+	})
+
+	It("should create one valid CEL expression", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		vap := &admissionregistrationv1.ValidatingAdmissionPolicy{}
+
+		Expect(request.Client.Get(request.Context, key, vap)).To(Succeed())
+		Expect(vap.Spec.Validations).To(HaveLen(1))
+
+		celEnv, err := cel.NewEnv()
+		Expect(err).ToNot(HaveOccurred())
+
+		_, issues := celEnv.Parse(vap.Spec.Validations[0].Expression)
+		Expect(issues.Err()).ToNot(HaveOccurred())
+	})
+})
+
+func TestVMDeleteProtection(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "VM Delete Protection Suite")
+}
--- a/internal/operands/vm-delete-protection/resources.go
+++ b/internal/operands/vm-delete-protection/resources.go
@@ -1,0 +1,69 @@
+package vm_delete_protection
+
+import (
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	kubevirt "kubevirt.io/api/core"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+)
+
+const vmDeleteProtectionCELExpression = `(!has(oldObject.metadata.labels) || !(variables.label in oldObject.metadata.labels) || !oldObject.metadata.labels[variables.label].matches('^(true|True)$'))`
+
+func newValidatingAdmissionPolicy() *admissionregistrationv1.ValidatingAdmissionPolicy {
+	var apiVersions []string
+	for _, version := range kubevirtv1.ApiSupportedVersions {
+		apiVersions = append(apiVersions, version.Name)
+	}
+
+	return &admissionregistrationv1.ValidatingAdmissionPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: virtualMachineDeleteProtectionPolicyName,
+		},
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicySpec{
+			FailurePolicy: ptr.To(admissionregistrationv1.Fail),
+			MatchConstraints: &admissionregistrationv1.MatchResources{
+				ResourceRules: []admissionregistrationv1.NamedRuleWithOperations{
+					{
+						RuleWithOperations: admissionregistrationv1.RuleWithOperations{
+							Operations: []admissionregistrationv1.OperationType{
+								admissionregistrationv1.Delete,
+							},
+							Rule: admissionregistrationv1.Rule{
+								APIGroups:   []string{kubevirt.GroupName},
+								APIVersions: apiVersions,
+								Resources:   []string{"virtualmachines"},
+							},
+						},
+					},
+				},
+			},
+			Variables: []admissionregistrationv1.Variable{
+				{
+					Name:       "label",
+					Expression: `string('kubevirt.io/vm-delete-protection')`,
+				},
+			},
+			Validations: []admissionregistrationv1.Validation{
+				{
+					Expression:        vmDeleteProtectionCELExpression,
+					MessageExpression: `'VirtualMachine ' + string(oldObject.metadata.name) + ' cannot be deleted, remove delete protection'`,
+				},
+			},
+		},
+	}
+}
+
+func newValidatingAdmissionPolicyBinding() *admissionregistrationv1.ValidatingAdmissionPolicyBinding {
+	return &admissionregistrationv1.ValidatingAdmissionPolicyBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: virtualMachineDeleteProtectionPolicyName,
+		},
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicyBindingSpec{
+			PolicyName: virtualMachineDeleteProtectionPolicyName,
+			ValidationActions: []admissionregistrationv1.ValidationAction{
+				admissionregistrationv1.Deny,
+			},
+		},
+	}
+}
--- a/tests/vm_delete_protection_test.go
+++ b/tests/vm_delete_protection_test.go
@@ -1,0 +1,90 @@
+package tests
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/rand"
+	kubevirtv1 "kubevirt.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"kubevirt.io/ssp-operator/tests/env"
+)
+
+const deleteProtectionLabel = "kubevirt.io/vm-delete-protection"
+
+var _ = Describe("VM delete protection", func() {
+
+	var vm *kubevirtv1.VirtualMachine
+
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
+		if vm != nil {
+			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+			Expect(err).To(Or(Not(HaveOccurred()), MatchError(errors.IsNotFound, "errors.IsNotFound")))
+
+			if err == nil {
+				Eventually(func() error {
+					if err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm); err != nil {
+						return err
+					}
+
+					vm.Labels[deleteProtectionLabel] = "False"
+					return apiClient.Update(ctx, vm)
+				}, env.ShortTimeout(), time.Second).Should(Succeed())
+
+				Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+				waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
+				vm = nil
+			}
+		}
+	})
+
+	DescribeTable("should not allow to delete a VM if the protection is enabled", func(labelValue string) {
+		vm = createVMWithDeleteProtection(labelValue)
+
+		Expect(apiClient.Delete(ctx, vm)).To(
+			MatchError(ContainSubstring("VirtualMachine %v cannot be deleted, remove delete protection", vm.Name)))
+	},
+		Entry("using True as value", "True"),
+		Entry("using true as value", "true"),
+	)
+
+	DescribeTable("should be able to delete a VM if the protection is disabled", func(labelValue string) {
+		vm = createVMWithDeleteProtection(labelValue)
+
+		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+	},
+		Entry("using False as value", "False"),
+		Entry("using false as value", "false"),
+		Entry("using value different from false or False", "niceValue"),
+	)
+
+	It("should be able to delete a VM if the VM does not have any label", func() {
+		vm = createVMWithLabels(nil)
+
+		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+	})
+})
+
+func createVMWithDeleteProtection(protected string) *kubevirtv1.VirtualMachine {
+	return createVMWithLabels(map[string]string{deleteProtectionLabel: protected})
+}
+
+func createVMWithLabels(labels map[string]string) *kubevirtv1.VirtualMachine {
+	vmName := fmt.Sprintf("testvmi-%v", rand.String(10))
+	vmi := NewMinimalVMIWithNS(strategy.GetNamespace(), vmName)
+	vm := NewVirtualMachine(vmi)
+
+	vm.Labels = labels
+	eventuallyCreateVm(vm)
+
+	return vm
+}
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -27,6 +27,18 @@ rules:
 - apiGroups:
   - admissionregistration.k8s.io
   resources:
+  - validatingadmissionpolicies
+  - validatingadmissionpolicybindings
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
+  - admissionregistration.k8s.io
+  resources:
   - validatingwebhookconfigurations
   verbs:
   - create
--- a/data/olm-catalog/ssp-operator.clusterserviceversion.yaml
+++ b/data/olm-catalog/ssp-operator.clusterserviceversion.yaml
@@ -95,6 +95,18 @@ spec:
         - apiGroups:
           - admissionregistration.k8s.io
           resources:
+          - validatingadmissionpolicies
+          - validatingadmissionpolicybindings
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - update
+          - watch
+        - apiGroups:
+          - admissionregistration.k8s.io
+          resources:
           - validatingwebhookconfigurations
           verbs:
           - create
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/fsnotify/fsnotify v1.8.0
 	github.com/go-logr/logr v1.4.2
+	github.com/google/cel-go v0.22.1
 	github.com/kubevirt/monitoring/pkg/metrics/parser v0.0.0-20230706095033-373a95665d5a
 	github.com/machadovilaca/operator-observability v0.0.24
 	github.com/onsi/ginkgo/v2 v2.22.0
@@ -61,7 +62,6 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
-	github.com/google/cel-go v0.22.1 // indirect
 	github.com/google/gnostic-models v0.6.9 // indirect
 	github.com/google/go-cmp v0.6.0 // indirect
 	github.com/google/gofuzz v1.2.0 // indirect
--- a/internal/controllers/setup.go
+++ b/internal/controllers/setup.go
@@ -19,6 +19,7 @@ import (
 	tekton_cleanup "kubevirt.io/ssp-operator/internal/operands/tekton-cleanup"
 	template_validator "kubevirt.io/ssp-operator/internal/operands/template-validator"
 	vm_console_proxy "kubevirt.io/ssp-operator/internal/operands/vm-console-proxy"
+	vm_delete_protection "kubevirt.io/ssp-operator/internal/operands/vm-delete-protection"
 	template_bundle "kubevirt.io/ssp-operator/internal/template-bundle"
 	vm_console_proxy_bundle "kubevirt.io/ssp-operator/internal/vm-console-proxy-bundle"
 )
@@ -69,6 +70,7 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		common_instancetypes.New(),
 		data_sources.New(templatesBundle.DataSources),
 		tekton_cleanup.New(),
+		vm_delete_protection.New(),
 	}
 
 	if runningOnOpenShift {
--- a/tests/cleanup_test.go
+++ b/tests/cleanup_test.go
@@ -18,6 +18,7 @@ import (
 	tekton_cleanup "kubevirt.io/ssp-operator/internal/operands/tekton-cleanup"
 	template_validator "kubevirt.io/ssp-operator/internal/operands/template-validator"
 	vm_console_proxy "kubevirt.io/ssp-operator/internal/operands/vm-console-proxy"
+	vm_delete_protection "kubevirt.io/ssp-operator/internal/operands/vm-delete-protection"
 )
 
 var _ = Describe("Cleanup", func() {
@@ -40,6 +41,7 @@ var _ = Describe("Cleanup", func() {
 			template_validator.WatchClusterTypes,
 			vm_console_proxy.WatchClusterTypes,
 			tekton_cleanup.WatchClusterTypes,
+			vm_delete_protection.WatchClusterTypes,
 		} {
 			allWatchTypes = append(allWatchTypes, f()...)
 		}


internal/controllers/vap_controller.go
```diff
@@ -0,0 +1,180 @@
...
+			},
+			Variables: []admissionregistrationv1.Variable{
+				{
+					Name:       "label",
+					Expression: `string('kubevirt.io/vm-delete-protection')`,
+				},
+			},
+			Validations: []admissionregistrationv1.Validation{
+				{
+					Expression:        `(!has(oldObject.metadata.labels) || !(variables.label in oldObject.metadata.labels) || !oldObject.metadata.labels[variables.label].matches('^(true|True)$'))`,
```
akrejcir, line 174: Would it be enough to test that the label exists, instead of checking its value?

akrejcir (commented):
Why did you create a separate controllers instead of adding these new resources into an operand? We usually add new resources in operands. That way, they can be configured in the SSP resource.

Can you rename this PR and the commit? This is definitely not a `chore`.
Maybe something like:
```
feat: Add VM delete protection
```

jcanocan:
> Why did you create a separate controllers instead of adding these new resources into an operand? We usually add new resources in operands. That way, they can be configured in the SSP resource.

It is my understanding that operands are better fit when you are using CRD than controllers. In this case, we are just using a build-in feature such as VAP and VAPB. Given that, I'm not particularly against adding it though an operand. In fact, I question this myself multiple times. As we consider best. 

> Can you rename this PR and the commit? This is definitely not a `chore`. Maybe something like:
> 
> ```
> feat: Add VM delete protection
> ```

You are absolutely right. Thanks.

internal/controllers/vap_controller.go
jcanocan: I've considered clearer from the user point of view. My idea is: by forcing the user to add the "True" value, we make sure that user really wants the protection enabled, and I think it will avoid confusions. That being said, I'm not against it if we find it out better.

akrejcir:
> > Why did you create a separate controllers instead of adding these new resources into an operand? We usually add new resources in operands. That way, they can be configured in the SSP resource.
> 
> It is my understanding that operands are better fit when you are using CRD than controllers. In this case, we are just using a build-in feature such as VAP and VAPB. Given that, I'm not particularly against adding it though an operand. In fact, I question this myself multiple times. As we consider best.

I'm not sure what do you mean by "operands are better fit when you are using CRD than controllers". Can you explain it more?

The SSP controller is just deploys various resources based on what is configured in the SSP CR. VAP and VAPB are one of these resources. Currently we don't configure them, but maybe we can. Operands are an abstraction used to group related resources together in the code, so it is easier to understand. For example we deploy multiple ClusterRole objects, and it would be harder to understand if they were all in one package.

jcanocan:
> > > Why did you create a separate controllers instead of adding these new resources into an operand? We usually add new resources in operands. That way, they can be configured in the SSP resource.
> > 
> > 
> > It is my understanding that operands are better fit when you are using CRD than controllers. In this case, we are just using a build-in feature such as VAP and VAPB. Given that, I'm not particularly against adding it though an operand. In fact, I question this myself multiple times. As we consider best.
> 
> I'm not sure what do you mean by "operands are better fit when you are using CRD than controllers". Can you explain it more?
> 
> The SSP controller deploys various resources based on what is configured in the SSP CR. VAP and VAPB are one of these resources. Currently we don't configure them, but maybe we can. Operands are an abstraction used to group related resources together in the code, so it is easier to understand. For example we deploy multiple ClusterRole objects, and it would be harder to understand if they were all in one package.
> 
> I would say that this is the exact use case to add a new operand.

All right! You convinced me! Thanks. Let's create an operand.

jcanocan:
v2:
- Dropped vap and vapb controllers.
- Added a new operand to handle both vap and vapb resources.

internal/common/labels.go
```diff
@@ -22,11 +22,12 @@ func (a AppComponent) String() string {
...
-	AppComponentSchedule        AppComponent = "schedule"
-	AppComponentTemplating      AppComponent = "templating"
-	AppComponentTektonPipelines AppComponent = "tektonPipelines"
-	AppComponentTektonTasks     AppComponent = "tektonTasks"
+	AppComponentMonitoring          AppComponent = "monitoring"
+	AppComponentSchedule            AppComponent = "schedule"
+	AppComponentTemplating          AppComponent = "templating"
+	AppComponentTektonPipelines     AppComponent = "tektonPipelines"
+	AppComponentTektonTasks         AppComponent = "tektonTasks"
+	AppComponetVMDeletionProtection              = "VMDeletionProtection"
```
akrejcir, line 30: Please keep this consistent, and fix the typo:
```
AppComponentVMDeletionProtection AppComponent = "vmDeletionProtection"
```

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"kubevirt.io/ssp-operator/internal/common"
+	"kubevirt.io/ssp-operator/internal/operands"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Define RBAC rules needed by this operand:
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingadmissionpolicies,verbs=get;list;create;watch;update
+// +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=validatingadmissionpolicybindings,verbs=get;list;create;watch;update
```
akrejcir, line 13: These 2 lines can be combined into one line.
We also need deletion permissions, so that the operator can clean up, when it's removed.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+
+const (
+	operandName                                     = "vm-delete-protection"
+	operandComponent                                = common.AppComponetVMDeletionProtection
+	virtualMachineDeleteProtectionPolicyName        = "ssp-vm-deletion-protection-policy"
+	virtualMachineDeleteProtectionPolicyBindingName = "ssp-vm-deletion-protection-binding"
```
akrejcir, line 19: Maybe we can use fully qualified names for the cluster resources, like `ssp.kubevirt.io/vm-deletion-protection` ?

Can you please remove the suffixes `-policy` and `binding`? It is redundant, because we know the object type.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (v *VMDeleteProtection) Reconcile(request *common.Request) ([]common.ReconcileResult, error) {
+	return common.CollectResourceStatus(request,
+		reconcileVAP,
+		reconcileVAPB,
+	)
+}
+
+func (v *VMDeleteProtection) Cleanup(request *common.Request) ([]common.CleanupResult, error) {
+	return common.DeleteAll(request)
```
akrejcir, line 51: This function takes as parameter objects that should be deleted.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+
+func reconcileVAP(request *common.Request) (common.ReconcileResult, error) {
+	return common.CreateOrUpdate(request).
+		ClusterResource(NewVMDeleteProtectionValidatingAdmissionPolicy()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAP := found.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			expectedVAP := expected.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+
+			foundVAP.Labels = expectedVAP.Labels
```
akrejcir, line 64: Labels are copied already for all objects. They don't need to be part of the UpdateFunc(). 
See here: https://github.com/kubevirt/ssp-operator/blob/992331739e363a4e92f7c10b2724b39ab64f783d/internal/common/resource.go#L178

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+
+func reconcileVAPB(request *common.Request) (common.ReconcileResult, error) {
+	return common.CreateOrUpdate(request).
+		ClusterResource(NewVMDeleteProtectionValidatingAdmissionPolicyBinding()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAPB := found.(*admissionregistrationv1.ValidatingAdmissionPolicyBinding)
+			expectedVAPB := expected.(*admissionregistrationv1.ValidatingAdmissionPolicyBinding)
+
+			foundVAPB.Labels = expectedVAPB.Labels
```
akrejcir, line 78: No need to copy labels, same as above.

internal/operands/vm-delete-protection/reconcile_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	It("should create VM deletion protection resources", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		ExpectResourceExists(NewVMDeleteProtectionValidatingAdmissionPolicy(), request)
+		ExpectResourceExists(NewVMDeleteProtectionValidatingAdmissionPolicyBinding(), request)
+	})
+
+	DescribeTable("should update labels if changed", func(obj client.Object, objName string) {
```
akrejcir, line 70: These tests are incorrect. SSP should not remove user defined labels and annotations.
This functionality is partially covered in the tests for `internal/common`: https://github.com/kubevirt/ssp-operator/blob/f8a7389aa787fb1712fe54c63b09848f33ac208c/internal/common/resource_test.go#L72

internal/operands/vm-delete-protection/reconcile_test.go
```diff
@@ -0,0 +1,151 @@
...
+	)
+
+	It("should update VAP spec if changed", func() {
+		_, err := operand.Reconcile(&request)
+		Expect(err).ToNot(HaveOccurred())
+
+		vap := &admissionregistrationv1.ValidatingAdmissionPolicy{}
+		key := client.ObjectKey{Name: virtualMachineDeleteProtectionPolicyName}
+
+		Expect(request.Client.Get(request.Context, key, vap)).ToNot(HaveOccurred())
```
akrejcir, line 105: Use `.To(Succeed())` instead.

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,69 @@
...
+	return &admissionregistrationv1.ValidatingAdmissionPolicy{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ValidatingAdmissionPolicy",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: virtualMachineDeleteProtectionPolicyName,
+		},
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicySpec{
+			FailurePolicy: &failPolicy,
```
akrejcir, line 20: Use `ptr.To(admissionregistrationv1.Fail)` instead of a variable.

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,69 @@
...
+			Variables: []admissionregistrationv1.Variable{
+				{
+					Name:       "label",
+					Expression: `string('kubevirt.io/vm-delete-protection')`,
+				},
+			},
+			Validations: []admissionregistrationv1.Validation{
+				{
+					Expression:        `(!has(oldObject.metadata.labels) || !(variables.label in oldObject.metadata.labels) || !oldObject.metadata.labels[variables.label].matches('^(true|True)$'))`,
+					MessageExpression: `'VirtualMachine ' + string(oldObject.metadata.name) + ' cannot be deleted, remove delete protection'`,
```
akrejcir, line 46: It would be useful to add some parsing and checking for this expression so that unit tests would fail if there is a syntax error.

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,69 @@
...
+				},
+			},
+		},
+	}
+}
+
+func NewVMDeleteProtectionValidatingAdmissionPolicyBinding() *admissionregistrationv1.ValidatingAdmissionPolicyBinding {
```
akrejcir, line 53: This can be private.

tests/vm_deletion_protection_test.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	})
+
+	It("should not allow to delete a VM if the protection is enabled", func() {
+		createVMWithDeleteProtection("True")
+
+		err := apiClient.Delete(ctx, vm)
+
+		Expect(err).To(
+			MatchError(ContainSubstring(fmt.Sprintf("VirtualMachine %v cannot be deleted, remove delete protection", vm.Name))))
```
akrejcir, line 58: ContainSubstring() already takes format string parameter, so no need to call fmt.Sprintf()

tests/vm_deletion_protection_test.go
```diff
@@ -0,0 +1,83 @@
...
+		createVMWithDeleteProtection("False")
+
+		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
+	})
+
+	It("should be able to delete a VM if the VM does not have any label", func() {
+		createVMWithDeleteProtection("True")
+
+		Eventually(func(g Gomega) error {
```
akrejcir, line 71: No need to use function with (g Gomega) parameter, because it is not used.

akrejcir:
Please also add the new operand to the `tests/cleanup_test.go`, so it will check that cleanup works.

internal/common/labels.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: I've used `ssp.kubevirt.io-vm-delete-protection`, because `/` symbol is not allowed in object names. However, I'm not convinced about it, it sounds weird imho.

jcanocan: Dropped suffixes. Without them, both names are the same, so I've also combined the strings into a single one.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,82 @@
...
+		ClusterResource(NewVMDeleteProtectionValidatingAdmissionPolicy()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAP := found.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			expectedVAP := expected.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+
+			foundVAP.Labels = expectedVAP.Labels
+			foundVAP.Spec = expectedVAP.Spec
+		}).
+		Reconcile()
```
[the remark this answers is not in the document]
jcanocan: It just has the typeChecker and generation fields. Since we are providing the cel expression and the user would not be able to change it, there is nothing to check.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	ssp "kubevirt.io/ssp-operator/api/v1beta2"
+	"kubevirt.io/ssp-operator/internal/common"
+	. "kubevirt.io/ssp-operator/internal/test-utils"
+)
+
+var log = logf.Log.WithName("vm_delete_protection_operand")
```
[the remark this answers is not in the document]
jcanocan: Dropped.

internal/operands/vm-delete-protection/reconcile_test.go
jcanocan: I've dropped the test, is there anything additional that we should cover in a separate test about labels in your opinion?

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,69 @@
+package vm_delete_protection
+
+import (
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func NewVMDeleteProtectionValidatingAdmissionPolicy() *admissionregistrationv1.ValidatingAdmissionPolicy {
```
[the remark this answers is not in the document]
jcanocan: Adjusted.

internal/operands/vm-delete-protection/resources.go
jcanocan: Done.

internal/operands/vm-delete-protection/resources.go
jcanocan: I've tried my best to do this using `cel-go`. Unfortunately, I was not able to implement such thing. VAP includes a build-in typechecking. Maybe we can add a functional test checking if this filed is nil. WDYT?

internal/operands/vm-delete-protection/resources.go
jcanocan: Adjusted.

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+	}
+}
+
+func NewVMDeleteProtectionValidatingAdmissionPolicyBinding() *admissionregistrationv1.ValidatingAdmissionPolicyBinding {
+	return &admissionregistrationv1.ValidatingAdmissionPolicyBinding{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "ValidatingAdmissionPolicyBinding",
+		},
```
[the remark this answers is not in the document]
jcanocan: Dropped.

tests/vm_deletion_protection_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kubevirtv1 "kubevirt.io/api/core/v1"
```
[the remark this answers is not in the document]
jcanocan: Done.

tests/vm_deletion_protection_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"kubevirt.io/ssp-operator/tests/env"
+)
+
+var _ = Describe("VM delete protection", func() {
+	const deleteProtectionLabel = "kubevirt.io/vm-delete-protection"
+
+	var vm *kubevirtv1.VirtualMachine
+
+	var createVMWithDeleteProtection = func(protected string) {
```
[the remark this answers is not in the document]
jcanocan: Done.

tests/vm_deletion_protection_test.go
```diff
@@ -0,0 +1,83 @@
...
+		vm.Labels = make(map[string]string)
+		vm.Labels[deleteProtectionLabel] = protected
+		eventuallyCreateVm(vm)
+	}
+
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
```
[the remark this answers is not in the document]
jcanocan: Done.

tests/vm_deletion_protection_test.go
```diff
@@ -0,0 +1,83 @@
...
+				vm.Labels[deleteProtectionLabel] = "False"
+				return apiClient.Update(ctx, vm)
+			}, env.ShortTimeout(), time.Second).Should(Succeed())
+		}
+	})
+
+	It("should not allow to delete a VM if the protection is enabled", func() {
+		createVMWithDeleteProtection("True")
+
+		err := apiClient.Delete(ctx, vm)
```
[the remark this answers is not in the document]
jcanocan: Done.

tests/vm_deletion_protection_test.go
jcanocan: Ah, you right. Thanks. Done.

tests/vm_deletion_protection_test.go
jcanocan: Dropped.

jcanocan:
> Please also add the new operand to the `tests/cleanup_test.go`, so it will check that cleanup works.

Added.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: I understand. What about `kubevirt-vm-delete-protection` or `ssp-kubevirt-vm-delete-protection`?

internal/operands/vm-delete-protection/resources.go
akrejcir: I see. I've tried it also, but it is not simple. For now, can you only do a simple check with parser?
```
celEnv, err := cel.NewEnv()
if err != nil {
	// handle error
}
_, issues := celEnv.Parse(expression)
if issues != nil && issues.Err() != nil {
	// handle error
}
```

You can probably put this code in the `New()` when the operand is created.
And because the `expression` is hard coded, we can panic if parsing fails.

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,87 @@
...
+					return err
+				}
+
+				vm.Labels[deleteProtectionLabel] = "False"
+				return apiClient.Update(ctx, vm)
+			}, env.ShortTimeout(), time.Second).Should(Succeed())
+
+			Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+			waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
+		}
```
akrejcir, line 43: Can you simply Delete the VM, without calling Get and Update?
```
if vm != nil {
	Expect(apiClient.Delete(ctx, vm)).To(Or(Succeed(), MatchError(errors.IsNotFound, "errors.IsNotFound")))
	vm = nil
}

```

internal/operands/vm-delete-protection/reconcile.go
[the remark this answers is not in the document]
akrejcir: It also has `.status.conditions` that may be useful.
Let's check the `.status.typeChceking` at least, because we cannot do it easily locally. You can use `.StatusFunc()` for that. Here are examples of default status functions:
https://github.com/kubevirt/ssp-operator/blob/992331739e363a4e92f7c10b2724b39ab64f783d/internal/common/resource.go#L498

internal/operands/vm-delete-protection/reconcile_test.go
akrejcir: No. IIUC there is no special behavior for labels in VAP or VAPB.

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,69 @@
...
+					{
+						RuleWithOperations: admissionregistrationv1.RuleWithOperations{
+							Operations: []admissionregistrationv1.OperationType{
+								admissionregistrationv1.Delete,
+							},
+							Rule: admissionregistrationv1.Rule{
+								APIGroups:   []string{"kubevirt.io"},
+								APIVersions: []string{"*"},
+								Resources:   []string{"virtualmachines"},
+							},
```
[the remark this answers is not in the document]
akrejcir: Can you also use the loop to include all supported versions, not only the latest? Otherwise, the validation will only be triggered for `v1` not `v1alpha3`.

internal/operands/vm-delete-protection/resources.go
akrejcir: Can you also rename it to be shorter? 
```
newValidatingAdmissionPolicyBinding()
```

We have only one in this package, so we don't need to specify that it is for VM deletion protection.

internal/operands/vm-delete-protection/reconcile.go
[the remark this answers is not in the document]
jcanocan: Done. I'm unsure which fields of `ResourceStatus` should be filled if and issue is found. Please, take a look.

internal/operands/vm-delete-protection/resources.go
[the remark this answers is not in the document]
jcanocan: Done.

internal/operands/vm-delete-protection/resources.go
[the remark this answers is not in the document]
jcanocan: Done.

internal/operands/vm-delete-protection/resources.go
jcanocan: Done.

internal/operands/vm-delete-protection/resources.go
jcanocan: Done.

tests/vm_delete_protection_test.go
jcanocan: That would be risky, IMHO. If for some reason, a VM stays with the protection and policy enabled, the test suite won't be able to delete the VM. In this way, we are sure that the protection for these particular VMs is removed.

akrejcir: Right, of course. Ignore my previous comment, I didn't realize that.

internal/operands/vm-delete-protection/reconcile.go
akrejcir: IMHO the first one is better.

internal/operands/vm-delete-protection/reconcile.go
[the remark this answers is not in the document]
akrejcir: Also `Degraded`. I think `Progressing` is not needed, because there is no progress.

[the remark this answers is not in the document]
jcanocan: All right. Done.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Adjusted.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+type VMDeleteProtection struct{}
+
+var _ operands.Operand = &VMDeleteProtection{}
+
+func New() operands.Operand {
+	if err := checkCelExpression(); err != nil {
+		panic("Invalid VM delete protection CEL expression")
```
akrejcir, line 38: Show also the error message in panic. It will be useful for debugging.

akrejcir:
Can you also remove unneeded text from the PR's description?

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

akrejcir:
Thanks, looks good now.

/lgtm

akrejcir:
> > Can you also remove unneeded text from the PR's description?
> 
> Adjusted. Is it better now?

These is still the empty section: `Special notes for your reviewer:` , if you have nothing to write there, can you remove it?

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,98 @@
+package vm_delete_protection
+
+import (
+	"fmt"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
```
0xFelix, line 5: nit: Import order (e.g. stdlib first, then external, then local pkgs)

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+}
+
+type VMDeleteProtection struct{}
+
+var _ operands.Operand = &VMDeleteProtection{}
+
+func New() operands.Operand {
+	if err := checkCelExpression(); err != nil {
+		panic(fmt.Errorf("invalid VM delete protection CEL expression %v", err))
```
0xFelix, line 38: We should be careful with panicking. Also, this adds a new import for a rather complex package. Are we sure we need that?

IMO we are better of failing the reconciliation when creating/updating a VAP is rejected by the admitter. I also see `vap.Status.TypeChecking` being evaluated below.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,98 @@
...
+		ClusterResource(newValidatingAdmissionPolicy()).
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAP := found.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			expectedVAP := expected.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+
+			foundVAP.Spec = expectedVAP.Spec
+		}).
+		StatusFunc(func(resource client.Object) common.ResourceStatus {
+			status := common.ResourceStatus{}
```
0xFelix, line 74: Inline to L82

internal/operands/vm-delete-protection/resources.go
```diff
@@ -0,0 +1,85 @@
...
+		Spec: admissionregistrationv1.ValidatingAdmissionPolicyBindingSpec{
+			PolicyName: virtualMachineDeleteProtectionPolicyName,
+			ValidationActions: []admissionregistrationv1.ValidationAction{
+				admissionregistrationv1.Deny,
+			},
+		},
+	}
+}
+
+func checkCelExpression() error {
```
0xFelix, line 72: Not sure about the value of this. Seems problematic to me.

0xFelix (commented):
I would strongly suggest against adding the CEL check / panicking if the CEL is not what we expect. Afterall the expression is controlled by us and test cases should ensure it works as expected.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Maybe we can keep the unit test performed over this expression, and drop the panic. In this way, we will have some kind of feedback in case the expression would become incorrect (maybe because something in CEL syntax is changed or deprecated).

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
akrejcir: I think this is a good use of panic, because it is a programmer's error.  We can move this check to a unit test.

The package is already imported, by another dependency. We were depending on it even before this PR.

jcanocan: Adjusted.

0xFelix: I see, it was an indirect import before. Still, I think it makes more sense to simply verify deployed VAPs work as expected.

internal/operands/vm-delete-protection/reconcile_test.go
```diff
@@ -0,0 +1,127 @@
+package vm_delete_protection
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/cel-go/cel"
```
0xFelix, line 7: nit: import order

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,87 @@
...
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
+		err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+		if !errors.IsNotFound(err) {
+			Eventually(func() error {
+				err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+				if err != nil {
```
0xFelix, line 33: nit: inline

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	var vm *kubevirtv1.VirtualMachine
+
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
+		err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+		if !errors.IsNotFound(err) {
```
0xFelix, line 30: This check looks suspicious. What if another error occured?

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,87 @@
...
+		vm = createVMWithDeleteProtection("True")
+
+		Eventually(func() error {
+			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+			if err != nil {
+				return err
+			}
+			vm.Labels = nil
+			return apiClient.Update(ctx, vm)
+		}, env.ShortTimeout(), time.Second).Should(Succeed())
```
0xFelix, line 70: Can't you simply create a VM without the label in the first place?

tests/vm_delete_protection_test.go
jcanocan: Added an assertion in case other error has occurred.

tests/vm_delete_protection_test.go
jcanocan: The idea behind this test is to check the corner case that the VMs does not have any labels at all `labels == nil`. I can create the VM without the `kubevirt.io/vm-delete-protection`, but it does not guarantee that the VM has some default label, e.g., `kubevirt.io/vm`.

tests/vm_delete_protection_test.go
jcanocan: done

internal/operands/vm-delete-protection/reconcile_test.go
jcanocan: Done.

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
+		err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+		Expect(err).To(Or(Not(HaveOccurred()), MatchError(errors.IsNotFound, "VM not found")))
+
+		if !errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
jcanocan: Yes. Done.

tests/vm_delete_protection_test.go
jcanocan: Changed to create the VM without any labels.

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	It("should be able to delete a VM if the protection is disabled", func() {
+		vm = createVMWithDeleteProtection(map[string]string{deleteProtectionLabel: "False"})
+
+		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
+	})
+
+	It("should be able to delete a VM if the VM does not have any label", func() {
+		vm = createVMWithDeleteProtection(nil)
```
0xFelix, line 62: The name `createVMWithDeleteProtection` no longer holds true. Can you make this func take a bool argument and extract the basic VM generation logic to another helper and use it here and in `createVMWithDeleteProtection`?

jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,96 @@
...
+		WithAppLabels(operandName, operandComponent).
+		UpdateFunc(func(expected, found client.Object) {
+			foundVAP := found.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			expectedVAP := expected.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+
+			foundVAP.Spec = expectedVAP.Spec
+		}).
+		StatusFunc(func(resource client.Object) common.ResourceStatus {
+			vap := resource.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			if vap.Status.TypeChecking != nil && len(vap.Status.TypeChecking.ExpressionWarnings) != 0 {
```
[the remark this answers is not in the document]
jcanocan: Really nice catch! Added. PTAL.

internal/operands/vm-delete-protection/reconcile_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Expect(vapb.Spec).To(Equal(newValidatingAdmissionPolicyBinding().Spec))
+	})
+
+	It("should create a valid CEL expression", func() {
+		celEnv, err := cel.NewEnv()
+		Expect(err).ToNot(HaveOccurred())
+
+		_, issues := celEnv.Parse(vmDeleteProtectionCELExpression)
+		Expect(issues.Err()).ToNot(HaveOccurred())
+	})
```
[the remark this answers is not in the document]
jcanocan: Adjusted.

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	var vm *kubevirtv1.VirtualMachine
+
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
+		err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
+		Expect(err).To(Or(Not(HaveOccurred()), MatchError(errors.IsNotFound, "VM not found")))
```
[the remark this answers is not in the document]
jcanocan: Ah, I see. Thanks. Adjusted.

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,89 @@
...
+	},
+		Entry("using True as value", "True"),
+		Entry("using true as value", "true"),
+	)
+
+	DescribeTable("should be able to delete a VM if the protection is disabled", func(labelValue string) {
+		vm = createVMWithDeleteProtection(labelValue)
+
+		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
```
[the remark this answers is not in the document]
jcanocan: Dropped

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,89 @@
...
+		Entry("using False as value", "False"),
+		Entry("using false as value", "false"),
+		Entry("using value different from false or False", "niceValue"),
+	)
+
+	It("should be able to delete a VM if the VM does not have any label", func() {
+		vm = createVMWithLabels(nil)
+
+		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
```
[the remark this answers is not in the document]
jcanocan: Dropped

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,89 @@
...
+var _ = Describe("VM delete protection", func() {
+
+	var vm *kubevirtv1.VirtualMachine
+
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	AfterEach(func() {
+		err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
```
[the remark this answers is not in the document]
jcanocan: Added.

tests/vm_delete_protection_test.go
```diff
@@ -0,0 +1,89 @@
...
+				if err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm); err != nil {
+					return err
+				}
+
+				vm.Labels[deleteProtectionLabel] = "False"
+				return apiClient.Update(ctx, vm)
+			}, env.ShortTimeout(), time.Second).Should(Succeed())
+
+			Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+			waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
```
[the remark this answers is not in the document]
jcanocan: Added.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,101 @@
...
+			foundVAP := found.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			expectedVAP := expected.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+
+			foundVAP.Spec = expectedVAP.Spec
+		}).
+		StatusFunc(func(resource client.Object) common.ResourceStatus {
+			vap := resource.(*admissionregistrationv1.ValidatingAdmissionPolicy)
+			if vap.Status.TypeChecking == nil {
+				return common.ResourceStatus{
+					Progressing: ptr.To("Delete protection VAP type checking in progress"),
```
akrejcir, line 75: Also the other fields need to be set. If the object is progressing, it means that it is not yet available and it is degraded.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,101 @@
...
+				}
+			} else if vap.Status.TypeChecking != nil && len(vap.Status.TypeChecking.ExpressionWarnings) != 0 {
```
akrejcir, line 77: The above block returns, so `else` and `vap.Status.TypeChecking != nil` are not needed.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
jcanocan: Done.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func WatchClusterTypes() []operands.WatchType {
+	return []operands.WatchType{
+		{Object: &admissionregistrationv1.ValidatingAdmissionPolicy{}},
+		{Object: &admissionregistrationv1.ValidatingAdmissionPolicyBinding{}},
+	}
+}
+
+type VMDeleteProtection struct{}
```
codingben, line 34: nit: I'd prefer to name it as `virtualMachineDeleteProtection`.

internal/common/labels.go
```diff
@@ -22,11 +22,12 @@ func (a AppComponent) String() string {
...
-	AppComponentSchedule        AppComponent = "schedule"
-	AppComponentTemplating      AppComponent = "templating"
-	AppComponentTektonPipelines AppComponent = "tektonPipelines"
-	AppComponentTektonTasks     AppComponent = "tektonTasks"
+	AppComponentMonitoring           AppComponent = "monitoring"
+	AppComponentSchedule             AppComponent = "schedule"
+	AppComponentTemplating           AppComponent = "templating"
+	AppComponentTektonPipelines      AppComponent = "tektonPipelines"
+	AppComponentTektonTasks          AppComponent = "tektonTasks"
+	AppComponentVMDeletionProtection AppComponent = "vmDeleteProtection"
```
codingben, line 30: nit: I'd suggest to stick to `AppComponentVmDeleteProtection` instead of `AppComponentVMDeletionProtection`.

internal/operands/vm-delete-protection/reconcile.go
```diff
@@ -0,0 +1,105 @@
...
+					NotAvailable: &msg,
+					Degraded:     &msg,
+				}
+			}
+			return common.ResourceStatus{}
+		}).
+		Reconcile()
+}
+
+func reconcileVAPB(request *common.Request) (common.ReconcileResult, error) {
```
codingben, line 94: nit: I'd name it as `reconcileVAPBinding` and also the variables below then should be `foundVapBinding` and `expectedVapBinding`.

codingben (approved):
I'm late as a reviewer due to sick leave. I think it looks good and can be merged, just a few nits about naming.

/lgtm

jcanocan:
@codingben Sorry, I didn't have the time to address your comments before it got merged. I will in a follow-up.

codingben:
> @codingben Sorry, I didn't have the time to address your comments before it got merged. I will in a follow-up.

No need to sorry because it was only nits and I came late to the review cycle.

iholder101:
Hey @jcanocan,

Too bad I wasn't aware of this PR before it merged.
The design proposal for this excact approach was rejected with many concerns from multiple people: https://github.com/kubevirt/community/pull/363.

I find it concerning that it was merged without addressing the concerns that were raised up there and that no one of the reviewers were pinged to look at this PR.

Am I missing something? Were the concerns addressed in any way?
I think we need to consider reverting this before users can lay hand on it.

iholder101:
> To be honest, I wasn't, still isn't, clear to me which repositories are covered by the design proposals and VEPs. As far as I know, only Kubevirt/Kubevirt is affected. Therefore, I understood that those concerns wouldn't affect to side projects like SSP and the scope of reviewers were just SSP maintainers/contributors.

I'm not referring strictly to formalities, such as whether VEPs apply to this repository, but rather to the fact that this proposal faced considerable pushback for reasons that aren’t discussed or even mentioned in this PR.

This feature raised concerns by multiple people upstream, wouldn't it make sense to consider its potential impact on downstream usage as well?
At the very least, shouldn't we have a more in-depth discussion about these concerns before making it available to downstream customers and committing to its implementation?

In general, adopting downstream-only solutions, especially if they were rejected upstream, is risky and goes again the upstream-first principle. They are sometimes justified, but in these cases I'd expect a clear explanation to why they are absolutely necessary.

PR #1212, opened 2025-01-15
doc: Update disclaimer in readTemplates

nestoracunablanco:
This update explains why the check of the teamplateArch is still needed. It maintains compatibility in case changes for other architectures are delivered solely through the generic common templates.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/template-bundle/bundle.go
+++ b/internal/template-bundle/bundle.go
@@ -73,9 +73,9 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
 		if !ok {
 			return nil, err
 		}
-		// DISCLAIMER: This is a temporary solution for delivering templates related to the host architecture.
-		// Once the common templates are released based on architecture, this change will no longer be necessary.
-		// Instead, a modification will be required in setup.go to specify the bundle to read from.
+		// The template bundles are delivered separately based on architecture.
+		// However, in cases where the generic template bundle includes architectures that are
+		// not released separately, this filter can still be useful.
 		if templateArch == runtime.GOARCH {
 			bundle = append(bundle, template)
 		}


internal/template-bundle/bundle.go
```diff
@@ -73,9 +73,10 @@ func readTemplates(filename string) ([]templatev1.Template, error) {
 		if !ok {
 			return nil, err
 		}
-		// DISCLAIMER: This is a temporary solution for delivering templates related to the host architecture.
-		// Once the common templates are released based on architecture, this change will no longer be necessary.
-		// Instead, a modification will be required in setup.go to specify the bundle to read from.
+		// DISCLAIMER: This change is necessary for templates related to unsupported
+		// architectures. By implementing this, we ensure that we always filter by
+		// architecture. Additionally, the releases of a new architecture template
+		// can be managed separately from the generic template.
```
akrejcir, line 79: I'm unsure what you want to say here. 
The comment starts with `This change is necessary...`, but nothing is changing in this piece of code. It only filters templates by architecture.

akrejcir:
> /retest

You don't need to run CI for this PR, because it only changes a comment. Even if CI fails because of a flake, we don't need to spend resources to run it again.

nestoracunablanco:
> > /retest
> 
> You don't need to run CI for this PR, because it only changes a comment. Even if CI fails because of a flake, we don't need to spend resources to run it again.

Understood. I thought it was a strict requirement for merging that all tests must pass.

nestoracunablanco:
@akrejcir, what are your thoughts on the new comment? If you're still not satisfied with it, I can drop the PR."

akrejcir:
I would remove the `DISCLAIMER`, because I think this is not a disclaimer. Otherwise it looks good, thanks.
Sorry for later review.

nestoracunablanco:
> I would remove the `DISCLAIMER`, because I think this is not a disclaimer. Otherwise it looks good, thanks. Sorry for later review.

Done.

PR #1215, opened 2025-01-16
do-not-merge: test the arch of the images with Ubuntu latest

nestoracunablanco:
Note: Please do not merge this pull request. It is solely a test to verify that the container images are correctly pushed to quay.io, specifically against my local quay namespace.

Context: Publishing with the older version of Ubuntu resulted in a manifest containing two amd64 images. While the cause of the error is not fully understood, I have tested the process using the latest version of Ubuntu, and the error has been resolved. This test aims to confirm that the behavior of the latest Ubuntu version from GitHub Actions is consistent.

Once the test is complete, this pull request will be closed.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -41,3 +41,25 @@ jobs:
         go-version-file: 'go.mod'
     - name: Check commited files
       run: hack/check-commited-changes.sh
+
+  publish_operator_image-test-only:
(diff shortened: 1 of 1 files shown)


nestoracunablanco:
I don't have the necessary credentials to run the test. I will look for an alternative approach within my fork.

PR #1218, opened 2025-01-17
Add dependabot configuration for api

jcanocan:

Add dependabot configuration for api. It will weekly update api dependencies.

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,20 @@ updates:
     open-pull-requests-limit: 3
     labels:
       - "release-note-none"
+  - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -14,6 +14,22 @@ updates:
     open-pull-requests-limit: 3
     labels:
       - "release-note-none"
+version: 2
+updates:
```
0xFelix, line 18: Drop these lines, so you only add a new item to the existing `updates`.

jcanocan: Done.

0xFelix (commented):
/approve

PR #1230, opened 2025-01-23
feat: simplify renovate config

ksimon1:
removed go version constraints as they were not working the way we need.
Removed unnecessary rules for packages on release branches. Disabled grouping of the dependencies into single PR - this should help to bump dependencies more easier, because PR with more dependencies, have more chances, that it will introduce incompatible dependencies

feat: simplify renovate config

Fixes #

**Release note**:

```release-note

```


jcanocan:
Thanks @ksimon1!
/lgtm

renovate.json
```diff
@@ -1,12 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "baseBranches": ["main", "release-v0.20", "release-v0.19", "release-v0.18"],
-  "constraints": {
-    "go": "1.22"
-  },
+  "baseBranches": ["main", "/release-v[0-9]\\.[1-9][0-9]/"],
```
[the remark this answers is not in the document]
ksimon1: yes, let me update it

akrejcir:
The regex would have been simpler if we could use negative lookahead: https://regex101.com/r/M4Qexj/1

/approve

ksimon1:
> The regex would have been simpler if we could use negative lookahead: https://regex101.com/r/M4Qexj/1 But only certain regex libraries support it.
> 
> /approve

From https://github.com/renovatebot/renovate/discussions/16472 it looks like renovate does not support it

renovate.json
```diff
@@ -17,32 +12,16 @@
     {
       "groupName": "all dependencies",
       "groupSlug": "all",
-      "matchBaseBranches": ["main"],
-      "enabled": false,
-      "matchPackagePatterns": [
-        "*"
-      ]
-    }, {
       "enabled": false,
```
0xFelix, line 15: Updates are not enabled?

ksimon1: No, RH instance does not support golang (they removed the go support few months ago after we introduced it in our projects). It was not working properly and the updates were not correctly composed.

0xFelix: Are there plans to switch to another renovate instance?

ksimon1: no

0xFelix (approved):
/lgtm

PR #1240, opened 2025-01-27
tests: Set DICT RetentionPolicy to None

akrejcir:
The default retention policy is `All`, which means that DataSource is not removed when DataImportCron is removed. This caused cluster health check to fail.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

codingben (approved):
/lgtm

akrejcir:
The tests need to be changed.

/hold

Issue #1239, opened 2025-01-27
Flake in VM delete protection tests AfterEach


0xFelix:
Looks like the Matcher for `errors.IsNotFound` is not correct, as the call to the client returns a `StatusError` instead.

See flake in this lane:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/kubevirt_ssp-operator/1238/pull-ci-kubevirt-ssp-operator-main-e2e-functests/1883732618336276480#1:build-log.txt%3A1939

/cc @jcanocan

akrejcir:
Error message says:
```
Expected success, but got an error:
```

Using `errors.IsNotFound` is probably ok, because the returned `StatusError` is a not found error:
```
Status: "Failure",
Message: "virtualmachines.kubevirt.io \"testvmi-pmp67t7cmm\" not found",
Reason: "NotFound",
```

0xFelix:
But does the `errors.IsNotFound` matcher detect a `StatusError` as such?

akrejcir:
Yes, that is exactly what it checks:
https://github.com/kubevirt/ssp-operator/blob/a902c1923853221616f35b1efdc41a0c2a1892a0/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L527-L536

https://github.com/kubevirt/ssp-operator/blob/a902c1923853221616f35b1efdc41a0c2a1892a0/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L809-L814

https://github.com/kubevirt/ssp-operator/blob/a902c1923853221616f35b1efdc41a0c2a1892a0/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L74-L78

0xFelix:
Ah, I see. Actually, it is not that matcher that fails.

I suspect there is a race condition where the second call to Get in the Eventually loop breaks.

There is no errors.IsNotFound check there.

jcanocan:
Thanks for raising and taking a look at this issue. 

I think @0xFelix has the point. I can think about two options to address this issue.

1) Wait for VM deletion in the test in which we expect the VM to be deleted.
2) Check if the VM deletion `DeletionTimestamp  != nil` in the AfterEach in:

https://github.com/kubevirt/ssp-operator/blob/main/tests/vm_delete_protection_test.go#L33

IMHO, 1) would is the best option. I will post a PR soon.

jcanocan:
Addressed.
/close

PR #1244, opened 2025-01-28
chore(test): Add cleanup in VM delete protection test

jcanocan:

There is a race condition between the deletion in those test in which the deletion is expected to happen and the `AfterEach` `IsNotFound` check. This may cause that the `AfterEach` tries to update the VM to remove the label, but the VM is already gone.

It adds a `WaitForDeletion` in those test to be sure that the VM is removed before triggering the `AfterEach`. Moreover, it adds a check in the `AfterEach` to check if the VM is already being deleted.

Fixes # https://github.com/kubevirt/ssp-operator/issues/1239

**Release note**:

```release-note
None.
```

### DIFF
--- a/tests/vm_delete_protection_test.go
+++ b/tests/vm_delete_protection_test.go
@@ -30,7 +30,7 @@ var _ = Describe("VM delete protection", func() {
 			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
 			Expect(err).To(Or(Not(HaveOccurred()), MatchError(errors.IsNotFound, "errors.IsNotFound")))
 
-			if err == nil {
+			if err == nil && vm.DeletionTimestamp == nil {
 				Eventually(func() error {
 					if err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm); err != nil {
 						return err
@@ -61,6 +61,7 @@ var _ = Describe("VM delete protection", func() {
 		vm = createVMWithDeleteProtection(labelValue)
 
 		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
 	},
 		Entry("using False as value", "False"),
 		Entry("using false as value", "false"),
@@ -71,6 +72,7 @@ var _ = Describe("VM delete protection", func() {
 		vm = createVMWithLabels(nil)
 
 		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
 	})
 })
 


tests/vm_delete_protection_test.go
```diff
@@ -58,6 +57,11 @@ var _ = Describe("VM delete protection", func() {
 	)
 
 	DescribeTable("should be able to delete a VM if the protection is disabled", func(labelValue string) {
+		DeferCleanup(func() {
```
akrejcir, line 60: No need to use `DeferCleanup` here. The `waitForDeletion` can be called after the `Expect(apiClient.Delete(ctx, vm)).To(Succeed())`. Otherwise it would wait for deletion that did not happen.

Btw combining `DeferCleanup()` and `AfterEach()` in a single test case is not a good idea, because their ordering does not follow tree structure. First all `AfterEach()` are executed and then all `DeferCleanup()`.

tests/vm_delete_protection_test.go
```diff
@@ -29,7 +29,6 @@ var _ = Describe("VM delete protection", func() {
 		if vm != nil {
 			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
 			Expect(err).To(Or(Not(HaveOccurred()), MatchError(errors.IsNotFound, "errors.IsNotFound")))
-
```
akrejcir, line 32: Please also add a check for the deletion timestamp. If it is there, skip the rest of the code.

jcanocan: Done.

tests/vm_delete_protection_test.go
```diff
@@ -61,6 +61,8 @@ var _ = Describe("VM delete protection", func() {
 		vm = createVMWithDeleteProtection(labelValue)
 
 		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
+		vm = nil
```
0xFelix, line 65: Without this line, it should still work if we keep `MatchError(errors.IsNotFound, "errors.IsNotFound")` in the cleanup.

tests/vm_delete_protection_test.go
```diff
@@ -71,6 +73,9 @@ var _ = Describe("VM delete protection", func() {
 		vm = createVMWithLabels(nil)
 
 		Expect(apiClient.Delete(ctx, vm)).To(Succeed())
+		waitForDeletion(client.ObjectKeyFromObject(vm), &kubevirtv1.VirtualMachine{})
+		vm = nil
+
```
0xFelix, line 78: Same + empty line

tests/vm_delete_protection_test.go
jcanocan: Agree, let's try.

tests/vm_delete_protection_test.go
jcanocan: Done.

tests/vm_delete_protection_test.go
```diff
@@ -30,7 +30,7 @@ var _ = Describe("VM delete protection", func() {
 			err := apiClient.Get(ctx, client.ObjectKeyFromObject(vm), vm)
 			Expect(err).To(Or(Not(HaveOccurred()), MatchError(errors.IsNotFound, "errors.IsNotFound")))
 
-			if err == nil {
+			if err == nil && vm.DeletionTimestamp != nil {
```
akrejcir, line 33: If `vm.DeletionTimestamp != nil`, that means that the VM is being deleted, and we should not call `Update`.

0xFelix (commented):
/approve

PR #1256, opened 2025-01-29
fix: regex in renovate config

ksimon1:
fix: regex in renovate config

previous regex caused Renovate to evaluate branches created by renovate as matching. To fix this, the ^ and $ are added to the regex.

**Release note**:
```
NONE
```


0xFelix (approved):
/approve
/lgtm

PR #1277, opened 2025-02-17
cleanup: Remove net.Pipe in port-forwarding in functests

akrejcir:
We can use the data stream directly, instead of using a net.Pipe to copy data to and from it.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/port-forwarding.go
+++ b/tests/port-forwarding.go
@@ -1,6 +1,7 @@
 package tests
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -8,7 +9,6 @@ import (
 	"strconv"
 	"sync/atomic"
 
-	. "github.com/onsi/ginkgo/v2"
 	core "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


tests/port-forwarding.go
```diff
@@ -28,6 +28,17 @@ type portForwarderImpl struct {
 
 var _ PortForwarder = &portForwarderImpl{}
 
+type portForwardConn struct {
+	net.Conn
+	streamConnCloser io.Closer
+}
+
+func (p *portForwardConn) Close() error {
+	err1 := p.Conn.Close()
```
0xFelix, line 37: Inline?

0xFelix (commented):
/approve

tests/port-forwarding.go
akrejcir: I wanted to make sure that the first connection was closed before the second one. But I see that golang defines the order of evaluation of function arguments, unlike other languages, where it is up to the compiler optimization.
https://go.dev/ref/spec#Order_of_evaluation

jcanocan:
Thanks @akrejcir !
/lgtm

akrejcir:
> /retest-required

@jcanocan , I see that this command probably retriggers all tests. In the future, can you use `/retest` instead? It will only rerun failed tests.

PR #1278, opened 2025-02-17
cleanup: Fix various issues found by code inspection

akrejcir:
Fixed:
- Handle unhandled errors
- Removed unused code
- Stopped using deprecated methods
- Code style issues

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/env/env_test.go
+++ b/internal/env/env_test.go
@@ -12,10 +12,11 @@ import (
 
 var _ = Describe("environments", func() {
 	It("should return correct value for OPERATOR_VERSION when variable is set", func() {
-		os.Setenv(OperatorVersionKey, "v0.0.1")
+		Expect(os.Setenv(OperatorVersionKey, "v0.0.1")).To(Succeed())
+		defer func() { Expect(os.Unsetenv(OperatorVersionKey)).To(Succeed()) }()
+
 		res := GetOperatorVersion()
 		Expect(res).To(Equal("v0.0.1"), "OPERATOR_VERSION should equal")
-		os.Unsetenv(OperatorVersionKey)
 	})
 
 	It("should return correct value for OPERATOR_VERSION when variable is not set", func() {
--- a/internal/template-validator/validator/app.go
+++ b/internal/template-validator/validator/app.go
@@ -2,7 +2,6 @@ package validator
 
 import (
 	"crypto/tls"
-	"fmt"
 	"net/http"
 	"os"
 
@@ -130,7 +129,9 @@ func (app *App) Run() {
 
 func registerReadinessProbe() {
 	http.HandleFunc("/readyz", func(resp http.ResponseWriter, req *http.Request) {
-		fmt.Fprintf(resp, "ok")
+		if _, err := resp.Write([]byte("ok")); err != nil {
+			logger.Log.Error(err, "Failed to write response to /readyz")
+		}
 	})
 }
 
--- a/pkg/monitoring/metrics/template-validator/metrics.go
+++ b/pkg/monitoring/metrics/template-validator/metrics.go
@@ -9,8 +9,3 @@ func SetupMetrics() error {
 		templateMetrics,
 	)
 }
-
-// ListMetrics registered prometheus metrics
-func ListMetrics() []operatormetrics.Metric {
-	return operatormetrics.ListMetrics()
-}
--- a/tests/crypto_policy_test.go
+++ b/tests/crypto_policy_test.go
@@ -40,7 +40,7 @@ var _ = Describe("Crypto Policy", func() {
 		}
 
 		// Note that "crypto/tls" does not support setting max tls version to anything below 1.2
-		oldPermutation tlsConfigTestPermutation = tlsConfigTestPermutation{
+		oldPermutation = tlsConfigTestPermutation{
 			openshiftTLSPolicy: &old,
 			allowedConfigs: []clientTLSOptions{
 				{
@@ -51,7 +51,7 @@ var _ = Describe("Crypto Policy", func() {
 			disallowedConfigs: []clientTLSOptions{},
 		}
 
-		intermediatePermutation tlsConfigTestPermutation = tlsConfigTestPermutation{
+		intermediatePermutation = tlsConfigTestPermutation{
 			openshiftTLSPolicy: &intermediate,
 			allowedConfigs: []clientTLSOptions{
 				{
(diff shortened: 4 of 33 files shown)


pkg/monitoring/metrics/template-validator/metrics.go
```diff
@@ -9,8 +9,3 @@ func SetupMetrics() error {
 		templateMetrics,
 	)
 }
-
-// ListMetrics registered prometheus metrics
-func ListMetrics() []operatormetrics.Metric {
-	return operatormetrics.ListMetrics()
-}
```
akrejcir, line 16: @machadovilaca , I've found this unused function and removed it. Is this a bug, that it is not used anywhere?

internal/env/env_test.go
```diff
@@ -12,10 +12,10 @@ import (
 
 var _ = Describe("environments", func() {
 	It("should return correct value for OPERATOR_VERSION when variable is set", func() {
-		os.Setenv(OperatorVersionKey, "v0.0.1")
+		Expect(os.Setenv(OperatorVersionKey, "v0.0.1")).To(Succeed())
 		res := GetOperatorVersion()
 		Expect(res).To(Equal("v0.0.1"), "OPERATOR_VERSION should equal")
-		os.Unsetenv(OperatorVersionKey)
```
0xFelix, line 18: Maybe DeferCleanup that so the Env is always unset?

internal/template-validator/validator/app.go
```diff
@@ -130,7 +130,10 @@ func (app *App) Run() {
 
 func registerReadinessProbe() {
 	http.HandleFunc("/readyz", func(resp http.ResponseWriter, req *http.Request) {
-		fmt.Fprintf(resp, "ok")
+		_, err := fmt.Fprintf(resp, "ok")
+		if err != nil {
```
0xFelix, line 134: Inline it?

internal/template-validator/validator/app.go
```diff
@@ -130,7 +130,10 @@ func (app *App) Run() {
 
 func registerReadinessProbe() {
 	http.HandleFunc("/readyz", func(resp http.ResponseWriter, req *http.Request) {
-		fmt.Fprintf(resp, "ok")
+		_, err := fmt.Fprintf(resp, "ok")
```
0xFelix, line 133: `resp.Write([]byte("ok"))` would probably achieve the same without an additional dependency.

tests/crypto_policy_test.go
```diff
@@ -204,7 +204,7 @@ func getCaCertificate() []byte {
 func tryToAccessEndpoint(pod core.Pod, serviceName string, subpath string, port uint16, tlsConfig clientTLSOptions, insecure bool) (attemptedUrl string, err error) {
 	conn, err := portForwarder.Connect(&pod, port)
 	Expect(err).ToNot(HaveOccurred())
-	defer conn.Close()
+	defer Expect(conn.Close()).To(Succeed())
```
0xFelix, line 207: Will it run if another Expect fails?

tests/port-forwarding.go
```diff
@@ -43,17 +44,19 @@ func (p *portForwarderImpl) Connect(pod *core.Pod, remotePort uint16) (net.Conn,
 	headers.Set(core.PortForwardRequestIDHeader, strconv.Itoa(int(requestId)))
 	errorStream, err := streamConnection.CreateStream(headers)
 	if err != nil {
-		streamConnection.Close()
-		return nil, err
+		closeErr := streamConnection.Close()
```
0xFelix, line 47: Inline?

tests/port-forwarding.go
```diff
@@ -43,17 +44,19 @@ func (p *portForwarderImpl) Connect(pod *core.Pod, remotePort uint16) (net.Conn,
...
+	if err := errorStream.Close(); err != nil {
+		return nil, err
+	}
 
 	headers.Set(core.StreamType, core.StreamTypeData)
 	dataStream, err := streamConnection.CreateStream(headers)
 	if err != nil {
-		streamConnection.Close()
-		return nil, err
+		closeErr := streamConnection.Close()
```
0xFelix, line 58: Same

tests/validator_test.go
```diff
@@ -3,6 +3,7 @@ package tests
 import (
 	"crypto/tls"
 	"crypto/x509"
+	goerrors "errors"
```
0xFelix, line 6: Import `"k8s.io/apimachinery/pkg/api/errors"` as `k8serrors` instead?

internal/env/env_test.go
akrejcir: Done.

internal/template-validator/validator/app.go
akrejcir: done

internal/template-validator/validator/app.go
akrejcir: Good idea, done.

tests/crypto_policy_test.go
akrejcir: Yes, it will work. A panic can happen during panicking. They will just wrap around each other.
https://go.dev/play/p/Eq5sdpNkDl

tests/port-forwarding.go

tests/validator_test.go
akrejcir: done

akrejcir:
In multiple PRs we have problem with the `golangci-lint` timeout. I will look into it...

0xFelix (commented):
/approve

Thanks!

codingben (approved):
/lgtm
/hold

@akrejcir Do you want to hold it until you get a response from @machadovilaca to your question about an unused function?

pkg/monitoring/metrics/template-validator/metrics.go
machadovilaca: In usually call these functions the tooling to generate the docs, but when setting up multiple metrics, in the registry we only need to list from one. That's why the ListMetrics from metrics/ssp-operator is used but not this one

akrejcir: So, this was not a bug, that the function was not called?

machadovilaca: No, we can remove it. On a future release of the operator-observability library, it will support different registries, and then we will improve the docs to setup and list the metrics separately.

akrejcir:
The failures were probably not flakes, but caused by this PR. I'm looking into it.

ksimon1:
@akrejcir the sonarqube failure was introduced in this PR, or it fails for all PRs?

PR #1280, opened 2025-02-18
feat: Remove tekton code

akrejcir:
Removed all code related to tekton.

The tekton cleanup code is no longer needed, because it was released in the previous version.
We expect that users update sequentially, so the cleanup code has already removed old tekton resources from their cluster.

Fixes: https://issues.redhat.com/browse/CNV-51977

**Release note**:
```release-note
Removed all functionality related to tekton.
```


ksimon1:
I can see file `config/crd/bases/ssp.kubevirt.io_ssps.yaml` still contains tekton attributes (it has still v1beta2 api version). Does it needs to be regenerated to have the v1beta2 api version?

0xFelix (approved):
Nice cleanup!

/approve

PR #1291, opened 2025-02-26
[release-v0.22] fix: Add back API v1beta1 and set it to not be served

akrejcir:
When the API definition was missing from CRD, the OLM would reject upgrading from previous version to this one.

Before removing an API version, OLM requires it to be set as not served.

Fixes: https://issues.redhat.com/browse/CNV-57082

**Release note**:
```release-note
None
```


data/crd/ssp.kubevirt.io_ssps.yaml
```diff
@@ -16,6 +16,3261 @@ spec:
...
+                type: boolean
+              phase:
+                description: Phase is the current phase of the deployment
+                type: string
+              targetVersion:
+                description: The desired version of the resource
+                type: string
+            type: object
+        type: object
+    served: true
```
0xFelix, line 3270: Should be not served?

webhooks/ssp_webhook.go
```diff
@@ -32,17 +33,27 @@ import (
...
 
 var ssplog = logf.Log.WithName("ssp-resource")
 
 func Setup(mgr ctrl.Manager) error {
 	err := ctrl.NewWebhookManagedBy(mgr).
+		For(&sspv1beta1.SSP{}).
+		WithValidator(newSspValidator(mgr.GetClient())).
+		Complete()
+	if err != nil {
+		return fmt.Errorf("failed to create webhook for v1beta2.SSP")
```
0xFelix, line 53:
```suggestion
		return fmt.Errorf("failed to create webhook for v1beta1.SSP")
```

But should it actually create a webhook, since it is not served anymore?

webhooks/ssp_webhook.go
```diff
@@ -189,6 +200,22 @@ func newSspValidator(clt client.Client) *sspValidator {
 
 func getSsp(obj runtime.Object) (*sspv1beta2.SSP, error) {
 	switch sspObj := obj.(type) {
+	case *sspv1beta1.SSP:
```
0xFelix, line 203: If it is not served, is this actually possible?

data/crd/ssp.kubevirt.io_ssps.yaml
akrejcir: Thanks for noticing!

webhooks/ssp_webhook.go
akrejcir: I've removed the webhook code. It will not be called, because the API is not served.

webhooks/ssp_webhook.go
akrejcir: removed

0xFelix (commented):
/approve

PR #1329, opened 2025-03-17
tests: fix race when checking recreated resources

akrejcir:
Sometimes, the API server may not be up-to-date and report an object missing. This caused tests to fail, when SSP recreated an object, but API server was still not aware of it.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

jcanocan:
Thanks!
/lgtm

PR #1356, opened 2025-04-10
fix: vm_controller: Don't fail if PVC or PV does not exist

akrejcir:
`vm_controller` does not show error if PVC or PV do not exist.
This can happen, if the VM is newly created and the PVC and PV were not yet created.

Fixes: https://issues.redhat.com/browse/CNV-35800
Fixes: https://issues.redhat.com/browse/CNV-35801

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/controllers/vm_controller.go
+++ b/internal/controllers/vm_controller.go
@@ -106,6 +106,7 @@ func getVmCrd() string {
 }
 
 func (v *vmController) setVmVolumesMetrics(ctx context.Context, vm *kubevirtv1.VirtualMachine) error {
+	var metricWasSet bool
 	for _, volume := range vm.Spec.Template.Spec.Volumes {
 		volumeName := ""
 		if volume.DataVolume != nil {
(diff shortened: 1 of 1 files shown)


akrejcir:
@machadovilaca , is it correct to do this? That the new controller code will not reenqueue the VM, when PVC or PV do not exist.

machadovilaca:
> @machadovilaca , is it correct to do this? That the new controller code will not reenqueue the VM, when PVC or PV do not exist.

but in a few seconds they will exist, no?
and then we would want to correctly populate the metric

0xFelix (commented):
I'm trying to understand, does it mean there will be no more missing PVC/PV errors and metrics?

akrejcir:
> I'm trying to understand, does it mean there will be no more missing PVC/PV errors and metrics?

Yes. IIUC, the metric has no meaning if PVC and PV does not exist. Logging an error, when they are missing just pollutes the log.

PR #1369, opened 2025-04-17
fix: subset of enabled architectures for running e2e tests

ksimon1:
fix: subset of enabled architectures for running e2e tests

This commit adds another variants of arm string to enabled architectures for running e2e tests

Another variant of this PR would be to drop the check completelly

**Release note**:
```release-note
fix: subset of enabled architectures for running e2e tests

```

### DIFF
--- a/tests/tests_suite_test.go
+++ b/tests/tests_suite_test.go
@@ -52,9 +52,11 @@ var (
 )
 
 const (
-	amd64 = "amd64"
-	s390x = "s390x"
-	arm   = "aarch64"
+	amd64   = "amd64"
+	s390x   = "s390x"
+	arm     = "arm"
+	arm64   = "arm64"
+	aarch64 = "aarch64"
 )
 
 type TestSuiteStrategy interface {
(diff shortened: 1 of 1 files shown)


tests/tests_suite_test.go
```diff
@@ -605,7 +606,7 @@ func retrieveNodeArchitecture() string {
 	nodes := nodeList.Items
 	Expect(nodes).NotTo(BeEmpty(), "No nodes found")
 	architecture := nodes[0].Status.NodeInfo.Architecture
-	Expect(architecture).To(BeElementOf([]string{s390x, amd64, arm}), "Unsupported architecture")
+	Expect(architecture).To(BeElementOf([]string{s390x, amd64, arm, aarch64}), "Unsupported architecture")
```
geetikakay, line 609: should we remove all the arch's dependency and checks?

tests/tests_suite_test.go
```diff
@@ -615,7 +616,7 @@ func retrieveQemuMachineType(architecture string) string {
 		return "q35"
 	case s390x:
 		return "s390-ccw-virtio"
-	case arm:
+	case arm, aarch64:
```
geetikakay, line 619: nit: How about 
	case amd64:
		return "q35"
	case s390x:
		return "s390-ccw-virtio"
	default:
		return "virt"

tests/tests_suite_test.go
```diff
@@ -52,9 +52,10 @@ var (
 )
 
 const (
-	amd64 = "amd64"
-	s390x = "s390x"
-	arm   = "aarch64"
+	amd64   = "amd64"
+	s390x   = "s390x"
+	arm     = "arm"
```
geetikakay, line 57: will this be arm64 ?
oc get nodes -o json | jq '.items[].status.nodeInfo.architecture'
"arm64"

tests/tests_suite_test.go
ksimon1: For sure it is one of the possible approach. I would like to give others some space to comment. Maybe there was a reason, why this check was introduced

PR #1394, opened 2025-05-19
[release-v0.18] tests: Set DICT RetentionPolicy to None

akrejcir:
Manual cherry-pick of https://github.com/kubevirt/ssp-operator/pull/1240 .
Package `k8s.io/utils/ptr` is not in this version, so I've used a pointer to variable.

The default retention policy is `All`, which means that DataSource is not removed when DataImportCron is removed. This caused cluster health check to fail.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/dataSources_test.go
+++ b/tests/dataSources_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"fmt"
 	"reflect"
+	"strings"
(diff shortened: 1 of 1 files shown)


0xFelix (commented):
/approve

PR #1405, opened 2025-06-05
api: deprecate v1beta2 and use v1beta3

akrejcir:
Deprecated `v1beta2` API version.

Changed operator code to use `v1beta3` instead.

**Release note**:
```release-note
None
```


akrejcir:
@jcanocan can you please give another `lgtm`? I needed to rebase, so CI would pass.

PR #1407, opened 2025-06-09
fix: Add delete permission to PrometheusRules and ServiceMonitors

akrejcir:
These objects are created and owned by ssp-operator, so it needs to have permission to delete them.
Otherwise their creation fails with an error like this:

```
servicemonitors.monitoring.coreos.com "prometheus-k8s-rules-cnv" is forbidden: cannot set an ownerRef on a resource you can't delete:
```

**Release note**:
```release-note
None
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -153,6 +153,7 @@ rules:
   - servicemonitors
   verbs:
   - create
+  - delete
   - list
   - update
   - watch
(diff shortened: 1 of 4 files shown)


akrejcir:
The `delete` permission was removed nearly 2 years ago in: https://github.com/kubevirt/ssp-operator/pull/616
I'm unsure why the error is shown now, maybe a new kubernetes version requires the permission.

PR #1415, opened 2025-06-17
chore: webhook: Move placement validation to the end

akrejcir:
Placement validation calls API server, so it is better to do it last, when we are sure that the SSP object is correct.

**Release note**:
```release-note
None
```


akrejcir:
No benchmarks, I just wanted to move the most expensive operation last in the webhook.
But it is very small optimization.

PR #1418, opened 2025-06-20
feat: Add common templates for heterogeneous clusters

akrejcir:
Added logic to create common templates for multiple architectures.

Implements a part of the SSP work of: https://github.com/kubevirt/enhancements/pull/49

Jira: https://issues.redhat.com/browse/CNV-61575

**Release note**:
```release-note
Added logic to create common templates for multiple architectures.
```

### DIFF
--- a/internal/controllers/setup.go
+++ b/internal/controllers/setup.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 
 	"github.com/go-logr/logr"
 	v1 "github.com/openshift/api/config/v1"
@@ -48,10 +49,7 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		return nil, fmt.Errorf("failed to check if running on openshift: %w", err)
 	}
 
-	templatesBundleFile, err := template_bundle.RetrieveCommonTemplatesBundleFile(templateBundleDir)
-	if err != nil {
-		return nil, err
-	}
+	templatesBundleFile := filepath.Join(templateBundleDir, fmt.Sprintf("common-templates-%s.yaml", common_templates.Version))
 	templatesBundle, err := template_bundle.ReadBundle(templatesBundleFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read template bundle: %w", err)
--- a/internal/controllers/setup.go
+++ b/internal/controllers/setup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"slices"
 
 	"github.com/go-logr/logr"
 	v1 "github.com/openshift/api/config/v1"
@@ -55,7 +56,7 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		return nil, fmt.Errorf("failed to read template bundle: %w", err)
 	}
 
-	dataSourceNames, err := template_bundle.CollectDataSourceNames(templates)
+	dataSourceCollection, err := template_bundle.CollectDataSources(templates)
 	if err != nil {
 		return nil, fmt.Errorf("failed to collect DataSource names from templates: %w", err)
 	}
@@ -66,16 +67,25 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		return nil, fmt.Errorf("failed to read vm-console-proxy bundle: %w", err)
 	}
 
+	// Temporarily, only create DataSources for one architecture.
+	// TODO: Create all DataSources, when multi-arch DataSource logic is implemented
+	dataSourceNames := slices.Collect(dataSourceCollection.Names())
+
 	sspOperands := []operands.Operand{
 		data_sources.New(dataSourceNames),
 		vm_delete_protection.New(),
 	}
 
 	if runningOnOpenShift {
+		templatesOperand, err := common_templates.New(templates)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create common templates operand: %w", err)
+		}
+
 		sspOperands = append(sspOperands,
 			metrics.New(),
 			template_validator.New(),
-			common_templates.New(templates),
+			templatesOperand,
 			vm_console_proxy.New(vmConsoleProxyBundle),
 		)
 	}
--- a/internal/controllers/setup.go
+++ b/internal/controllers/setup.go
@@ -50,9 +50,14 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 	}
 
 	templatesBundleFile := filepath.Join(templateBundleDir, fmt.Sprintf("common-templates-%s.yaml", common_templates.Version))
-	templatesBundle, err := template_bundle.ReadBundle(templatesBundleFile)
+	templates, err := template_bundle.ReadTemplates(templatesBundleFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read template bundle: %w", err)
+	}
+
+	dataSourceNames, err := template_bundle.CollectDataSourceNames(templates)
+	if err != nil {
+		return nil, fmt.Errorf("failed to collect DataSource names from templates: %w", err)
 	}
 
 	vmConsoleProxyBundlePath := vm_console_proxy_bundle.GetBundlePath()
@@ -62,7 +67,7 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 	}
 
 	sspOperands := []operands.Operand{
-		data_sources.New(templatesBundle.DataSources),
+		data_sources.New(dataSourceNames),
 		vm_delete_protection.New(),
 	}
 
@@ -70,7 +75,7 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		sspOperands = append(sspOperands,
 			metrics.New(),
 			template_validator.New(),
-			common_templates.New(templatesBundle.Templates),
+			common_templates.New(templates),
 			vm_console_proxy.New(vmConsoleProxyBundle),
 		)
 	}
--- a/internal/operands/common-templates/architecture_test.go
+++ b/internal/operands/common-templates/architecture_test.go
@@ -1,0 +1,48 @@
+package common_templates
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	templatev1 "github.com/openshift/api/template/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"kubevirt.io/ssp-operator/internal/architecture"
+)
+
+var _ = Describe("GetTemplateArch()", func() {
+	It("should read architecture from label", func() {
+		const testArch = architecture.ARM64
+		template := &templatev1.Template{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test",
+				Labels: map[string]string{
+					TemplateArchitectureLabel: string(testArch),
+				},
+			},
+		}
+		Expect(GetTemplateArch(template)).To(Equal(testArch))
+	})
+
+	It("should use default architecture if label is missing", func() {
+		template := &templatev1.Template{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test",
+			},
+		}
+		Expect(GetTemplateArch(template)).To(Equal(TemplateDefaultArchitecture))
+	})
+
+	It("should fail, if architecture value is not known", func() {
+		template := &templatev1.Template{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test",
+				Labels: map[string]string{
+					TemplateArchitectureLabel: "unknown-arch",
+				},
+			},
+		}
+		_, err := GetTemplateArch(template)
+		Expect(err).To(MatchError(ContainSubstring("unknown architecture: unknown-arch")))
+	})
+})
--- a/internal/operands/common-templates/constants.go
+++ b/internal/operands/common-templates/constants.go
@@ -1,9 +1,13 @@
 package common_templates
 
+import "kubevirt.io/ssp-operator/internal/architecture"
+
 const (
 	TemplateVersionLabel         = "template.kubevirt.io/version"
 	TemplateTypeLabel            = "template.kubevirt.io/type"
 	TemplateTypeLabelBaseValue   = "base"
+	TemplateArchitectureLabel    = "template.kubevirt.io/architecture"
+	TemplateDefaultArchitecture  = architecture.AMD64
 	TemplateOsLabelPrefix        = "os.template.kubevirt.io/"
 	TemplateFlavorLabelPrefix    = "flavor.template.kubevirt.io/"
 	TemplateWorkloadLabelPrefix  = "workload.template.kubevirt.io/"
--- a/internal/operands/common-templates/reconcile.go
+++ b/internal/operands/common-templates/reconcile.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/blang/semver/v4"
@@ -13,9 +14,11 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"kubevirt.io/ssp-operator/internal/architecture"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	ssp "kubevirt.io/ssp-operator/api/v1beta3"
+	"kubevirt.io/ssp-operator/internal/architecture"
 	"kubevirt.io/ssp-operator/internal/common"
 	"kubevirt.io/ssp-operator/internal/env"
 	"kubevirt.io/ssp-operator/internal/operands"
@@ -41,30 +44,22 @@ func WatchClusterTypes() []operands.WatchType {
 }
 
 type commonTemplates struct {
-	templatesBundle   []templatev1.Template
-	deployedTemplates map[string]bool
+	templatesByArch map[architecture.Arch][]templatev1.Template
 }
 
 var _ operands.Operand = &commonTemplates{}
 
 func New(templates []templatev1.Template) (operands.Operand, error) {
-	var filteredTemplates []templatev1.Template
+	templatesByArch := map[architecture.Arch][]templatev1.Template{}
 	for _, template := range templates {
 		arch, err := GetTemplateArch(&template)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get architecture of template %s: %w", template.Name, err)
 		}
-
-		if arch == defaultArchitecture {
-			filteredTemplates = append(filteredTemplates, template)
-		}
+		templatesByArch[arch] = append(templatesByArch[arch], template)
 	}
 
-	deployedTemplates := make(map[string]bool)
-	for _, t := range filteredTemplates {
-		deployedTemplates[t.Name] = true
-	}
-	return &commonTemplates{templatesBundle: filteredTemplates, deployedTemplates: deployedTemplates}, nil
+	return &commonTemplates{templatesByArch: templatesByArch}, nil
 }
 
 func (c *commonTemplates) Name() string {
@@ -85,7 +80,17 @@ func (c *commonTemplates) WatchTypes() []operands.WatchType {
 }
 
 func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ReconcileResult, error) {
-	reconcileTemplatesResults, err := common.CollectResourceStatus(request, reconcileTemplatesFuncs(c.templatesBundle)...)
+	clusterArchs, err := getClusterArchitectures(request.Instance)
+	if err != nil {
+		return nil, err
+	}
+
+	var templates []templatev1.Template
+	for _, arch := range clusterArchs {
+		templates = append(templates, c.templatesByArch[arch]...)
+	}
+
+	reconcileTemplatesResults, err := common.CollectResourceStatus(request, reconcileTemplatesFuncs(templates)...)
 	if err != nil {
 		return nil, err
 	}
@@ -94,7 +99,7 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.Reconcile
 		incrementTemplatesRestoredMetric(reconcileTemplatesResults, request.Logger)
 	}
 
-	oldTemplateFuncs, err := c.reconcileOlderTemplates(request)
+	oldTemplateFuncs, err := c.deprecateOrDeleteOldTemplates(request, templates, clusterArchs)
 	if err != nil {
 		return nil, err
 	}
@@ -105,6 +110,50 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.Reconcile
 	}
 
 	return append(reconcileTemplatesResults, oldTemplatesResults...), nil
+}
+
+func getClusterArchitectures(sspObj *ssp.SSP) ([]architecture.Arch, error) {
+	if sspObj.Spec.Cluster == nil {
+		if ptr.Deref(sspObj.Spec.EnableMultipleArchitectures, false) {
+			return nil, fmt.Errorf(".spec.cluster cannot be nil, if .spec.enableMultipleArchitectures is true")
+		}
+		return []architecture.Arch{defaultArchitecture}, nil
+	}
+
+	archs := sspObj.Spec.Cluster.WorkloadArchitectures
+	if len(archs) == 0 {
+		archs = sspObj.Spec.Cluster.ControlPlaneArchitectures
+	}
+
+	if len(archs) == 0 {
+		return nil, fmt.Errorf("no architectrues are defined in .spec.cluster")
+	}
+
+	if ptr.Deref(sspObj.Spec.EnableMultipleArchitectures, false) {
+		result := make([]architecture.Arch, 0, len(archs))
+		for _, archStr := range archs {
+			arch, err := architecture.ToArch(archStr)
+			if err != nil {
+				return nil, err
+			}
+			result = append(result, arch)
+		}
+		return result, nil
+	}
+
+	// For single architecture case, we prefer the first of ControlPlaneArchitectures.
+	// If there are none, we use the first of WorkloadArchitectures.
+	archStr := archs[0]
+	if len(sspObj.Spec.Cluster.ControlPlaneArchitectures) > 0 {
+		archStr = sspObj.Spec.Cluster.ControlPlaneArchitectures[0]
+	}
+
+	arch, err := architecture.ToArch(archStr)
+	if err != nil {
+		return nil, err
+	}
+
+	return []architecture.Arch{arch}, nil
 }
 
 func operatorIsUpgrading(request *common.Request) bool {
@@ -127,21 +176,22 @@ func incrementTemplatesRestoredMetric(reconcileResults []common.ReconcileResult,
 
 func (c *commonTemplates) Cleanup(request *common.Request) ([]common.CleanupResult, error) {
 	var objects []client.Object
-	namespace := request.Instance.Spec.CommonTemplates.Namespace
-
 	deprecatedTemplates, err := getDeprecatedTemplates(request)
 	if err != nil {
 		return nil, err
 	}
 
 	for _, obj := range deprecatedTemplates.Items {
-		obj.ObjectMeta.Namespace = namespace
 		objects = append(objects, &obj)
 	}
 
-	for index := range c.templatesBundle {
-		c.templatesBundle[index].ObjectMeta.Namespace = namespace
-		objects = append(objects, &c.templatesBundle[index])
+	ownedTemplates, err := common.ListOwnedResources[templatev1.TemplateList, templatev1.Template](request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list owned templates: %w", err)
+	}
+
+	for _, template := range ownedTemplates {
+		objects = append(objects, &template)
 	}
 
 	return common.DeleteAll(request, objects...)
@@ -184,13 +234,12 @@ func getOldTemplatesLabelSelector() labels.Selector {
 	return labels.NewSelector().Add(*baseRequirement, *versionRequirement)
 }
 
-func (c *commonTemplates) reconcileOlderTemplates(request *common.Request) ([]common.ReconcileFunc, error) {
-	existingTemplates := &templatev1.TemplateList{}
-	err := request.Client.List(request.Context, existingTemplates, &client.ListOptions{
+func (c *commonTemplates) deprecateOrDeleteOldTemplates(request *common.Request, deployedTemplates []templatev1.Template, archs []architecture.Arch) ([]common.ReconcileFunc, error) {
+	oldTemplates := &templatev1.TemplateList{}
+	err := request.Client.List(request.Context, oldTemplates, &client.ListOptions{
 		LabelSelector: getOldTemplatesLabelSelector(),
 		Namespace:     request.Instance.Spec.CommonTemplates.Namespace,
 	})
-
 	// There might not be any templates (in case of a fresh deployment), so a NotFound error is accepted
 	if err != nil && !errors.IsNotFound(err) {
 		return nil, err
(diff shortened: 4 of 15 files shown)


internal/operands/common-templates/reconcile.go
```diff
@@ -167,58 +205,64 @@ func getOldTemplatesLabelSelector() labels.Selector {
...
+		// Delete the template, if it is not one of the cluster architectures
+		if !slices.Contains(archs, GetTemplateArch(&template)) {
+			funcs = append(funcs, reconcileDeleteTemplate(&template))
 			continue
 		}
 
-		// if template has higher version than is defined in ssp operator, keep it as it is. If parsing
-		// of template version fails, continue with adding deprecated label
+		// If template has lower version, than what is defined in ssp operator, deprecate it.
+		// Deprecate also, if version label cannot be parsed.
```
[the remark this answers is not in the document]
akrejcir: Old templates did not have version labels. I wanted to keep backward compatibility in this PR. We can change it in a later PR.

api/v1beta3/ssp_types.go
```diff
@@ -47,11 +47,26 @@ type CommonTemplates struct {
...
+
+	// ControlPlaneArchitectures is a list of control plane architectures supported by the cluster
+	ControlPlaneArchitectures []string `json:"controlPlaneArchitectures,omitempty"`
+}
+
 // SSPSpec defines the desired state of SSP
 type SSPSpec struct {
+	// EnableMultipleArchitectures enables deployment of common Templates,
+	// DataSources and DataImportCrons for multiple node architectures.
+	EnableMultipleArchitectures *bool `json:"enableMultipleArchitectures,omitempty"`
```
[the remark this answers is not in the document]
akrejcir: I don't know how to decide. Personally I prefer pointers for optional fields.

internal/controllers/setup.go
```diff
@@ -63,16 +66,25 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		return nil, fmt.Errorf("failed to read vm-console-proxy bundle: %w", err)
 	}
 
+	// Temporarily, only create DataSources for one architecture.
+	// TODO: Create all DataSources, when multi-arch DataSource logic is implemented
+	dataSourceNames := make([]string, 0, len(dataSourceNamesAndArchs))
```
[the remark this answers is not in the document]
akrejcir: It will have the same functionality. It's just my code style to allocate the length when we know it.

internal/template-bundle/bundle.go
```diff
@@ -103,18 +59,32 @@ func extractDataSources(templates []templatev1.Template) ([]cdiv1beta1.DataSourc
 		}
 
 		namespace, exists := findDataSourceNamespace(template)
-		if !exists {
-			continue
+		// This check is needed, so later code can assume that all DataSources
+		// should be created in the internal.GoldenImagesNamespace
+		if exists && namespace != internal.GoldenImagesNamespace {
```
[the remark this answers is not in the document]
akrejcir: Yes, SSP only creates data sources in that one namespace. But it may happen that common-templates YAML bundle will have different value for `DATA_SOURCE_NAMESPACE` parameter.

internal/controllers/setup.go
```diff
@@ -48,13 +49,15 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		return nil, fmt.Errorf("failed to check if running on openshift: %w", err)
 	}
 
-	templatesBundleFile, err := template_bundle.RetrieveCommonTemplatesBundleFile(templateBundleDir)
+	templatesBundleFile := filepath.Join(templateBundleDir, fmt.Sprintf("common-templates-%s.yaml", common_templates.Version))
```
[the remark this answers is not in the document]
akrejcir: I'd prefer to leave it this way. The string is not used in any other place, and it would be harder to read the code, because reader would need to look at the constant in different place to verify number and types of parameters.

internal/operands/common-templates/architecture_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	templatev1 "github.com/openshift/api/template/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("GetTemplateArch()", func() {
+	It("should read architecture form label", func() {
```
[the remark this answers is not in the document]
akrejcir: done, thanks.

internal/operands/common-templates/reconcile.go
```diff
@@ -90,6 +110,18 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.Reconcile
...
+func getClusterArchitectures(sspObj *ssp.SSP) []string {
+	var clusterArchs []string
+	if sspObj.Spec.EnableMultipleArchitectures != nil && *sspObj.Spec.EnableMultipleArchitectures {
+		clusterArchs = sspObj.Spec.Cluster.WorkloadArchitectures
+	} else if sspObj.Spec.Cluster != nil {
+		clusterArchs = []string{sspObj.Spec.Cluster.ControlPlaneArchitectures[0]}
+	} else {
+		clusterArchs = []string{runtime.GOARCH}
+	}
+	return clusterArchs
```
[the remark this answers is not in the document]
akrejcir: done

internal/operands/common-templates/reconcile.go
```diff
@@ -68,7 +73,22 @@ func (c *commonTemplates) WatchTypes() []operands.WatchType {
...
+	clusterArchs := getClusterArchitectures(request.Instance)
+	if len(clusterArchs) == 0 {
+		return []common.ReconcileResult{}, fmt.Errorf("no cluster architectures defined")
+	}
+
+	var templates []templatev1.Template
+	templatesSet := map[string]struct{}{}
+	for _, arch := range clusterArchs {
+		for i := range c.templatesByArch[arch] {
+			template := &c.templatesByArch[arch][i]
```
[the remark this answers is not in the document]
akrejcir: done

internal/operands/common-templates/reconcile.go
```diff
@@ -167,58 +205,64 @@ func getOldTemplatesLabelSelector() labels.Selector {
...
+
+	existingTemplates := map[string]templatev1.Template{}
+	for _, template := range oldTemplates.Items {
+		existingTemplates[template.Name] = template
+	}
+	for _, template := range ownedTemplates {
+		existingTemplates[template.Name] = template
+	}
+
+	funcs := make([]common.ReconcileFunc, 0, len(existingTemplates))
```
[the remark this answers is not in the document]
akrejcir: done. In this case, the `funcs` will be mostly empty, so preallocating it to larger size is a waste.

api/v1beta3/ssp_types.go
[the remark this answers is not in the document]
0xFelix: Let's keep it as a ptr

internal/operands/common-templates/reconcile_test.go
```diff
@@ -32,23 +33,18 @@ const (
...
+	futureVersion = "v999.999.999"
 )
 
-func TestTemplates(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Common Templates Suite")
-}
-
 var _ = Describe("Common-Templates operand", func() {
+	const defaultArch = runtime.GOARCH
```
0xFelix, line 40: I'd rather set this to a fixed value. Tests should behave the same on all arches.

0xFelix: Possible you could iterate over all possible / supported values?

internal/operands/common-templates/reconcile.go
```diff
@@ -122,9 +150,14 @@ func (c *commonTemplates) Cleanup(request *common.Request) ([]common.CleanupResu
...
-	for index := range c.templatesBundle {
-		c.templatesBundle[index].ObjectMeta.Namespace = namespace
-		objects = append(objects, &c.templatesBundle[index])
+	ownedTemplates, err := common.ListOwnedResources[templatev1.TemplateList, templatev1.Template](request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list owned templates: %w", err)
+	}
+
+	for _, template := range ownedTemplates {
+		template.Namespace = namespace
```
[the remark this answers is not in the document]
akrejcir: thanks, that was a bug. I probably copied the code from above.

internal/operands/common-templates/reconcile_test.go
akrejcir: In `ssp-operator/internal/operands/common-templates/reconcile.go` the function `getClusterArchitectures()` returns `runtime.GOARCH` if nothing else is specified in the SSP object. It is to preserve backward compatibility on clusters different than `amd64`.
Unit tests only copy this logic. 

But I agree, that unit tests should be independent of the arch. I will think how to fix this.

akrejcir: done

tests/commonTemplates_test.go
```diff
@@ -354,6 +328,125 @@ var _ = Describe("Common templates", func() {
...
+		})
+		waitUntilDeployed()
+	})
+
+	AfterAll(func() {
+		strategy.RevertToOriginalSspCr()
+	})
+
+	DescribeTable("created resource", func(res *testResource) {
+		Expect(apiClient.Get(ctx, res.GetKey(), res.NewResource())).To(Succeed())
```
geetikakay, line 400: here we check for resource to be created is it worth to check label TemplateArchitectureLabel here

internal/operands/common-templates/constants.go
```diff
@@ -4,8 +4,11 @@ const (
 	TemplateVersionLabel         = "template.kubevirt.io/version"
 	TemplateTypeLabel            = "template.kubevirt.io/type"
 	TemplateTypeLabelBaseValue   = "base"
+	TemplateArchitectureLabel    = "template.kubevirt.io/architecture"
 	TemplateOsLabelPrefix        = "os.template.kubevirt.io/"
 	TemplateFlavorLabelPrefix    = "flavor.template.kubevirt.io/"
 	TemplateWorkloadLabelPrefix  = "workload.template.kubevirt.io/"
 	TemplateDeprecatedAnnotation = "template.kubevirt.io/deprecated"
+
```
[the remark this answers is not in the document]
akrejcir: done

internal/operands/common-templates/reconcile.go
```diff
@@ -36,18 +42,19 @@ func WatchClusterTypes() []operands.WatchType {
...
 var _ operands.Operand = &commonTemplates{}
 
 func New(templates []templatev1.Template) operands.Operand {
-	deployedTemplates := make(map[string]bool)
-	for _, t := range templates {
-		deployedTemplates[t.Name] = true
+	templatesByArch := map[string][]templatev1.Template{}
+	for _, template := range templates {
+		templateArch := GetTemplateArch(&template)
+		templatesByArch[templateArch] = append(templatesByArch[templateArch], template)
```
0xFelix, line 54: Why not dedup the slices or only add the template if it's not already in the slice? Could be as easy as calling `slices.Sort` and `slices.Compact`. Would save to implement the custom logic to create the set in `Reconcilce`.

internal/operands/common-templates/reconcile.go
```diff
@@ -90,6 +111,16 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.Reconcile
 	return append(reconcileTemplatesResults, oldTemplatesResults...), nil
 }
 
+func getClusterArchitectures(sspObj *ssp.SSP) []string {
+	if sspObj.Spec.EnableMultipleArchitectures != nil && *sspObj.Spec.EnableMultipleArchitectures {
+		return sspObj.Spec.Cluster.WorkloadArchitectures
+	}
+	if sspObj.Spec.Cluster != nil {
```
0xFelix, line 118: `!= nil` doesn't mean it's not empty.

0xFelix: Check for `len(sspObj.Spec.Cluster) > 0`

internal/operands/common-templates/reconcile.go
```diff
@@ -90,6 +111,16 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.Reconcile
 	return append(reconcileTemplatesResults, oldTemplatesResults...), nil
 }
 
+func getClusterArchitectures(sspObj *ssp.SSP) []string {
+	if sspObj.Spec.EnableMultipleArchitectures != nil && *sspObj.Spec.EnableMultipleArchitectures {
```
0xFelix, line 115: Same here? Check for length?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -93,12 +90,21 @@ var _ = Describe("Common-Templates operand", func() {
 			Logger:          log,
 			VersionCache:    common.VersionCache{},
 		}
+
+		filteredTemplates = make([]templatev1.Template, 0, len(testTemplates))
+		for i := range testTemplates {
+			template := &testTemplates[i]
```
0xFelix, line 96: It's okay to assign this in for/range, no?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -32,26 +32,23 @@ const (
...
+	futureVersion = "v999.999.999"
 )
 
-func TestTemplates(t *testing.T) {
-	RegisterFailHandler(Fail)
-	RunSpecs(t, "Common Templates Suite")
-}
-
 var _ = Describe("Common-Templates operand", func() {
+	const amd64 = "amd64"
```
0xFelix, line 39: Why limit these tests to a single arch?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -93,12 +90,21 @@ var _ = Describe("Common-Templates operand", func() {
 			Logger:          log,
 			VersionCache:    common.VersionCache{},
 		}
+
+		filteredTemplates = make([]templatev1.Template, 0, len(testTemplates))
```
0xFelix, line 94: What if this is empty? Should fail the test?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -352,39 +320,132 @@ var _ = Describe("Common-Templates operand", func() {
...
+		BeforeEach(func() {
+			request.Instance.Spec.EnableMultipleArchitectures = ptr.To(true)
+			request.Instance.Spec.Cluster = &ssp.Cluster{
+				WorkloadArchitectures:     []string{amd64, arm64},
+				ControlPlaneArchitectures: []string{amd64},
+			}
+
+			armTemplates = []templatev1.Template{}
+			for _, template := range testTemplates {
+				if GetTemplateArch(&template) == arm64 {
```
0xFelix, line 338: Also test with multiple templates from different arches?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -352,39 +320,132 @@ var _ = Describe("Common-Templates operand", func() {
...
+				ExpectResourceNotExists(&template, request)
+			}
+		})
+	})
 })
 
 func getTestTemplates() []templatev1.Template {
-	return []templatev1.Template{{
+	const amdArch = "amd64"
+	const armArch = "arm64"
```
0xFelix, line 427: Make these top level consts in the file?

internal/operands/data-sources/reconcile.go
```diff
@@ -62,7 +62,12 @@ type dataSources struct {
 
 var _ operands.Operand = &dataSources{}
 
-func New(sources []cdiv1beta1.DataSource) operands.Operand {
+func New(sourceNames []string) operands.Operand {
```
0xFelix, line 65: `sourceNames` does not contain any duplicates?

internal/template-bundle/bundle.go
```diff
@@ -103,18 +61,32 @@ func extractDataSources(templates []templatev1.Template) ([]cdiv1beta1.DataSourc
...
+
+		templateArch := common_templates.GetTemplateArch(template)
+		dataSourceArchs[name][templateArch] = struct{}{}
+	}
+
+	result := make(map[string][]string, len(dataSourceArchs))
+	for name, archs := range dataSourceArchs {
+		result[name] = make([]string, 0, len(archs))
+		for arch := range archs {
+			result[name] = append(result[name], arch)
```
0xFelix, line 85: Why go the extra route by first collecting arches into a map and then creating a slice from it? You could collect them into a slice and sort/compact it after. This is very hard to understand IMO.

internal/operands/common-templates/reconcile.go
akrejcir: I think there are no duplicates in the templates.
The set that is created in `Reconcile` is used to check if a template exists, not to avoid duplicates.
I will try to make it easier to understand.

internal/operands/common-templates/reconcile.go
akrejcir: `.Spec.Cluster` is not a slice, its type is `*Cluster`.

internal/operands/common-templates/reconcile.go
akrejcir: `EnableMultipleArchitectures` is `*bool`, not a slice.

internal/operands/common-templates/reconcile_test.go
akrejcir: done

internal/operands/common-templates/reconcile_test.go
akrejcir: done

internal/template-bundle/bundle.go
akrejcir: done

internal/operands/common-templates/reconcile.go
akrejcir: This set is used to quickly filter templates based on architectures, so that `Reconcile` can only create compatible templates.

internal/operands/common-templates/reconcile.go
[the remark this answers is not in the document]
jcanocan: All right, thanks for the clarification.

internal/operands/common-templates/reconcile.go
akrejcir: I see. Those are checked in webhook. But I can add the check here.

akrejcir: Sorry, for confusion.
I see now that if the arch slice is empty, then the reconcile will fail, because of this block of code:
```
	clusterArchs := getClusterArchitectures(request.Instance)
	if len(clusterArchs) == 0 {
		return []common.ReconcileResult{}, fmt.Errorf("no cluster architectures defined")
	}
```

Is it enough?

akrejcir: I will fix the nil ptr.

internal/controllers/setup.go
```diff
@@ -63,16 +67,20 @@ func CreateControllers(ctx context.Context, apiReader client.Reader) ([]Controll
 		return nil, fmt.Errorf("failed to read vm-console-proxy bundle: %w", err)
 	}
 
+	// Temporarily, only create DataSources for one architecture.
+	// TODO: Create all DataSources, when multi-arch DataSource logic is implemented
+	dataSourceNames := slices.Collect(dataSourceCollection.Names())
+
 	sspOperands := []operands.Operand{
-		data_sources.New(templatesBundle.DataSources),
+		data_sources.New(dataSourceNames),
```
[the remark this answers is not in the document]
akrejcir: I'd like to keep it separate, so the comment is visible. It will be changed in a future PR.

api/v1beta3/ssp_types.go
```diff
@@ -47,11 +47,26 @@ type CommonTemplates struct {
...
+
 	// TemplateValidator is configuration of the template validator operand
 	TemplateValidator *TemplateValidator `json:"templateValidator,omitempty"`
 
+	// Cluster specifies what node architectures are present in the cluster.
+	Cluster *Cluster `json:"cluster,omitempty"`
```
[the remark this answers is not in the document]
jcanocan: All right!

internal/controllers/setup.go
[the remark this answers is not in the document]
jcanocan: Perfectly

internal/operands/common-templates/reconcile_test.go
0xFelix: `for i, template := range testTemplates`

0xFelix (commented):
@nunnatsa As the author of the VEP, can you give this a rough once over? I.e. checking the name of API fields, labels/annotations and the behavior that the tests assert?

internal/operands/common-templates/reconcile.go
```diff
@@ -68,7 +76,17 @@ func (c *commonTemplates) WatchTypes() []operands.WatchType {
 }
 
 func (c *commonTemplates) Reconcile(request *common.Request) ([]common.ReconcileResult, error) {
-	reconcileTemplatesResults, err := common.CollectResourceStatus(request, reconcileTemplatesFuncs(c.templatesBundle)...)
+	clusterArchs := getClusterArchitectures(request.Instance)
+	if len(clusterArchs) == 0 {
+		return []common.ReconcileResult{}, fmt.Errorf("no cluster architectures defined")
```
nunnatsa, line 81: I would return nil here.

internal/operands/data-sources/resource.go
```diff
@@ -5,13 +5,32 @@ import (
...
 
+func newDataSource(name string) cdiv1beta1.DataSource {
+	return cdiv1beta1.DataSource{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: internal.GoldenImagesNamespace,
+		},
+		Spec: cdiv1beta1.DataSourceSpec{
+			Source: cdiv1beta1.DataSourceSource{
+				PVC: &cdiv1beta1.DataVolumeSourcePVC{
```
nunnatsa, line 25: Is the only option is pvc? Shouldn't we prefer VolumeSnapsho? @arnongilboa

tests/webhook_test.go
```diff
@@ -183,6 +184,21 @@ var _ = Describe("Validation webhook", func() {
...
+				err := apiClient.Create(ctx, newSspV1Beta3, client.DryRunAll)
+				Expect(err).To(MatchError(ContainSubstring(".spec.cluster needs to be non-nil, if multi-architecture is enabled")))
+			})
+
+			It("should fail if control plane architectures are empty", func() {
+				newSspV1Beta3.Spec.Cluster = &sspv1beta3.Cluster{
+					ControlPlaneArchitectures: nil,
+				}
+				err := apiClient.Create(ctx, newSspV1Beta3, client.DryRunAll)
+				Expect(err).To(MatchError(ContainSubstring(".spec.cluster.controlPlaneArchitectures cannot be empty")))
```
nunnatsa, line 200: Not sure about this. We have clusters without control plane nodes, e.g. hypershift clusters. Anyway, it is possible anf we can't reject ssp changes for that.

nunnatsa (commented):
Impressive work, @akrejcir 🚀 

Added few inline comments.

Two general comments:
1. IIUC, the default architecture is taken from the ssp pod's node architecture. IIRC, this is not what was agreed, but this: if there id only one workloads architecture, then this is the default. Ig not, then the cp arch is the default.
2. The current implementation rejects ssp creation/updates if the cp arch list is empty. This is wrong, because we know clusters w/o cp nodes, e.g. hypershift clusters. HCO will always populate the cluster object, even if multi arch fg is off.

internal/operands/common-templates/reconcile.go
```diff
@@ -90,6 +108,27 @@ func (c *commonTemplates) Reconcile(request *common.Request) ([]common.Reconcile
...
+
+	if ptr.Deref(sspObj.Spec.EnableMultipleArchitectures, false) {
+		return sspObj.Spec.Cluster.WorkloadArchitectures
+	}
+
+	if len(sspObj.Spec.Cluster.ControlPlaneArchitectures) == 0 {
+		return nil
+	}
+
+	return []string{sspObj.Spec.Cluster.ControlPlaneArchitectures[0]}
```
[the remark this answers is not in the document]
akrejcir: I thought that when multi-architecture is disabled, we want to use the single architecture of the control plane. I will read the VEP again, maybe I misunderstood.

akrejcir:
> 1. IIUC, the default architecture is taken from the ssp pod's node architecture. IIRC, this is not what was agreed, but this: if there is only one workloads architecture, then this is the default. Ig not, then the cp arch is the default.

I kept the current behavior for backward compatibility, if there are no architectures defined in spec.
I will change it what you've wrote, use workload arch and if it is missing, use control plane arch.

> 2. The current implementation rejects ssp creation/updates if the cp arch list is empty. This is wrong, because we know clusters w/o cp nodes, e.g. hypershift clusters. HCO will always populate the cluster object, even if multi arch fg is off.

I see, I didn't know that. I will change the code.

internal/operands/common-templates/reconcile.go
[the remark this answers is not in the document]
akrejcir: Before this PR, we use `runtime.GOARCH` to choose which VM templates to create.
In this PR, when FG is off, I thought it would be better to use the `.spec.Cluster` if it is not nil, instead of `runtime.GOARCH`.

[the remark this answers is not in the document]
nunnatsa: Is it the same default architecture used for the architecture agnostic DataSorce, that points to the default one? If not, then I guess you should keep doing what you're doing now. If yes, we'll need to think about it.

api/v1beta3/ssp_types.go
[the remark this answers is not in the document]
akrejcir: I also prefer the first option.
@jcanocan @0xFelix, what do you think?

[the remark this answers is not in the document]
jcanocan: First option looks good to me in such case

tests/webhook_test.go
akrejcir: done

internal/operands/common-templates/reconcile.go
akrejcir: done

internal/operands/common-templates/reconcile.go
akrejcir: done

internal/operands/common-templates/reconcile.go
```diff
@@ -167,58 +227,68 @@ func getOldTemplatesLabelSelector() labels.Selector {
...
+		if err != nil {
+			// On error, we assume that the template has unknown architecture,
+			// and it will be deleted.
+			templateArch = ""
+		}
+
+		if !slices.Contains(archs, templateArch) {
+			funcs = append(funcs, reconcileDeleteTemplate(&template))
+			continue
+		}
```
0xFelix, line 280: Why are these checks necessary? Templates will be deleted regardless once L291 is reached?

internal/operands/common-templates/reconcile_test.go
```diff
@@ -93,12 +92,23 @@ var _ = Describe("Common-Templates operand", func() {
 			Logger:          log,
 			VersionCache:    common.VersionCache{},
 		}
+
+		filteredTemplates = make([]templatev1.Template, 0, len(testTemplates))
+		for _, template := range testTemplates {
+			templateArch, err := GetTemplateArch(&template)
+			Expect(err).ToNot(HaveOccurred())
+
+			if templateArch == architecture.AMD64 {
```
0xFelix, line 101: Why filter only for AMD64?

0xFelix: Maybe group existing tests a little better to make it obvious they are for the single arch template use case? And better separate test vars/helpers for both cases?

internal/template-bundle/bundle_test.go
```diff
@@ -1,123 +1,147 @@
...
+			}, {
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "windows10-desktop-medium",
+				},
+				Objects: []k8sruntime.RawExtension{{
+					Object: testVmTemplate,
+				}},
+				Parameters: []templatev1.Parameter{{
+					Name:  "DATA_SOURCE_NAME",
+					Value: "win10",
```
0xFelix, line 123: Use the const here

internal/template-bundle/bundle_test.go
```diff
@@ -1,123 +1,147 @@
...
+			testTemplates := []templatev1.Template{{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "centos-stream8-server-medium",
+				},
+				Objects: []k8sruntime.RawExtension{{
+					Object: testVmTemplate,
+				}},
+				Parameters: []templatev1.Parameter{{
+					Name:  "DATA_SOURCE_NAME",
+					Value: "centos-stream8",
```
0xFelix, line 109: Same

tests/commonTemplates_test.go
```diff
@@ -254,37 +249,17 @@ var _ = Describe("Common templates", func() {
...
+
 			Expect(apiClient.Create(ctx, oldTemplate)).ToNot(HaveOccurred(), "creation of dummy old template failed")
 		})
 
 		AfterEach(func() {
 			Expect(apiClient.Delete(ctx, oldTemplate)).ToNot(HaveOccurred(), "deletion of dummy old template failed")
-			Expect(apiClient.Delete(ctx, ownerTemplate)).ToNot(HaveOccurred(), "deletion of dummy owner template failed")
 		})
 
-		It("[test_id:5621]should replace ownerReference with owner annotations for older templates", func() {
```
0xFelix, line 273: Not required anymore?

tests/commonTemplates_test.go
```diff
@@ -355,6 +330,131 @@ var _ = Describe("Common templates", func() {
...
+		})
+
+		AfterAll(func() {
+			updateSsp(func(foundSsp *ssp.SSP) {
+				foundSsp.Spec.EnableMultipleArchitectures = ptr.To(true)
+			})
+			waitUntilDeployed()
+		})
+
+		DescribeTable("should delete template from non-default arch", func(res *testResource, arch architecture.Arch) {
```
0xFelix, line 442: The expected outcome for the test cases is unclear. What should happen on which arch?

tests/commonTemplates_test.go
akrejcir: No. This test was for backward compatibility with a very old version of ssp-operator, before it was written in golang. Now none of the templates can have `ownerReferences` pointing to SSP, because they are in differetnt namespace than SSP.

internal/operands/common-templates/reconcile.go
akrejcir: This function handles two cases for existing templates, that are owned by SSP, but are not currently being deployed (they are not in `deployedTemplates` parameter).

- If the existing template has architecture that is no longer supported by the cluster, the template is removed.
- If the existing template has architecture that is supported, the template is deprecated, by adding annotation and label.

internal/operands/common-templates/architecture_test.go
```diff
@@ -0,0 +1,35 @@
...
+	. "github.com/onsi/gomega"
+
+	templatev1 "github.com/openshift/api/template/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"kubevirt.io/ssp-operator/internal/architecture"
+)
+
+var _ = Describe("GetTemplateArch()", func() {
+	It("should read architecture from label", func() {
```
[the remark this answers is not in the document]
akrejcir: done

internal/template-bundle/bundle_test.go
akrejcir: done

tests/commonTemplates_test.go
0xFelix: ACK

internal/operands/common-templates/architecture_test.go
```diff
@@ -0,0 +1,35 @@
...
+		const testArch = architecture.ARM64
+		template := &templatev1.Template{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test",
+				Labels: map[string]string{
+					TemplateArchitectureLabel: string(testArch),
+				},
+			},
+		}
+		Expect(GetTemplateArch(template)).To(Equal(testArch))
```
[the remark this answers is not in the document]
0xFelix: ACK but also easy to miss

0xFelix (commented):
One more question

internal/operands/common-templates/reconcile.go
akrejcir: To keep the same backward compatible behavior. So far, before this PR, when a common template was removed from SSP, for example an old RHEL, we would keep the template in the cluster, because they may still be using it.

But now with multi-arch, when all nodes of a specific architecture are removed from the cluster, it does not make sense to keep the old template, because it cannot be run.

0xFelix (commented):
/approve

PR #1448, opened 2025-07-17
fix: clean up goroutines on error

### LINKED ISSUE #667: Add goroutine completion on error
/kind enhancement

Address TODO comments left in this PR - https://github.com/kubevirt/ssp-operator/pull/640#discussion_r1307192521

Make sure that goroutines finish when the function returns an error

Ruclo:
Followup PR to #640. Avoids a goroutine leak in the event of server.ListenAndServeTLS() returning an error.
Fixes #667 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,9 +22,11 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"golang.org/x/sync/errgroup"
 	"net/http"
 	"os"
 	"path"
+	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -120,6 +122,38 @@ type prometheusServer struct {
 	serverAddress string
 }
 
+func startCertWatcher(ctx context.Context, group *errgroup.Group, certWatcher *certwatcher.CertWatcher) {
+	group.Go(func() error {
+		if err := certWatcher.Start(ctx); err != nil {
+			setupLog.Error(err, "certificate watcher error")
+			return err
+		}
+		return nil
+	})
+}
+
+func startServer(ctx context.Context, group *errgroup.Group, server *http.Server) {
+	group.Go(func() error {
+		if err := server.ListenAndServeTLS("", ""); !errors.Is(err, http.ErrServerClosed) {
+			setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
+			return err
+		}
+		return nil
+	})
+
+	group.Go(func() error {
+		<-ctx.Done()
+		setupLog.Info("shutting down Prometheus metrics server")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+		if err := server.Shutdown(shutdownCtx); err != nil {
+			setupLog.Error(err, "error shutting down the HTTP server")
+			return err
+		}
+		return nil
+	})
+}
+
 // NeedLeaderElection implements the LeaderElectionRunnable interface, which indicates
 // the prometheus server doesn't need leader election.
 func (s *prometheusServer) NeedLeaderElection() bool {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -142,19 +142,17 @@ func (s *prometheusServer) Start(ctx context.Context) error {
...
 
+	subCtx, cancel := context.WithCancel(ctx)
 	go func() {
-		// TODO: change context, so it can be closed when
-		// this function returns an error
-		if err := certWatcher.Start(ctx); err != nil {
+		if err := certWatcher.Start(subCtx); err != nil {
 			setupLog.Error(err, "certificate watcher error")
 		}
+		cancel()
```
akrejcir, line 150: Use `defer cancel()` at the start of this goroutine. That is the usual way to call `cancel()`.

main.go
```diff
@@ -142,19 +142,17 @@ func (s *prometheusServer) Start(ctx context.Context) error {
...
 
 	idleConnsClosed := make(chan struct{})
 	go func() {
-		// TODO: make sure that the goroutine finishes when
-		// this function returns an error
-		<-ctx.Done()
+		<-subCtx.Done()
 		setupLog.Info("shutting down Prometheus metrics server")
 
 		if err := server.Shutdown(context.Background()); err != nil {
```
akrejcir, line 158: Please, add a check here, so that error is not logged if the server is already shutdown.
```
!errors.Is(err, http.ErrServerClosed)
```

main.go
```diff
@@ -165,13 +163,16 @@ func (s *prometheusServer) Start(ctx context.Context) error {
...
 	server.TLSConfig = s.getPrometheusTLSConfig(ctx, certWatcher)
 
-	if err := server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil && !errors.Is(err, http.ErrServerClosed) {
+	if err = server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil && !errors.Is(err, http.ErrServerClosed) {
 		setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
-		return err
+		cancel()
 	}
 
 	<-idleConnsClosed
```
akrejcir, line 171: It would be better to use `sync.WaitGroup` here to wait for both of the goroutines to finish before the function returns.

main.go
```diff
@@ -165,13 +163,16 @@ func (s *prometheusServer) Start(ctx context.Context) error {
...
-	return nil
+	if errors.Is(err, http.ErrServerClosed) {
```
akrejcir, line 172: It is usual that the last return in a function is the successful path `return nil`. Can you invert this `if` so that it returns the error?

main.go
```diff
@@ -165,13 +163,16 @@ func (s *prometheusServer) Start(ctx context.Context) error {
 
 	server.TLSConfig = s.getPrometheusTLSConfig(ctx, certWatcher)
 
-	if err := server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil && !errors.Is(err, http.ErrServerClosed) {
+	if err = server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil && !errors.Is(err, http.ErrServerClosed) {
 		setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
-		return err
+		cancel()
```
akrejcir, line 168: I think the `cancel()` should be called in any case, so it can be moved outside of the `if` block.

main.go
```diff
@@ -120,6 +122,38 @@ type prometheusServer struct {
...
+			setupLog.Error(err, "certificate watcher error")
+			return err
+		}
+		return nil
+	})
+}
+
+func startServer(ctx context.Context, group *errgroup.Group, server *http.Server) {
+	group.Go(func() error {
+		if err := server.ListenAndServeTLS("", ""); !errors.Is(err, http.ErrServerClosed) {
```
ksimon1, line 137: The previous code used `server.ListenAndServeTLS(s.certPath, s.keyPath)`. You are using empty string. Is it correct?

main.go
```diff
@@ -25,6 +25,9 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"time"
+
+	"golang.org/x/sync/errgroup"
```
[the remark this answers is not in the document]
ksimon1: This is still not updated

main.go
```diff
@@ -132,46 +166,21 @@ func (s *prometheusServer) Start(ctx context.Context) error {
...
-
-	server.TLSConfig = s.getPrometheusTLSConfig(ctx, certWatcher)
-
-	if err := server.ListenAndServeTLS(s.certPath, s.keyPath); err != nil && !errors.Is(err, http.ErrServerClosed) {
-		setupLog.Error(err, "Failed to start Prometheus metrics endpoint server")
-		return err
+	server := &http.Server{
+		Addr:      s.serverAddress,
+		Handler:   mux,
+		TLSConfig: s.getPrometheusTLSConfig(ctx, certWatcher),
```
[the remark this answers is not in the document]
Ruclo: The initial read of the certificate already happens in the New() function which initializes the certWatcher.

main.go
Ruclo: The documentation for the server.ListenAndServeTLS() function states
> Filenames containing a certificate and matching private key for the server must be provided if neither the [Server]'s TLSConfig.Certificates nor TLSConfig.GetCertificate are populated.

The GetCertificate should be populated by the GetConfigForClient callback which gets called during the tls handshake. After deploying this to my cluster,  the ssp operator's metrics service uses the TLS certificate.

0xFelix (commented):
/approve

PR #1450, opened 2025-07-17
api: Add new fields for multi-arch support

akrejcir:
Added API fields:
- `.spec.enableMultipleArchitectures` that enables multi-arch features.
- `.spec.cluster` that specifies architectures in cluster.

There is no implementation in this PR. The API is separated into its own PR, so we can create an alpha release.

Jira: https://issues.redhat.com/browse/CNV-61573

**Release note**:
```release-note
Added API fields ".spec.enableMultipleArchitectures" and ".spec.cluster".
```


jcanocan:
Thanks!
Nothing to comment! Great job!
/lgtm

PR #1464, opened 2025-08-07
DNM: Tighten network policies

0xFelix:

This PR tries to tighten network policies, so only the Kubernetes API can access services exposed by SSP and its operands. Only metrics are left open to be accessed by any client.

Fixes #

**Release note**:

```release-note
None
```


jcanocan:
Excellent contribution. My only comment is that you can reduce code duplication by extracting the common parts of `[]networkv1.NetworkPolicyPeer` in file `networkpolicies.go` as suggested by Sonar. Rather than that, it looks great.

PR #1465, opened 2025-08-07
chore: Bump golangci-lint version

0xFelix:

Bump golangci-lint version to v2.3.1. Run golangci-lint migrate to migrate existing configuration to the new v2 format and re-add dot-import-whitelist for ginkgo and gomega. This also fixes all new findings that occured after the upgrade.

Also avoids CI failues because of golangci-lint going haywire.

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,14 +1,35 @@
+version: "2"
 linters:
-  # Disable all linters.
-  disable-all: true
-  # Enable specific linter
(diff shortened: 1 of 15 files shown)


ksimon1:
Nit: It would be great to split the commit to two - one with changes in golangci.yaml and second one with changes in the code. Feel free to unhold if you don't want to do it :).
/hold
/approve

PR #1469, opened 2025-08-11
fix(networkpolicies): Apply multiple fixes to network policies

0xFelix:

Using named ports can lead to issues depending on the CNI implemenation in use. Specify the port number for vm-console-proxy's API instead.

Do not add selectors for host networking Pods (e.g. the kubeapi) as this
is very dependent on the CNI in use. E.g. for OVN this doesn't work at
the moment.

Fixes #

**Release note**:

```release-note
None
```


0xFelix:
@jcanocan @akrejcir Please take another look, had to add one more commit.

PR #1472, opened 2025-08-11
tests: common-templates: fix arch for old templates

akrejcir:
If the old template does not have architecture label, it would be removed when running test suite on non-amd architecture.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

PR #1503, opened 2025-08-26
tests: Wrap all API update calls in Eventually()

akrejcir:
Sometimes the Update calls fail, because there is a concurrent update to the object. This causes the test to fail and makes the test lane flaky.

This change makes the tests more stable.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

Btw. Patching instead of Updating could have been an alternative

PR #1509, opened 2025-08-27
[release-v0.21] tests: Wrap all API update calls in Eventually()

akrejcir:
This is a manual cherry-pick of https://github.com/kubevirt/ssp-operator/pull/1503 .
The import line was missing in the automatic cherry-pick.

Sometimes the Update calls fail, because there is a concurrent update to the object. This causes the test to fail and makes the test lane flaky.

This change makes the tests more stable.

**Release note**:
```release-note
None
```


codingben (approved):
/lgtm

PR #1511, opened 2025-08-27
enhancement: handle SIGTERM and SIGINT in template-validator

Ruclo:
This ensures that template validator gracefully shuts down after receiving SIGTERM or SIGINT.

@0xFelix
https://issues.redhat.com/browse/CNV-67397

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/template-validator/validator/app.go
+++ b/internal/template-validator/validator/app.go
@@ -3,8 +3,11 @@ package validator
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"net/http"
 	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	templatev1 "github.com/openshift/api/template/v1"
@@ -15,7 +18,6 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	kubevirtv1 "kubevirt.io/api/core/v1"
-
 	"kubevirt.io/ssp-operator/internal/template-validator/logger"
 	"kubevirt.io/ssp-operator/internal/template-validator/service"
 	"kubevirt.io/ssp-operator/internal/template-validator/tlsinfo"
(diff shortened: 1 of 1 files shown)


internal/template-validator/validator/app.go
```diff
@@ -99,7 +102,9 @@ func (app *App) Run() {
 		webhookServer.TLSConfig = createTLSConfig(tlsInfo)
 	}
 
-	g, ctx := errgroup.WithContext(context.Background())
+	signalsCtx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
```
0xFelix, line 105: Shouldn't it be using `os.*` signals instead of `syscall` ones?

internal/template-validator/validator/app.go
```diff
@@ -159,13 +164,13 @@ func (app *App) createWebhookServer(informers *virtinformers.Informers) *http.Se
 func startServer(server *http.Server, serverType string) error {
 	if server.TLSConfig != nil {
 		logger.Log.Info("TLS configured, serving "+serverType+" over HTTPS", "address", server.Addr)
-		if err := server.ListenAndServeTLS("", ""); err != nil && err != http.ErrServerClosed {
+		if err := server.ListenAndServeTLS("", ""); !errors.Is(err, http.ErrServerClosed) {
```
0xFelix, line 167: What if the error was `nil`?

internal/template-validator/validator/app.go
Ruclo: there are only SIGINT and SIGKILL defined in os. there is also this Signal interface which has this comment

> // A Signal represents an operating system signal.
// The usual underlying implementation is operating system-dependent:
// on Unix it is syscall.Signal.

internal/template-validator/validator/app.go
Ruclo: ListenAndServeTLS() always returns a non-nil error, Shutdown() and Close() cause ErrServerClosed to be returned from ListenAndServeTLS

internal/template-validator/validator/app.go
```diff
@@ -15,6 +16,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	kubevirtv1 "kubevirt.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
```
0xFelix, line 19: Template validator is not using controller-runtime otherwise. I'd prefer to stick to the solution you had before. :)

0xFelix (commented):
/approve
/ok-to-test

PR #1512, opened 2025-08-27
feat: Create PodDisruptionBudget for template validator

akrejcir:
At least one template validator pod should be available. 
This PR adds PodDisruptionBudget object that specifies this.

Fixes: https://issues.redhat.com/browse/CNV-67393

**Release note**:
```release-note
Create PodDisruptionBudget for template validator.
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -168,6 +168,17 @@ rules:
   - update
   - watch
 - apiGroups:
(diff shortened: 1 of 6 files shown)


jcanocan:
What about adding a new func test that ensures if the pods are deleted, they will be recreated? Otherwise, looks great!

PR #1514, opened 2025-08-27
chore: Set port names according to convention

akrejcir:
The convention is, that port name should be prefixed with protocol.

https://redhat-best-practices-for-k8s.github.io/guide/#k8s-best-practices-requirements-cnf-reqs

The new names are different, because the maximum port name length is 15 characters.

Jira: https://issues.redhat.com/browse/CNV-67392

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

PR #1529, opened 2025-09-02
feat: Add probes to operator and template validator

akrejcir:
Added:
- liveliness probe to template validator
- startup probe to ssp operator and template validator

The startup probes are the same as liveliness probes. We don't have specific logic for them.

Jira: https://issues.redhat.com/browse/CNV-67398

**Release note**:
```release-note
None
```

### DIFF
--- a/config/manager/manager.template.yaml
+++ b/config/manager/manager.template.yaml
@@ -58,10 +58,11 @@ spec:
             - "ALL"
         livenessProbe:
           httpGet:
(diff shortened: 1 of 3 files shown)


akrejcir:
> Is there really a benefit when using the same endpoints?

Yes, we don't need to implement other endpoints. The code would be the same for all of them.
This PR is to comply with the best practices.

PR #1542, opened 2025-09-08
chore: Increase linter timeout

akrejcir:
Sometimes the CI is flaky, because 5 minutes was not enough. Restarting the test helps, so it probably depends on the CI worker load.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

PR #1550, opened 2025-09-11
Trigger reconciliation at Deletiontimestamp change

Ruclo:
This PR makes sure that the ssp CR state is set to Deploying while deleting an owned resource.
For example, deleting the kubevirt-os-images namespace would not cause the change of state, because deleting this namespace required deleting many other resources, which took a lot of time.
I also extracted the Finalizer Changed Predicate.

Fixes # https://issues.redhat.com/browse/CNV-68906

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/misc_test.go
+++ b/tests/misc_test.go
@@ -8,6 +8,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -127,28 +128,28 @@ var _ = Describe("SSPOperatorReconcileSucceeded metric", func() {
 	})
 })
 
-func removeFinalizer(deploymentRes testResource, finalizerName string) {
+func removeFinalizer(res testResource, finalizerName string) {
 	Eventually(func() error {
-		deployment := &apps.Deployment{}
-		err := apiClient.Get(ctx, deploymentRes.GetKey(), deployment)
+		obj := res.NewResource()
+		err := apiClient.Get(ctx, res.GetKey(), obj)
 		if err != nil {
 			return err
 		}
 		// remove the finalizer so everything can go back to normal
-		controllerutil.RemoveFinalizer(deployment, finalizerName)
-		return apiClient.Update(ctx, deployment)
+		controllerutil.RemoveFinalizer(obj, finalizerName)
+		return apiClient.Update(ctx, obj)
 	}, env.ShortTimeout(), time.Second).ShouldNot(HaveOccurred())
 }
 
-func addFinalizer(deploymentRes testResource, finalizerName string) {
+func addFinalizer(res testResource, finalizerName string) {
 	Eventually(func() error {
-		deployment := &apps.Deployment{}
-		err := apiClient.Get(ctx, deploymentRes.GetKey(), deployment)
+		obj := res.NewResource()
+		err := apiClient.Get(ctx, res.GetKey(), obj)
 		if err != nil {
 			return err
 		}
-		controllerutil.AddFinalizer(deployment, finalizerName)
-		return apiClient.Update(ctx, deployment)
+		controllerutil.AddFinalizer(obj, finalizerName)
+		return apiClient.Update(ctx, obj)
 	}, env.ShortTimeout(), time.Second).ShouldNot(HaveOccurred())
 }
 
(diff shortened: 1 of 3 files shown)


tests/misc_test.go
```diff
@@ -371,6 +372,48 @@ var _ = Describe("RHEL VM creation", func() {
 	)
 })
 
+var _ = Describe("SSP Status Phase", func() {
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	It("should set phase to Deploying when owned resource gets deletionTimestamp", func() {
```
akrejcir, line 380: This test may work accidentally, because it is removing a namespace and it can take some time before it is removed. So the test has time to check the deletion time stamp.

Instead, I think it would be better to intentionally trigger the case that we want to test.
Can you change it, so that the test code adds a new finalizer to some object, and then deletes the object. This way we can be sure that the object will not be deleted before the test removes the finalzier.

It would also be better to use some small object for the test, like a ConfigMap, or a RoleBinding that is not used by SSP.

tests/misc_test.go
```diff
@@ -371,6 +372,48 @@ var _ = Describe("RHEL VM creation", func() {
...
+		watch, err := StartWatch(sspListerWatcher)
+		Expect(err).ToNot(HaveOccurred())
+		defer watch.Stop()
+
+		// Get the golden images namespace
+		goldenImagesNS := &core.Namespace{}
+		err = apiClient.Get(ctx, client.ObjectKey{Name: internal.GoldenImagesNamespace}, goldenImagesNS)
+		Expect(err).ToNot(HaveOccurred(), "Golden images namespace should exist")
```
akrejcir, line 388: It is more readable to use:
```go
Expect(apiClient.Get(...)).To(Succeed())
```

tests/misc_test.go
Ruclo: but the resource getting deleted needs to be watched by ssp ?

tests/misc_test.go
```diff
@@ -371,6 +371,52 @@ var _ = Describe("RHEL VM creation", func() {
 	)
 })
 
+var _ = Describe("SSP Status Phase", func() {
+	BeforeEach(func() {
+		waitUntilDeployed()
+	})
+
+	It("should set phase to Deploying when owned resource gets deletionTimestamp", func() {
+		watch, err := StartWatch(sspListerWatcher)
```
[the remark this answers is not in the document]
akrejcir: Watch is more difficult to understand, because you need to pay attention when it is started.

tests/misc_test.go
```diff
@@ -371,6 +373,52 @@ var _ = Describe("RHEL VM creation", func() {
...
+			ssp := getSsp()
+
+			available := conditionsv1.FindStatusCondition(ssp.Status.Conditions, conditionsv1.ConditionAvailable)
+			progressing := conditionsv1.FindStatusCondition(ssp.Status.Conditions, conditionsv1.ConditionProgressing)
+			degraded := conditionsv1.FindStatusCondition(ssp.Status.Conditions, conditionsv1.ConditionDegraded)
+
+			return (available.Status == core.ConditionFalse && strings.Contains(available.Message, message)) &&
+				(progressing.Status == core.ConditionTrue && strings.Contains(progressing.Message, message)) &&
+				(degraded.Status == core.ConditionTrue && strings.Contains(degraded.Message, message))
+		}, env.Timeout(), time.Second).Should(BeTrue())
```
akrejcir, line 414: If this fails, the error message will be only something like `expected True and got False`, that is not very helpful to see what went wrong.
Instead, can you change this to use nested `Gomega` instance like this?
```go
Eventually(func(g Gomega) {
    // ..code...
	g.Expect(available.Status).To(Equal(metav1.ConditionFalse))
	g.Expect(available.Message).To(ContainSubstring(message))
	// ...
}, env.ShortTimeout(), time.Second).Should(Succeed())
```

When using nested gomega like this, a failure will not abort the Eventually() loop, instead it will continue in next iteration. And when the timeout is reached, we will get sensible error message.

Also most of the tests use `env.ShortTimeout()`. The long timeout is only used for things that can take longer than a minute. For example importing container images, or deploying all SSP objects.

tests/misc_test.go
```diff
@@ -371,6 +373,52 @@ var _ = Describe("RHEL VM creation", func() {
...
+		}()
+
+		addFinalizer(configMapRes, finalizerName)
+
+		Expect(apiClient.Delete(ctx, &core.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configMapRes.Name,
+				Namespace: configMapRes.Namespace,
+			},
+		})).ToNot(HaveOccurred())
```
akrejcir, line 401: Please use `.To(Succeed())`.

tests/misc_test.go
```diff
@@ -371,6 +373,52 @@ var _ = Describe("RHEL VM creation", func() {
...
+			degraded := conditionsv1.FindStatusCondition(ssp.Status.Conditions, conditionsv1.ConditionDegraded)
+
+			return (available.Status == core.ConditionFalse && strings.Contains(available.Message, message)) &&
+				(progressing.Status == core.ConditionTrue && strings.Contains(progressing.Message, message)) &&
+				(degraded.Status == core.ConditionTrue && strings.Contains(degraded.Message, message))
+		}, env.Timeout(), time.Second).Should(BeTrue())
+
+		removeFinalizer(configMapRes, finalizerName)
+
+		Eventually(func() bool { return isStatusDeployed(getSsp()) }, env.Timeout(), time.Second).Should(BeTrue())
```
akrejcir, line 418: Please use short timeout.

tests/misc_test.go
```diff
@@ -371,6 +373,52 @@ var _ = Describe("RHEL VM creation", func() {
...
+			Name:      validator.ConfigMapName,
+			Namespace: strategy.GetNamespace(),
+			Resource:  &core.ConfigMap{},
+		}
+
+		defer func() {
+			removeFinalizer(configMapRes, finalizerName)
+		}()
+
+		addFinalizer(configMapRes, finalizerName)
```
akrejcir, line 394: The usual pattern is to swap these calls. A call to `defer` is done after the object is created. `addFinalizer` could fail and panic, and then there is no reason to call `removeFinalizer`.

PR #1556, opened 2025-09-22
fix(instancetypes): Protect instancetype/preference `controllerRevision` objects

jcanocan:

It refactors the VAP and VAPB VM delete protection function names to be more descriptive.

Moreover, it enables a protection for instancetype and preference `controllerRevision` objects, which ensures that they are not deleted before `VirtualMachine` objects. This is due to the fact that in some scenarios, such as when a VM is deleted protected, if the `controllerRevision` is deleted, but the VM object is still alive, users will not be able to patch or update the VM object. This leads to virtually deadlocking the VM object.

The VAP ensures that only the Kubernetes garbage collector and kubevirt-controller are able to delete these objects:

* Garbage-collector: It deletes objects when the `ownerReference`'d object is removed, e.g., the instancetype "test-it" with
  `ownerReference: test-vm` is removed when the VM test-vm is removed. Therefore, by only allowing it to drop these objects, we ensure the deletion order.
* Kubevirt-controller: It is required to no block this SA because it is the main owner of these `controllerRevision`, it needs to delete them in upgrade scenarios, live migration, etc.

Fixes # [CNV-69511](https://issues.redhat.com/browse/CNV-69511)

**Release note**:

```release-note
Add protection to preference and instancetype `controllerRevision` objects.
```

### DIFF
--- a/tests/vm_delete_protection_test.go
+++ b/tests/vm_delete_protection_test.go
@@ -7,8 +7,12 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/utils/ptr"
 	kubevirtv1 "kubevirt.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -16,7 +20,10 @@ import (
 	"kubevirt.io/ssp-operator/tests/env"
 )
 
-const deleteProtectionLabel = "kubevirt.io/vm-delete-protection"
+const (
+	deleteProtectionLabel = "kubevirt.io/vm-delete-protection"
+	vmReadyTimeout        = 5 * time.Minute
+)
 
 var _ = Describe("VM delete protection", func() {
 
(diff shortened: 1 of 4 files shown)


tests/vm_delete_protection_test.go
```diff
@@ -72,24 +79,192 @@ var _ = Describe("VM delete protection", func() {
...
+func createVMWithLabels(labels map[string]string, namespace string) *kubevirtv1.VirtualMachine {
 	vmName := fmt.Sprintf("testvmi-%v", rand.String(10))
-	vmi := NewMinimalVMIWithNS(strategy.GetNamespace(), vmName)
+	vmi := NewMinimalVMIWithNS(namespace, vmName)
+	vmi.Spec.Volumes = []kubevirtv1.Volume{
+		{
+			Name: "containerdisk",
+			VolumeSource: kubevirtv1.VolumeSource{
+				ContainerDisk: &kubevirtv1.ContainerDiskSource{
+					Image: "quay.io/containerdisks/debian:latest",
```
lyarwood, line 226: Why is this required?

jcanocan:
> Reflecting on this I'm not convinced by the approach as I'm worried it's too brittle and will break backup/restore tooling.

Could you explain how it will break backup/restore? Won't these operations be executed by kubevirt-controller?

> What about having SSP itself reconcile all VMs with the delete protection label and then applying or removing it from any additional resource we also need to protect? Possibly using the [objectgraph](https://kubevirt.io/api-reference/main/operations.html#_v1vm-objectgraph) subresource API?

Yes, we could apply the same concept. However, what other objects should we protect? AFAIK, we just have `controllerRevision` and VMs. One question, won't the `controllerRevision` be reconciled if we set the protection label?

> This would keep the VAPs simple, don't delete anything with the delete protection label while ensuring we programmatically include resources the VM depends on.

The requirements of this feature are just to protect VMs, in the past we explored what would happen if we overprotect objects like, launcher pods, VMIs, etc. and the result for instance, vms won't be able to stop, migrate, etc.

tests/vm_delete_protection_test.go
lyarwood: Yes, I was wondering if that's why you added this. The CR creation doesn't rely on the guest OS, just the VM control loop in `virt-controller`.

dominikholler:
@lyarwood isn't is a generic problem, independenly from the deletion protection, that a controller revision of a instancetype/preference can be deleted before the related VM is deleted? As long as there is a controller revision per VM, might there be a generic solution like having a finalizer on the controler revision, and during VM deletion the finalizer on the controller revision is removed?

PR #1564, opened 2025-09-29
ci: Include bundle in check of generated files

akrejcir:
`make bundle` now copies manifests to `data` directory. It is called in the GitHub action that checks uncommitted files.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

PR #1567, opened 2025-10-01
use tls for prometheus metrics in ssp operator and template validator

Ruclo:
Creates a separate service monitor targeting the template validator metrics and verifies the metrics service tls certificate when querying prometheus metrics.

Fixes CNV-55455

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/controllers/services_controller.go
+++ b/internal/controllers/services_controller.go
@@ -26,7 +26,6 @@ import (
 
 const (
 	ServiceManagedByLabelValue = "ssp-operator-services"
-	MetricsServiceName         = "ssp-operator-metrics"
 	OperatorName               = "ssp-operator"
 	ServiceControllerName      = "service-controller"
 )
@@ -38,13 +37,14 @@ func ServiceObject(namespace string, appKubernetesPartOfValue string) *v1.Servic
 		common.AppKubernetesVersionLabel:   env.GetOperatorVersion(),
 		common.AppKubernetesComponentLabel: ServiceControllerName,
 		metrics.PrometheusLabelKey:         metrics.PrometheusLabelValue,
+		metrics.MetricsServiceKey:          common.SspOperatorMetricsServiceName,
 	}
 	if appKubernetesPartOfValue != "" {
 		labels[common.AppKubernetesPartOfLabel] = appKubernetesPartOfValue
 	}
 	return &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      MetricsServiceName,
+			Name:      common.SspOperatorMetricsServiceName,
 			Namespace: namespace,
 			Labels:    labels,
 		},
@@ -136,7 +136,7 @@ func (s *serviceReconciler) setupController(mgr ctrl.Manager) error {
 		Named("service-controller").
 		For(&v1.Service{}, builder.WithPredicates(predicate.NewPredicateFuncs(
 			func(object client.Object) bool {
-				return object.GetName() == MetricsServiceName && object.GetNamespace() == s.operatorNamespace
+				return object.GetName() == common.SspOperatorMetricsServiceName && object.GetNamespace() == s.operatorNamespace
 			}))).
 		Complete(s)
 }
--- a/internal/controllers/ssp_controller.go
+++ b/internal/controllers/ssp_controller.go
@@ -61,24 +61,28 @@ const (
 
 // sspController reconciles a SSP object
 type sspController struct {
-	log              logr.Logger
-	operands         []operands.Operand
-	lastSspSpec      ssp.SSPSpec
-	subresourceCache common.VersionCache
-	topologyMode     osconfv1.TopologyMode
-	areCrdsMissing   bool
+	log                logr.Logger
+	operands           []operands.Operand
+	lastSspSpec        ssp.SSPSpec
+	subresourceCache   common.VersionCache
+	topologyMode       osconfv1.TopologyMode
+	areCrdsMissing     bool
+	olmDeployment      bool
+	sspServiceHostname string
 
 	client         client.Client
 	uncachedReader client.Reader
 	crdList        crd_watch.CrdList
 }
 
-func NewSspController(infrastructureTopology osconfv1.TopologyMode, operands []operands.Operand) Controller {
+func NewSspController(infrastructureTopology osconfv1.TopologyMode, operands []operands.Operand, olmDeployment bool, sspServiceHostname string) Controller {
 	return &sspController{
-		log:              ctrl.Log.WithName("controllers").WithName("SSP"),
-		operands:         operands,
-		subresourceCache: common.VersionCache{},
-		topologyMode:     infrastructureTopology,
+		log:                ctrl.Log.WithName("controllers").WithName("SSP"),
+		operands:           operands,
+		subresourceCache:   common.VersionCache{},
+		topologyMode:       infrastructureTopology,
+		olmDeployment:      olmDeployment,
+		sspServiceHostname: sspServiceHostname,
 	}
 }
 
@@ -175,16 +179,18 @@ func (s *sspController) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 	sspChanged := s.clearCacheIfNeeded(instance)
 
 	sspRequest := &common.Request{
-		Request:         req,
-		Client:          s.client,
-		UncachedReader:  s.uncachedReader,
-		Context:         ctx,
-		Instance:        instance,
-		InstanceChanged: sspChanged,
-		Logger:          reqLogger,
-		VersionCache:    s.subresourceCache,
-		TopologyMode:    s.topologyMode,
-		CrdList:         s.crdList,
+		Request:            req,
+		Client:             s.client,
+		UncachedReader:     s.uncachedReader,
+		Context:            ctx,
+		Instance:           instance,
+		InstanceChanged:    sspChanged,
+		Logger:             reqLogger,
+		VersionCache:       s.subresourceCache,
+		TopologyMode:       s.topologyMode,
+		CrdList:            s.crdList,
+		OLMDeployment:      s.olmDeployment,
+		SSPServiceHostname: s.sspServiceHostname,
 	}
 
 	if !isInitialized(sspRequest.Instance) {
--- a/internal/operands/metrics/resources.go
+++ b/internal/operands/metrics/resources.go
@@ -1,12 +1,15 @@
 package metrics
 
 import (
+	"fmt"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	v1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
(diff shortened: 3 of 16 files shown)


internal/operands/metrics/resources.go
```diff
@@ -61,31 +79,126 @@ func ServiceMonitorLabels() map[string]string {
...
+				ConfigMap: &v1.ConfigMapKeySelector{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: "openshift-service-ca.crt",
+					},
+					Key: "service-ca.crt",
+				},
+			},
+		},
+	}
+	tlsConfig.ServerName = ptr.To("virt-template-validator.kubevirt.svc")
```
0xFelix, line 148: Read that from the cert to be sure? The namespace can vary.

internal/operands/metrics/resources.go
```diff
@@ -61,31 +79,126 @@ func ServiceMonitorLabels() map[string]string {
...
+}
+
+// getCAConfigForServiceMonitor returns the appropriate CA configuration
+func getCAConfigForServiceMonitor() *promv1.SecretOrConfigMap {
+	if isOLMDeployment {
+		// OLM served certificate: use ssp-operator-service-cert secret with olmCAKey
+		return &promv1.SecretOrConfigMap{
+			Secret: &v1.SecretKeySelector{
+				LocalObjectReference: v1.LocalObjectReference{
+					Name: "ssp-operator-service-cert",
```
0xFelix, line 116: Names of objects that are not reconciled by SSP should be configurable.

internal/operands/metrics/resources.go
```diff
@@ -18,6 +21,14 @@ const (
 	PrometheusClusterRoleName    = "prometheus-k8s-ssp"
 	PrometheusServiceAccountName = "prometheus-k8s"
 	MetricsPortName              = "http-metrics"
+
+	TemplateValidatorMetricsServiceName = "template-validator-metrics"
```
akrejcir: Please use this constant also in `template-validator/resources.go` file, so that we are sure that it is the same string.

internal/operands/metrics/resources.go
```diff
@@ -61,31 +72,87 @@ func ServiceMonitorLabels() map[string]string {
...
+		ConfigMap: &v1.ConfigMapKeySelector{
+			LocalObjectReference: v1.LocalObjectReference{
+				Name: ServiceCABundle,
+			},
+			Key: ServiceCABUndleKey,
+		},
+	}
+}
+
+func olmManagedCABundle() promv1.SecretOrConfigMap {
```
akrejcir: Can you inline this function? It is only used in one place, and it is short.

internal/operands/metrics/resources.go
```diff
@@ -61,31 +72,87 @@ func ServiceMonitorLabels() map[string]string {
...
+	return promv1.SecretOrConfigMap{
+		Secret: &v1.SecretKeySelector{
```
akrejcir: Please add a comment, that OLM uses a self-signed CA, and that it is not available in a ConfigMap.
So that future developers know why we use a secret.

internal/operands/metrics/resources.go
```diff
@@ -61,31 +72,87 @@ func ServiceMonitorLabels() map[string]string {
...
+
+	serviceMonitor := newServiceMonitor(TemplateValidatorMetricsServiceName, request.Namespace, tlsConfig, metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			MetricsServiceKey: TemplateValidatorMetricsServiceName,
+		},
+	})
+	return &serviceMonitor
+}
+
+func newSspServiceMonitor(request common.Request) *promv1.ServiceMonitor {
```
akrejcir: Similarly here, can you only pass required fields to the function instead of whole `request` struct?

internal/operands/metrics/resources.go
```diff
@@ -61,31 +72,87 @@ func ServiceMonitorLabels() map[string]string {
...
+	tlsConfig := &promv1.TLSConfig{
+		SafeTLSConfig: promv1.SafeTLSConfig{
+			CA: getCAConfigForServiceMonitor(request.OLMDeployment),
+		},
+	}
+	tlsConfig.ServerName = ptr.To(request.SSPServiceHostname)
```
akrejcir: This can be included in the struct creation above.

main.go
```diff
@@ -210,24 +207,53 @@ func newPrometheusServer(metricsAddr string, cache cache.Cache) (*prometheusServ
...
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&olmDeployment, "olm-deployment", false, "Signalize deployment via OLM")
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	err := createCertificateSymlinks()
```
akrejcir: I see you've removed the symlinks. Are they not needed anymore?
If yes, can you do it in a separate commit, because it is not related to the metrics change.

internal/operands/metrics/resources.go
```diff
@@ -61,31 +72,87 @@ func ServiceMonitorLabels() map[string]string {
...
+	return serviceCABundle()
+}
+
+func newValidatorServiceMonitor(request common.Request) *promv1.ServiceMonitor {
+	tlsConfig := &promv1.TLSConfig{
+		SafeTLSConfig: promv1.SafeTLSConfig{
+			CA: serviceCABundle(),
+		},
+	}
+	tlsConfig.ServerName = ptr.To(fmt.Sprintf("virt-template-validator.%s.svc", request.Namespace))
```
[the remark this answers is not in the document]
Ruclo: What do you think about a common package in operands where these constants could be put or perhaps i could use the already present common package in internal?

internal/controllers/services_controller.go
```diff
@@ -143,8 +103,19 @@ func (s *serviceReconciler) setupController(mgr ctrl.Manager) error {
...
+		common.AppKubernetesVersionLabel:   env.GetOperatorVersion(),
+		common.AppKubernetesComponentLabel: ServiceControllerName,
+	}
 	appKubernetesPartOfValue := deployment.GetLabels()[common.AppKubernetesPartOfLabel]
-	service := ServiceObject(deployment.Namespace, appKubernetesPartOfValue)
+	if appKubernetesPartOfValue != "" {
+		labels[common.AppKubernetesPartOfLabel] = appKubernetesPartOfValue
+	}
+	common.UpdateStringMap(labels, service.GetLabels())
+
```
akrejcir: The labels can be simply added to the `service`, without creating a new map or calling `common.UpdateStringMap()`: 
```go
service := resources.SspMetricsService(deployment.Namespace)
service.Labels[common.AppKubernetesManagedByLabel] = ServiceManagedByLabelValue
service.Labels[common.AppKubernetesVersionLabel] = env.GetOperatorVersion()
...
```

internal/resources/metrics_service.go
```diff
@@ -0,0 +1,41 @@
...
+const (
+	MetricsServiceName = "ssp-operator-metrics"
+	MetricsPortName    = "http-metrics"
+)
+
+func PrometheusServiceLabels() map[string]string {
+	return common.PrometheusServiceLabels(MetricsServiceName)
+}
+
+func SspMetricsService(namespace string) *v1.Service {
```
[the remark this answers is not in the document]
Ruclo: it causes an import cycle so i had to extract it out of the controllers package.

internal/common/labels.go
```diff
@@ -71,3 +72,9 @@ func addInstanceLabels(requestInstance *ssp.SSP, to map[string]string) {
 func copyLabel(from, to map[string]string, key string) {
 	to[key] = from[key]
 }
+
+func PrometheusServiceLabels(serviceName string) map[string]string {
+	return map[string]string{
+		PrometheusLabelKey: serviceName,
+	}
+}
```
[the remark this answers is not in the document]
Ruclo: perhaps exporting only the function would be better? i also renamed the function to MetricsServiceLabels().

internal/operands/metrics/resources.go
akrejcir: The above comment still applies.

The string `template-validator-metrics` is used to connect 2 resources in the cluster: `Service` and `ServiceMonitor`. It is a good code style if the 2 strings are also the same constant in the code. Then we can be sure, that there is not a typo in one of them.

Can you extract the constant to a different package, so it does not cause import loop?

Ruclo:
I tried to extract the related functionality in the third commit that was part of this branch. Was this a bad thing ? Do you mind giving me a review on that ?

internal/controllers/ssp_controller.go
```diff
@@ -62,24 +62,27 @@ const (
...
-	crdList        crd_watch.CrdList
+	log                logr.Logger
+	operands           []operands.Operand
+	lastSspSpec        ssp.SSPSpec
+	subresourceCache   common.VersionCache
+	topologyMode       osconfv1.TopologyMode
+	areCrdsMissing     bool
+	olmDeployment      bool
+	sspServiceHostname string
+	client             client.Client
```
0xFelix: Keep space between L72/73?

config/default/kustomization.yaml
```diff
@@ -22,8 +22,6 @@ bases:
 - ../webhook
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 #- ../certmanager
-# [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
```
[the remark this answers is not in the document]
Ruclo: There is no prometheus directory in the parent directory, so it's not relevant. But this should be removed in another commit perhaps. I decided to keep it there for now.

main.go
```diff
@@ -210,24 +207,50 @@ func newPrometheusServer(metricsAddr string, cache cache.Cache) (*prometheusServ
...
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse certificate: %w", err)
+	}
+
+	if cert.Subject.CommonName != "" {
+		return cert.Subject.CommonName, nil
+	}
+
+	return cert.DNSNames[0], nil
```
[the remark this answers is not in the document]
Ruclo: I think they should match, can this be kept in place as a fallback?

tests/service_controller_test.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"kubevirt.io/ssp-operator/internal"
```
akrejcir: The same as above.

akrejcir:
Looks good now.

/approve

main.go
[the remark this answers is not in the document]
0xFelix: No?

PR #1640, opened 2025-12-11
fix: Use PATCH call to update metadata of immutable objects

akrejcir:
When a new field is added to the API of an immutable object before SSP has the new API code imported, the operator will ignore the new fields.

If it tries to update metadata, the UPDATE request will not contain the new field in spec, and so it will look like spec modification. The webhook would reject it.

Related jira: https://issues.redhat.com/browse/CNV-74677

**Release note**:
```release-note
None
```


0xFelix (commented):
Is there any downside to always patching instead of updating objects? I don't think we need to make a distinction between immutable spec or not regarding that.

akrejcir:
We can reopen if needed.

/close

PR #1723, opened 2026-03-24
fix: Alert should not show when version label on templates changes

akrejcir:
The label `app.kubernetes.io/version` on template is set from SSP object, and it should not trigger the alert.

This PR was assisted by AI.

Jira: https://redhat.atlassian.net/browse/CNV-77259

**Release note**:
```release-note
None
```


codingben (approved):
/lgtm

0xFelix (commented):
/approve

PR #1726, opened 2026-03-25
[release-v0.19] tests: Disable tekton tests for upstream CI

akrejcir:
The tekton is no longer installed properly from the github release yaml. Probably the upstream container image was removed from the repository, or is not accessible.

This commit also removes code that installs tekton in the upstream CI.

**Release note**:
```release-note
None
```

### DIFF
--- a/automation/common/deploy-kubevirt-and-cdi.sh
+++ b/automation/common/deploy-kubevirt-and-cdi.sh
@@ -12,9 +12,6 @@ metadata:
   name: ${NAMESPACE}
 EOF
 
-# Deploy Tekton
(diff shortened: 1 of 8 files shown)


0xFelix (commented):
Is there another LTS version that could work?

akrejcir:
This is the latest version with the API that we used. Newer version would need different CRD names and API.
I think it is not worth it to change the code on this old branch.

I've also checked the availability of later versions v0.68, but it was also not available.

0xFelix (commented):
/approve

PR #1732, opened 2026-03-27
[release-v0.18] tests: Disable tekton tests for upstream CI

akrejcir:
The tekton is no longer installed properly from the github release yaml. Probably the upstream container image was removed from the repository, or is not accessible.

This commit also removes code that installs tekton in the upstream CI.

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

PR #1760, opened 2026-04-21
chore: CI: configure Kubevirt to use emulation

akrejcir:
The cluster that runs tests for CI may not have nested virtualization enabled. Our tests only check if a VM is running, and they can be slower.

This change is needed for: https://github.com/openshift/release/pull/77840

**Release note**:
```release-note
None
```

### DIFF
--- a/automation/e2e-functests/run.sh
+++ b/automation/e2e-functests/run.sh
@@ -8,7 +8,7 @@ set -e
 # CI_OPERATOR_IMG - path of the operator image in the local repository accessible on the CI
 # CI_VALIDATOR_IMG - path of the validator image in the local repository accessible on the CI
 
-./automation/common/deploy-kubevirt-and-cdi.sh
+USE_EMULATION=true ./automation/common/deploy-kubevirt-and-cdi.sh
 
 # Install network policy to block traffic to/from SSP and its components by default
 KUBECTL=oc ./hack/install-network-policy.sh
--- a/automation/e2e-upgrade-functests/run.sh
+++ b/automation/e2e-upgrade-functests/run.sh
@@ -12,7 +12,7 @@ set -e
 SOURCE_DIR=$(dirname "$0")
 
 # Deploy KubeVirt and CDI
-./automation/common/deploy-kubevirt-and-cdi.sh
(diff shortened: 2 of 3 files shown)


akrejcir:
The CI does not work, and this PR is one of the change needed to fix it.

automation/e2e-functests/run.sh
```diff
@@ -8,7 +8,7 @@ set -e
 # CI_OPERATOR_IMG - path of the operator image in the local repository accessible on the CI
 # CI_VALIDATOR_IMG - path of the validator image in the local repository accessible on the CI
 
-./automation/common/deploy-kubevirt-and-cdi.sh
```
codingben, line 11: Does it mean that emulation will be always enabled even if the cluster node supports the nested virtualization?

automation/e2e-upgrade-functests/run.sh
```diff
@@ -12,7 +12,7 @@ set -e
 SOURCE_DIR=$(dirname "$0")
 
 # Deploy KubeVirt and CDI
-./automation/common/deploy-kubevirt-and-cdi.sh
```
codingben, line 15: I'm not sure about this `USE_EMULATION=true` when running the test on a local cluster. What if the developer will forget to disable it? Can you disable it from `Makefile`?

automation/e2e-functests/run.sh
akrejcir: Yes. We only have 2 tests that try running a VM. And they are not very important, maybe we can remove them in the future.

automation/e2e-upgrade-functests/run.sh
akrejcir: Usually this script `run.sh` is not run by a developer, but by the CI. I think we can assume that developer knows what he is doing. And even if emulation is enabled here, it is a kubevirt setting that can be disabled again on running cluster.

I sometimes run `deploy-kubevirt-and-cdi.sh` when I want to setup my development cluster, but not the `run.sh` script.

We cannot use `Makefile` because Kubevirt is deployed before running `make`.

codingben (approved):
/lgtm

akrejcir:
The CI was fixed, so this is no longer needed.

/close

PR #1787, opened 2026-05-06
fix: Do not change .status before reconciliation when SSP is unchanged

akrejcir:
Changing the `.status` to `Deploying` and then to `Deployed` on every reconciliation loop caused warnings in the event logs.

Jira: https://redhat.atlassian.net/browse/CNV-82471

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/controllers/ssp_controller.go
+++ b/internal/controllers/ssp_controller.go
@@ -220,7 +220,6 @@ func (s *sspController) Reconcile(ctx context.Context, req ctrl.Request) (res ct
 		}
 		reqLogger.Info(fmt.Sprintf("Pausing SSP operator on resource: %v/%v", instance.Namespace, instance.Name))
 		instance.Status.Paused = true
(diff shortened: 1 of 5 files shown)


akrejcir:
> - In `Reconcile`, the condition to run `preUpdateStatus` compares `Status.ObservedGeneration` to `instance.Generation`; since preUpdateStatus uses `request.Instance.Generation`, consider consistently using the same reference (`sspRequest.Instance`) throughout to avoid confusion and accidental divergence if the code ever stops sharing the same object.

Fixed.

> - `Paused` is now set to `false` in both `preUpdateStatus` and `updateStatus`; consider centralizing this unpause logic in one place (or a small helper) to avoid duplication and ensure future changes to pause behavior only need to be made once.

I want to keep this change simple.

jcanocan:
Nice job @akrejcir!
/lgtm

0xFelix (commented):
/approve

Thanks

PR #1810, opened 2026-05-11
fix: poll SSP status in test_id:6204 to wait for cluster to stabilize

geetikakay:
[test_id:6204] was doing single check on SSP phase and conditions causing a flaky failure if the operator was still reconciling. Wrap the assertions in Eventually to poll until the cluster stabilizes

Fixes #

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/validator_test.go
+++ b/tests/validator_test.go
@@ -343,20 +343,22 @@ var _ = Describe("Template validator operand", func() {
 	})
 
 	It("[test_id:6204]should set Deployed phase and conditions when validator pods are running", decorators.Conformance, func() {
-		foundSsp := getSsp()
+		Eventually(func(g Gomega) {
+			foundSsp := getSsp()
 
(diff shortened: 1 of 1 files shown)


akrejcir:
The tests already wait for SSP to me in phase `Deployed` in `waitUntilDeployed()` that is run in `BeforeEach()`. 
I'm not sure this PR helps. I will look into it.

/hold

geetikakay:
> The tests already wait for SSP to me in phase `Deployed` in `waitUntilDeployed()` that is run in `BeforeEach()`. I'm not sure this PR helps. I will look into it.
> 
> /hold

On 4.20, this failed occasionally.. could be a transient state so a single check sometimes passed and sometimes didn’t. 
Eventually could solve this part, if there are product issues and we will not reach phasedeployed , these will still fail and timeout.

PR #1811, opened 2026-05-13
test: fix flaky TLS test

akrejcir:
There is a race when the `vm-console-proxy` deployment is repeatedly deleted and recreated during a test run. Sometimes the pod name that was retrieved by the test no longer exists by the time the test tries to connect to it.

In this change, the Eventually loop also gets a new pod name.

**Release note**:
```release-note
None
```


codingben (approved):
/lgtm

jcanocan:
Nice catch!
/lgtm

0xFelix (commented):
/approve

PR #1767, opened 2026-04-24
test: Added test for TLS config on the vm-console-proxy pod

akrejcir:
**What this PR does / why we need it**:
The test checks that the vm-console-proxy pod uses the specified TLS config.

**Which issue(s) this PR fixes**: 
Jira: https://redhat.atlassian.net/browse/CNV-76854
Jira: https://redhat.atlassian.net/browse/CNV-77644

**Release note**:

```release-note
None
```

PR #1812, opened 2026-05-15
tests: Use signed client cert when connecting to vm-console-proxy

akrejcir:
There are 2 changes in this PR:
- Filter out insecure cipher suites from the ConfigMap passed to vm-console-proxy.
- Functional test uses a client certificate signed by the same CA as the one used by kubernetes API server.

This PR fixes failing downstream tests.

**Release note**:
```release-note
None
```

### DIFF
--- a/internal/operands/vm-console-proxy/reconcile.go
+++ b/internal/operands/vm-console-proxy/reconcile.go
@@ -2,6 +2,7 @@ package vm_console_proxy
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 
 	ocpv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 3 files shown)


codingben:
How did it work before?

akrejcir:
I've added this test case only recently, and it does not work in downstream since my previous PR: https://github.com/kubevirt/ssp-operator/pull/1767

akrejcir:
It's still not working, as we can see in the CI failure.

/hold

0xFelix (commented):
Can you request a certificate injected by Service CA operator instead?

akrejcir:
> Can you request a certificate injected by Service CA operator instead?

Good idea. I will look into it.

akrejcir:
I've tried using the `CertificateSigningRequest` (https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/) , but it didn't work. Probably API server uses a different CA, than the CSR.

akrejcir:
I've added another commit so SSP only passes valid cipher suites to `vm-console-proxy`.

internal/operands/vm-console-proxy/reconcile.go
```diff
@@ -443,7 +444,16 @@ func createTlsConfigFile(tlsProfile *ocpv1.TLSSecurityProfile) (string, error) {
...
+	secureCiphers := make(map[string]struct{})
+	for _, suite := range tls.CipherSuites() {
+		secureCiphers[suite.Name] = struct{}{}
+	}
+
+	for _, cipher := range crypto.OpenSSLToIANACipherSuites(profileSpec.Ciphers) {
+		if _, ok := secureCiphers[cipher]; ok {
+			result.Ciphers = append(result.Ciphers, cipher)
+		}
+	}
```
0xFelix, line 456: After this loop, return an error if `result.Ciphers` is empty but `profileSpec.Ciphers` wasn't?

0xFelix (commented):
One nit

/approve

internal/operands/vm-console-proxy/reconcile.go
akrejcir: done

akrejcir: Wait, that is not correct. For TLS 1.3, there some ciphers enabled by default and they are not part of the silice, so the slice can be empty. I will revert my change.

PR #1837, opened 2026-06-15
chore: Update golang version to 1.26

akrejcir:
Updated the golang version to 1.26

**Release note**:
```release-note
None
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ARG TARGET_ARCH
 RUN microdnf install -y make tar gzip which && microdnf clean all
 
 RUN ARCH=$(uname -m | sed 's/x86_64/amd64/') && \
-    curl -L https://go.dev/dl/go1.25.7.linux-${ARCH}.tar.gz -o go.tar.gz && \
+    curl -L https://go.dev/dl/go1.26.4.linux-${ARCH}.tar.gz -o go.tar.gz --fail && \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@ ARG TARGET_ARCH
 RUN microdnf install -y make tar gzip which && microdnf clean all
 
 RUN ARCH=$(uname -m | sed 's/x86_64/amd64/') && \
-    curl -L https://go.dev/dl/go1.25.7.linux-${ARCH}.tar.gz -o go.tar.gz && \
+    curl -L https://go.dev/dl/go1.26.4.linux-${ARCH}.tar.gz -o go.tar.gz && \
```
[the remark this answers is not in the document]
akrejcir: done

api/go.mod
```diff
@@ -1,6 +1,6 @@
 module kubevirt.io/ssp-operator/api
 
-go 1.25.7
+go 1.26
```
codingben, line 3: FWIW [kubevirt/kubevirt](https://github.com/kubevirt/kubevirt/blob/main/go.mod#L3) is still using 1.24, are you sure you want to merge it now?

ksimon1: We are not dependent on the kubevirt/kubevirt project. We can use newer versions even if they are still using an older version.

0xFelix (commented):
/approve

PR #1845, opened 2026-07-03
chore: Improve logging when tests fail to delete SSP

akrejcir:
Sometimes a test fails when waiting for SSP to be deleted, but then before the test lane exists, the SSP deletion succeeds, and we don't see the failure reason in the must-gather archive.

This changes prints the SSP object into the log in this case.

Jira: https://redhat.atlassian.net/issues?selectedIssue=CNV-89022

**Release note**:
```release-note
None
```


codingben (approved):
/lgtm

PR #1900, opened 2026-08-03
feat: Allow configuring TLS curve preference

akrejcir:
Read TLS group restrictions from the `.spec.tlsSecurityProfile` field on SSP object, and apply them to the webhook and template-validator TLS configs.

The TLS config is not yet passed to the `vm-console-proxy`, because the project needs to be updated to support it first.

Jira: https://redhat.atlassian.net/browse/CNV-84872
Jira: https://redhat.atlassian.net/browse/CNV-84871

**Release note**:
```release-note
None
```


main.go
```diff
@@ -88,27 +88,32 @@ func getConfigForClient(ctx context.Context, cfg *tls.Config, cache cache.Cache)
...
+	var tlsProfileSpec *ocpconfigv1.TLSProfileSpec
 	if tlsProfile.Type == ocpconfigv1.TLSProfileCustomType {
-		minVersion, err := crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
-		if err != nil {
-			return nil, err
-		}
-		cfg.MinVersion = minVersion
-		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &ctrl.Log)
-		return cfg, nil
+		tlsProfileSpec = &(*tlsProfile.Custom).TLSProfileSpec
```
0xFelix: No need to deref?

`&(tlsProfile.Custom).TLSProfileSpec` should work

tests/port-forwarding.go
```diff
@@ -28,15 +29,45 @@ type portForwarderImpl struct {
...
+func (p *portForwardConn) LocalAddr() net.Addr {
+	return portForwardAddr{}
+}
+
+func (p *portForwardConn) RemoteAddr() net.Addr {
+	return portForwardAddr{}
+}
+
+func (p *portForwardConn) SetDeadline(t time.Time) error {
+	return errors.ErrUnsupported
```
0xFelix: Can it not just return `nil`?

internal/common/crypto_policy.go
```diff
@@ -74,15 +82,15 @@ func CipherIDs(names []string, logger *logr.Logger) (cipherSuites []uint16) {
...
 
-func selectCipherSuitesAndMinTLSVersion(profile *ocpv1.TLSSecurityProfile) ([]string, ocpv1.TLSProtocolVersion) {
+func selectTLSProfileSpec(profile *ocpv1.TLSSecurityProfile) *ocpv1.TLSProfileSpec {
 	if profile == nil {
-		return nil, ""
+		return nil
 	}
 	if profile.Custom != nil {
-		return profile.Custom.Ciphers, profile.Custom.MinTLSVersion
+		return &profile.Custom.TLSProfileSpec
```
0xFelix: Perhaps return a deep copy here too? To avoid mutating the SSP object?

main.go
akrejcir: true, done.

tests/port-forwarding.go

internal/common/crypto_policy.go
akrejcir: done

tests/crypto_policy_test.go
```diff
@@ -93,12 +134,20 @@ var _ = Describe("Crypto Policy", func() {
 					MaxTLSVersion:      tls.VersionTLS13,
 					OpenSSLCipherNames: []string{},
 				},
+				{
+					MaxTLSVersion: tls.VersionTLS13,
+					Groups:        []ocpv1.TLSGroup{ocpv1.TLSGroupX25519},
+				},
 			},
 			disallowedConfigs: []clientTLSOptions{
```
0xFelix, line 142: Can there be combinations of TLS1.3 and unsupported ciphers?

akrejcir: No. The ciphers for TLS 1.3 are not configurable in the golang library. They are always present, and the `CipherSuites` field in `tls.Config` is ignored. This is also written in the documentation:
> CipherSuites is a list of enabled TLS 1.0–1.2 cipher suites. The order of the list is ignored. Note that TLS 1.3 cipher suites are not configurable.

0xFelix: Ok great!

main.go
```diff
@@ -88,27 +88,32 @@ func getConfigForClient(ctx context.Context, cfg *tls.Config, cache cache.Cache)
...
+	var tlsProfileSpec *ocpconfigv1.TLSProfileSpec
 	if tlsProfile.Type == ocpconfigv1.TLSProfileCustomType {
-		minVersion, err := crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
-		if err != nil {
-			return nil, err
-		}
-		cfg.MinVersion = minVersion
-		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &ctrl.Log)
-		return cfg, nil
+		tlsProfileSpec = &tlsProfile.Custom.TLSProfileSpec
```
0xFelix: Deep copy this one too?

internal/template-validator/virtinformers/vmcache.go
```diff
@@ -225,6 +245,7 @@ func (v *vmCache) Replace(list []interface{}, _ string) error {
 
 		key := vmCacheKey(metaObj)
 		val := newVmCacheValue(metaObj)
+		lastRv = metaObj.GetResourceVersion()
```
0xFelix: Should this not use the resource version passed as parameter to this func?

0xFelix (commented):
It might be worth adding a release note to this PR.

main.go
akrejcir: done

internal/template-validator/virtinformers/vmcache.go
akrejcir: Good point, I missed the parameter.

main.go
```diff
@@ -88,27 +88,32 @@ func getConfigForClient(ctx context.Context, cfg *tls.Config, cache cache.Cache)
...
-		minVersion, err := crypto.TLSVersion(string(tlsProfile.Custom.MinTLSVersion))
-		if err != nil {
-			return nil, err
-		}
-		cfg.MinVersion = minVersion
-		cfg.CipherSuites = common.CipherIDs(tlsProfile.Custom.Ciphers, &ctrl.Log)
-		return cfg, nil
+		tlsProfileSpec = tlsProfile.Custom.TLSProfileSpec.DeepCopy()
+	} else {
+		tlsProfileSpec = ocpconfigv1.TLSProfiles[tlsProfile.Type].DeepCopy()
```
0xFelix, line 101: This duplicates the selection logic from `selectTLSProfileSpec` in `internal/common/crypto_policy.go`. If that helper were exported, this block could be replaced with a single call, centralizing the logic and avoiding subtle divergences (e.g., this checks `Type == Custom` while the helper checks `Custom != nil`).

internal/template-validator/tlsinfo/tlsinfo.go
```diff
@@ -187,14 +190,27 @@ func (ti *TLSInfo) updateTLSOptions(directory string) {
...
 		ti.minTLSVersion = 0
 		ti.cipherSuites = nil
+		ti.curveIDs = nil
 		ti.tlsOptionsError = fmt.Errorf("TLS Configuration broken, min version misconfigured: %w", err)
 		logger.Log.Error(ti.tlsOptionsError, "TLS options are not valid",
 			"directory", directory)
 		return
 	}
 
+	curveIDs, err := common.CurveIDs(tlsOptions.TLSGroups, nil)
```
0xFelix, line 200: Passing `nil` for the logger here means unsupported TLS groups are silently dropped without any log output. In `main.go` the logger is provided so unsupported groups are logged. Consider passing `logger.Log` here for consistent diagnostics.

internal/common/crypto_policy.go
```diff
@@ -101,3 +109,38 @@ func tlsVersionToHumanReadable(version ocpv1.TLSProtocolVersion) (string, error)
...
+			curveIDs = append(curveIDs, curveID)
+		} else {
+			if logger != nil {
+				logger.WithName("TLSSecurityProfile").Info("Unsupported TLS group name: ", "TLS group name", group)
+			}
+			unsupported = append(unsupported, group)
+		}
+	}
+	if len(curveIDs) == 0 {
+		return nil, fmt.Errorf("all passed TLS curves are unsupported: %v", unsupported)
```
0xFelix, line 142: nit: The function parameter and OpenShift types use "groups" (`[]ocpv1.TLSGroup`), but the error message says "curves". Consider: `"all passed TLS groups are unsupported: %v"`.

0xFelix (commented):
Few more AI comments

internal/common/crypto_policy.go
akrejcir: done

internal/template-validator/tlsinfo/tlsinfo.go
akrejcir: done

main.go
akrejcir: I will do this in a future PR. There is a library with common functionality that is useful for this conversion from OpenShift TLS profile to the tls.Config. But I'm waiting until a PR is merged for the group support:
https://github.com/openshift/controller-runtime-common/pull/22

0xFelix: ACK

0xFelix (commented):
/approve

codingben (approved):
/lgtm

Looks good from my side.

PR #1902, opened 2026-08-05
tests: Retry VM creation if template validator webhook is not reachable

akrejcir:
Sometimes during test runs, the webhook may be unreachable for a short time. This can happen when a previous test has recreated the validator pods.

Jira: https://redhat.atlassian.net/browse/CNV-94356

**Release note**:
```release-note
None
```


codingben (approved):
Makes sense. Good catch.

/lgtm

PR #1924, opened 2026-08-25
tests: fix flaky network policy restore-with-pause tests

RoniKishner:
The `testNetworkPolicyResource` helper used in the functional test suite had an `UpdateFunc` that replaced the `NetworkPolicy` `PodSelector` with an empty `metav1.LabelSelector{}`. An empty `PodSelector` matches all pods in the namespace, including the ssp-operator pod itself.

Since that policy only permitted port 8768, the ssp-operator's validating webhook on port 9443 became unreachable while the SSP CR was paused for the test

Because unpausing requires an Update on the SSP CR, which must pass through that very webhook, a deadlock occurred

This fix changes the mutation to add an extra MatchLabels requirement of `ssp.kubevirt.io/test-mutated`, and since no real pod in the cluster has that label, the modified selector now matches zero pods instead of all pods.

Fixes #

The test's purpose is to verify the SSP operator detects and restores drift on a managed resource. For that to work, the EqualsFunc just needs to see a difference between the original spec and the mutated spec, the selection of pods has no effect on this test.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/validator_test.go
+++ b/tests/validator_test.go
@@ -59,7 +59,10 @@ func testNetworkPolicyResource(name, namespace string, expectedLabels map[string
 		Resource:       &networkv1.NetworkPolicy{},
 		ExpectedLabels: expectedLabels,
 		UpdateFunc: func(policy *networkv1.NetworkPolicy) {
-			policy.Spec.PodSelector = metav1.LabelSelector{}
+			if policy.Spec.PodSelector.MatchLabels == nil {
+				policy.Spec.PodSelector.MatchLabels = map[string]string{}
+			}
+			policy.Spec.PodSelector.MatchLabels["ssp.kubevirt.io/test-mutated"] = "true"
 		},
 		EqualsFunc: func(old *networkv1.NetworkPolicy, new *networkv1.NetworkPolicy) bool {
 			return reflect.DeepEqual(old.Spec, new.Spec)


tests/validator_test.go
```diff
@@ -59,7 +59,10 @@ func testNetworkPolicyResource(name, namespace string, expectedLabels map[string
 		Resource:       &networkv1.NetworkPolicy{},
 		ExpectedLabels: expectedLabels,
 		UpdateFunc: func(policy *networkv1.NetworkPolicy) {
-			policy.Spec.PodSelector = metav1.LabelSelector{}
+			if policy.Spec.PodSelector.MatchLabels == nil {
+				policy.Spec.PodSelector.MatchLabels = map[string]string{}
+			}
+			policy.Spec.PodSelector.MatchLabels["ssp.kubevirt.io/test-mutated"] = "true"
```
ksimon1, line 65: Where is this label `ssp.kubevirt.io/test-mutated` label defined in ssp project?

RoniKishner: It is not defined, and no 1 uses it.
the reason to add it, is so that it won't catch any pods and the selector will pick 0 pods instead of all of them

ksimon1: ok, please add this explanation to the commit message.

RoniKishner: Update the commit message and also the PR description.

jcanocan: A small comment in the line will be helpful too.

0xFelix (commented):
/approve

jcanocan:
> @jcanocan: #1924 failed to apply on top of branch "release-v0.25":
> 
> ```
> Patch is empty.
> hint: When you have resolved this problem, run "git am --continue".
> hint: If you prefer to skip this patch, run "git am --skip" instead.
> hint: To record the empty patch as an empty commit, run "git am --allow-empty".
> hint: To restore the original branch and stop patching, run "git am --abort".
> hint: Disable this message with "git config set advice.mergeConflict false"
> ```
> 
> Details

@RoniKishner could you please take a look at this?

PR #1941, opened 2026-09-07
tests: use NIST curves instead of X25519 in crypto policy test

akrejcir:
`X25519` is not a FIPS-approved key-exchange group. Downstream images are built with `GOEXPERIMENT=strictfipsruntime` where `crypto/tls` drops `X25519` from the supported curves.

The crypto policy test pinned its `allowed` group to `X25519`, so every allowed config requiring an `X25519` handshake failed with `tls: handshake failure`.

Using `secp256r1` for the allowed group and custom profile, keeping `secp521r1` and `secp384r1` as disallowed.

Jira: https://redhat.atlassian.net/browse/CNV-95805

**Release note**:
```release-note
None
```


0xFelix (commented):
/approve

